PR #137, opened 2020-10-01
feature: Markdown field

mihaimdm22:
Let me know what you think about current design and functionality of Markdown.

I am open to feedback!

A quick question? On which field are we going to put this type of field (currently on post - description)? I need an empty field for tests, since description is taken for Trix.

Screenshot 2020-10-26 at 19 04 30


adrianthedev:
This looks good! Let's try [this](https://github.com/Ionaru/easy-markdown-editor) package instead. Maybe they have a vue version too.

spec/system/avo/markdown_field_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+        textarea_markdown.fill_in with: "## Code \nInline `code` \n```js \nSample text here... \n``` \n--- Ordered \n1. Lorem ipsum dolor sit amet \n2. Consectetur adipiscing elit".html_safe
+
+        click_on 'Save'
+        wait_for_loaded
+
+        click_on 'Show Content'
+
+        description_element = find_field_element('description')
+        within(description_element) {
+          expect(find(:xpath, "//div[@class='v-show-content scroll-style scroll-style-border-radius']")).to have_text('Code')
```
adrianthedev, line 100: Can we make these a bit more restrictive? Like what should be between `<code></code>` tags to check for that html?

This worked for me:
`expect(find('.v-note-show.single-show code')).to have_text 'Works for us!!!'`

adrianthedev (commented):
Looking good. There are just a few tweaks I'd like to see here.

PR #174, opened 2020-10-29
fix: make fields Code, Country, Date, Datetime, Select and Trix disabled on readonly = true

mihaimdm22:
let me know what you think

there were a lot of fields with this thing missing

KeyValue is still not working but I will open another ticket and work on it cause there are many more problems


adrianthedev (approved):
GG!

Pushed a tiny change to Code. When it was readonly it did not show the content ut just the dash. And changed the naming ediatble->disabled.

PR #175, opened 2020-10-29
fix: KeyValue disable options issues

mihaimdm22:
There were some problems with the keyvalue field.

Pressing enter in the keyvalue key or value inputs doesnt submit the form. In case of value, enter either focuses on next row(if existing) or adds a new row.


adrianthedev (approved):

adrianthedev:
Disabled is not working for this, but I read your comment on https://github.com/avo-hq/avo/pull/174

PR #178, opened 2020-11-02
feature: allow no_confirmation for Actions

mihaimdm22:
Made the requested changes, everything should be fine now. Added the Make Admin no_confirmation: true action too.

Keep in touch that in order no_confirmation to work, no fields have to be on the action definition.


adrianthedev (commented):
Great job! Awesome start and a very elegant way to solve this! GG! 

Push it to over the finish line. If you need a quick "no_confirmation" action, try "Make Admin" for a "User".

PR #179, opened 2020-11-02
feat: height option for Code field

mihaimdm22:
Instead of `rows` and `auto` I thought about implementing a `height` option that will enforce the size of the field(if specified), otherwise it defaults to auto and the size will fit all content. I think it will be easier for the administrator to configure and also makes sense to default to auto. Let me know what you think about this.


adrianthedev (approved):
GG!

Pushed the same change to show view too!

PR #183, opened 2020-11-11
feature: set now/today button for datetime/date fields

mihaimdm22:
image

image


adrianthedev (changes requested):
We should change the value of flatpickr instead of `this.value`.

PR #185, opened 2020-11-12
fix: filters tests

mihaimdm22:
created also new filter on team


mihaimdm22:
There were issues with other tests. IDK why yet but I'll just make another PR.

PR #190, opened 2020-11-20
fix: loaders are not visible on some pages

adrianthedev:
Loaders on `new`, `show` and `create` views are not visible. This fixes that

loaders not visible


mihaimdm22 (changes requested):
Loader has different top margins.

adrianthedev:
I fixed that by displaying the filters and search bar. This way it is larger on Index but looks natural.

[](https://i.gyazo.com/9a6dd766339dfa8fbf7a8cd106af248d.gif)

PR #201, opened 2020-12-15
Update inline javascript tags to allow for csp nonces

venables:
Avo currently does not work if you have a content security policy which blocks `unsafe-inline`.  This Pull Request uses the built-in option for `javascript_tag` to allow a nonce to mark the inline script as safe.  

This change is also safe for apps without a content_security_policy.


adrianthedev (approved):
Thank you for your contribution! 🤙

adrianthedev:
Tagged under `0.4.4`. Thanks again!

PR #207, opened 2020-12-29
refactor: hotwire rewrite

adrianthedev:
With this PR we're pulling out VueJS from Avo and adding the Hotwire stack. This should make the app easier to extend and maintain. 🤙

We'll use this branch as a `develop` branch for this rewrite.


.vscode/settings.json
```diff
@@ -1,6 +1,6 @@
 {
   "ruby.specSaveFile": true,
-  "ruby.specCommand": "HEADFULL=true RAILS_ENV=test bundle exec rspec",
+  "ruby.specCommand": "HEADFULL=true RAILS_ENV=test rspec",
```
tmilewski, line 3: I don't know that this file should be in the repo.

adrianthedev: Yes. You're right, it shouldn't be commited to the repo. I'll remove it.

PR #216, opened 2021-01-17
(AVO-55) feature: toastr alerts

adrianthedev:
In order to test this PR go to the users page, update the `UserPolicy.index?` to be `false` and hit the keys `r r r` to force turbo to reload the page.

You should get redirected to `/avo` with a notice that you are not allowed to perform that action.

image


mihaimdm22 (approved):
Goog to go!

PR #219, opened 2021-01-17
(AVO-43) feature: belongs to relation

adrianthedev:


mihaimdm22 (commented):
Going to `http://localhost:3030/avo/resources/posts` raises this error for me:
- ActionView::MissingTemplate in Avo::Resources#index --- Missing partial avo/fields/index/_belongs-to-field with {:locale=>[:en], :formats=>[:html], :variants=>[], :handlers=>[:raw, :erb, :html, :builder, :ruby, :jbuilder]}. Searched in:

I think this is bcs of not including `_belongs-to-field` in the PR.

PR #222, opened 2021-01-18
(AVO-37) feature: Select field

mihaimdm22:
Changes to the field:

- merged enum and options in the same configuration called options
- values have to be (label, key)
- changed `display_value` to `display_with_value` (defaults to false)

Usage:
```ruby
select :stage, options: { 'Discovery': 'discovery', 'Ideea': 'ideea', 'Done': 'done', 'On hold': 'on hold', 'Cancelled': 'cancelled' }, placeholder: 'Choose the stage.', display_with_value: true
```
```ruby
select :stage, options: ::Project.stages, placeholder: 'Choose the stage.', display_with_value: false
```


app/views/avo/fields/edit/_select-field.html.erb
```diff
@@ -0,0 +1,23 @@
+<%
+  if !field[:options].nil?
```
adrianthedev, line 2:
```suggestion
  if field[:options].present?
```

app/views/avo/fields/edit/_select-field.html.erb
```diff
@@ -0,0 +1,23 @@
...
+    if field[:display_with_value]
+      options = field[:options].invert
+      options.each {|key, value| options[key] = value }
```
adrianthedev, line 5: Not sure this works as intended. I still get the values inverted.

```suggestion
      options = options.map do |key, value|
        [key, key]
      end.to_h
```

app/views/avo/fields/edit/_select-field.html.erb
```diff
@@ -0,0 +1,23 @@
...
+      initialValue = field[:value]
+    else
+      options = field[:options]
+      initialValue = options[field[:value]]
```
adrianthedev, line 9: Ruby's variables are `snake_case`.

```suggestion
      initial_value = options[field[:value]]
```

app/views/avo/fields/index/_select-field.html.erb
```diff
@@ -0,0 +1,11 @@
+<%
+  if field[:display_with_value] && !field[:options].nil? && field[:options][field[:value]]
```
adrianthedev, line 2:
```suggestion
  if field[:display_with_value] && field[:options].present? && field[:options][field[:value]]
```

app/views/avo/fields/edit/_select-field.html.erb
adrianthedev, line 2: This whole `if/else` is very difficult to understand. Also I think it doesn't really work in all situations. 

The reason is that if it's an `enum` we never detect it. And because we don't detect it we'll never know if the `value` is truly the value. Might be the `label` and the `enum` casted it.

Example:

If in the DB we have `done` when we configure the enum, we'll always get the label `Done`. If it's just an `options` hash we'll get the raw value `done`.
If we get the label `Done` we won't be able to treat that.

I'd bring back the `enum` option on the field.

adrianthedev (changes requested):
The whole `if/else` in `edit` is very difficult to comprehend. Also we can't detect if it's an `enum` casted value.

Let's pair on this 😉

app/views/avo/fields/edit/_select-field.html.erb
mihaimdm22: I've come up with something that works. Let me know when you review in order to look on it together. There's no need for both enum and options.

PR #226, opened 2021-01-20
(AVO-38) feature: Status field

mihaimdm22:


app/views/avo/fields/common/_status.html.erb
```diff
@@ -0,0 +1,43 @@
...
+  classes += "text-red-700" if status == 'fail'
+%>
+<div class="flex flex-shrink-0">
+    <div class="<%= classes %>"><%= label %></div>
+    <% if status == 'load' %>
+      <div class="spinner ml-1"></div>
+    <% end %>
+</div>
+
+<style>
```
adrianthedev, line 12: We should not add style tags like this. 

I think we should place them inside `frontend/css/components/status.css` and import them into `application.css`

app/views/avo/fields/common/_status.html.erb
```diff
@@ -0,0 +1,43 @@
+<%
+  classes = "inline-block "
+  classes += "text-red-700" if status == 'fail'
+%>
+<div class="flex flex-shrink-0">
+    <div class="<%= classes %>"><%= label %></div>
+    <% if status == 'load' %>
```
adrianthedev, line 7: Could we do `loading` instead of `load` and `failed` instead of `fail`? I know we had it like that before, but I think it makes more sense like this.

app/views/avo/fields/index/_status-field.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= index_field_wrapper field: field do %>
+  <%
+    status = 'success'
+    status = 'fail' if field[:failed_when].include? field[:value].to_sym
+    status = 'load' if field[:loading_when].include? field[:value].to_sym
+  %>
```
adrianthedev, line 6: I think we can leave all of this out and not have to deal with the duplication if we update the `hydrate_field` like so.

```ruby
      def hydrate_field(fields, model, resource, view)
        status = 'success'
        status = 'fail' if @failed_when.include? fields[:value].to_sym
        status = 'load' if @loading_when.include? fields[:value].to_sym

        {
          status: status,
        }
      end
```

adrianthedev (commented):
There are just a few tweaks I'd like to make:

- move the styling
- refactor the `hydrate_field` method
- change the states to `loading` & `failed`

app/views/avo/fields/index/_status-field.html.erb
mihaimdm22: we don't have the value set there, so that's why we need to compute on the view components

adrianthedev: we don't? it's in the `fields` (which should be renamed).

mihaimdm22: I've tried to get the value in there as `fields[:value]` and it returned `nil`. Didn't investigate further.

mihaimdm22: This is what `fields` contains:
```ruby
{:id=>:status, :computed=>false, :name=>"Status", :translation_key=>nil, :block=>nil, :component=>"status-field", :required=>false, :readonly=>false, :updatable=>true, :sortable=>false, :nullable=>true, :null_values=>[nil, ""], :computable=>true, :is_array_param=>false, :format_using=>false, :placeholder=>"Status", :help=>nil, :default=>nil, :value=>nil}
```

PR #227, opened 2021-01-21
(AVO-30) feature: Heading field

mihaimdm22:


app/views/avo/fields/common/_heading.html.erb
```diff
@@ -0,0 +1,13 @@
+<div class="flex items-start py-1 leading-tight border-t bg-gray-700 text-white text-xs">
+    <div class="p-3 pl-6 h-full w-full">
+      <% if value.present? %>
+        <% if as_html %>
+          <%== value %>
```
adrianthedev, line 5: I didn't know you could do that. I always used `.html_safe` GG!

adrianthedev (approved):

PR #277, opened 2021-03-11
refactor: remove intermediary avo filters

adrianthedev:
This PR sheds two intermediary `Filter` classes from Avo, moving their functionality to the public base classes.

Each class inherits from the `Filter` class and adds it's own configuration (`self.template`). Not sure if we should expose this in the public base classes.

What do you think?


adrianthedev:
Not such a good ideea after all

PR #325, opened 2021-04-05
feature: breadcrumbs and dynamic page title

adrianthedev:
Docs for [breadcrumbs](https://docs.avohq.io/1.0/customization.html#breadcrumbs) and [page titles](https://docs.avohq.io/1.0/customization.html#page-titles).


lib/avo/app.rb
```diff
@@ -38,6 +38,12 @@ def cache_store
         app[:cache_store]
       end
 
+      def app_name
+        ::Rails.application.class.module_parent_name
```
xtr3me, line 42: Should this be filled with the app_name configuration?
```
Avo.configuration.app_name
```

adrianthedev: I was burning the midnight oil and that's why I haven't merged it yet. 
I was looking for some sort of configuration mechanism. `Avo.configuration.app_name` looks perfect.

One other thing I wanted to do is add a package like https://github.com/kpumuk/meta-tags. So there's more info for the page title.

xtr3me: Combine that with the breadcrumbs ticket on the board and you have yourself a nice polished addition to Avo.
The only thing i'm unsure about is if Hotwire (turbo) allows you to set the title when you are changing pages. (Since it generally doesn't send the html skeleton (`html`, `head` & `body` tags))

https://github.com/hotwired/turbo-rails/blob/fec33d9bc767aec612b283620d2a74e78c1f90ae/app/controllers/turbo/frames/frame_request.rb

adrianthedev: I see that Turbo changes the title when doing full page navigation and doesn't interfere when loading a frame. Seems like that's good.

adrianthedev: Good idea to bundle it with the breadcrumbs feature. Do you recommend any gem for breadcrumbs?
I found https://github.com/weppos/breadcrumbs_on_rails to be quite popular.

xtr3me: I've checked out a couple of them and the one you listed seems really nice :+1:

PR #339, opened 2021-04-10
Fix enums for integer values

### LINKED ISSUE #3233: Mapbox address field
# Mapbox Field

These snippets below are what I had after a workshop I held at Visuality last year. They won't deliver the finisehd result, but it's a great starting point.
This will give you a custom field that you can add to your app and will search mapbox.

### Approach

The user will start to type some text in the input and when they select an option from the dropdown they will get a payload at their disposal from mapbox.

```ruby
# Let's imagine that mapbox returns this payload for the query
# *I don't have the response in front of me
{
  full_address: "55 George Washington st., New York, USA",
  street_address: "55 George Washington st.",
  street_number: "55",
  city: "New York",
  state: "NY",
  country: "USA",
}
```

#### API

```ruby
# when the form gets submitted it will populate the `address` column with the full address received from mapbox and all the parsed fields onto the record.

# Nothing happens when the form gets submitted
field :address, as: :mapbox_address

# All the fields from the response are passed to the record
field :address, as: :mapbox_address, map: :all

# only the fields in `map` are being passed to the record
field :address, as: :mapbox_address, map: [:full_address, :street]

# the fields from `map` are passed to the record with the key being overwritten by `custom_mapping`
field :address, as: :mapbox_address, map: [:street_address, :city, :country], custom_mapping: {street_address: :street, city: :town}

# give the user access to the `record` and `respomse` and they can choose what happens
field :address, as: :mapbox_address, populate: -> {
  # maybe we can use `update_using` instead here?
  record.street = response[:street_address]
}
```

If the record has an `address` column/property, let's display that in the UI, but let's think about a way to show the input only if the user wants it. We can talk about it.

### Snippets

```ruby
class DynamicAddressField < Avo::Fields::BaseField
  def initialize(name, **args, &block)
    super(name, **args, &block)
  end
end
```

```erb
<!-- edit component html -->
  <div data-controller="address">
    <%= @form.text_field @field.id,
      class: classes("w-full"),
      autocomplete: "shipping address-line1",
      disabled: disabled? %>
  </div>
<% end %>
```

```js
import { Controller } from '@hotwired/stimulus'

  static targets = ['element', "addressTextInput"]

  connect() {
    console.log('address_field_controller', this.addressTextInputTarget)
    // console.log(this.elementTarget)
    // this.initField()
  }

  onInput(e) {
    console.log("onInput", e.target.value)
  }

  initField() {
    mapboxsearch.autofill({
      accessToken: "pk.eyJ..."
    });
  }
}
```

marcelobarreto:
Hi, as I mentioned, the bug was raised because a Rails behavior, but, the enum was broken for integer values, so I do fixed it.

image
image
image

Fixes #3233


adrianthedev:
Thank you @marcelobarreto!
I'll review and merge it on monday.

lib/avo/fields/select_field.rb
```diff
@@ -31,12 +31,16 @@ def options_for_select
...
+          else
+            value
+          end
+        elsif display_value
           value
-        elsif enum.present?
-          options[value]
         else
-          options.invert[value]
+          options.invert.stringify_keys[value]
```
adrianthedev, line 43: This conditional is more complex than I wanted it to be, but it works in all scenarios (enum hash, enum array, options hash and all versions with `display_value` `true/false`).

adrianthedev (approved):
Apologies for the late review. The select field with `enum`, `options` and `display_value: true/false` arguments have a lot of scenarios.

adrianthedev:
Thank you for your contribution @marcelobarreto 💪

PR #375, opened 2021-04-29
only shows items in the user's dropdown if available

nicolascochin:


adrianthedev:
I pushed another tiny update where we hide the dropdown arrow if the sign-out is not present. That whole menu should probably be refactored a bit better later down the road.

Thank you for your contribution @nicolascochin!

nicolascochin:
:+1:

PR #423, opened 2021-06-01
Allow custom view component directory

### LINKED ISSUE #415: [Feature request] Allow custom view component directory
From https://github.com/avo-hq/avo/discussions/414

The custom field generator generates several view components that are used to render the field. It would be nice if the generator could make use of a configuration option that allows the directory of these generated components to change. Currently, it places the components in `app/components`, but in my apps, I put components in different directories, such as `app/view_components` or `app/frontend/components`.

nulldotpro:
This branch closes #415.


adrianthedev (approved):

PR #437, opened 2021-06-17
fix: remove created_at assumption on index

adrianthedev:


spec/dummy/config/initializers/friendly_id.rb
```diff
@@ -0,0 +1,107 @@
+# FriendlyId Global Configuration
```
adrianthedev, line 1: This is unrelated but needed to get done.

Issue #436, opened 2021-06-17
[BUG]Table View Sort Not Work if Using `deafult_scope` or `resolve_query_scope`


luonet:
### Describe the bug
If you are using the `default_scope` in the model,
or `resolve_query_scope` in the resource;
When you set a field as `sortable`,
the record in the table view would not be sortable by the field.

And the record would be sortable only if you remove the `default_scope` from the model,
and `resolve_query_scope` from the resource.

### System configuration
**Avo version**: 
1.8.1

**Rails version**:
6.1.3.2

**Ruby version**:
3.0.0

### Additional context
I think this bug might be caused by not `unscope(:order)`.
If we want to sort only by the `sortable` field, we should `unscope(:order)` of the query object first.

mihaimdm22:
hey @luonet 

first of all thanks for reporting this and also suggesting how to solve it

we added a fix for it and tested for both `default_scope` and `resolve_query_scope` and it works

we'll release very soon -> get latest version of gem -> test it -> let me know if you have additional questions

luonet:
Hi @mihaimdm22

First, thanks for your fix and response.
So fast!

I noticed following lines added:
```ruby
unless @index_params[:sort_by].eql? :created_at
  @query = @query.unscoped
end
```

I have one concern:
The `@query.unscoped` will unscope all scopes, and I doubt it might be too agressive?

For example, suppose we declared a default_scope in the Movie model,
```ruby
  default_scope { where(published: true).order(release_date: :desc)
```

If we apply unscoped to the default_scope, the order and where scope would all be removed.
So when you click the field sorting in the movies index, the unpublished movies will show!

So I think it's better to only unscope order in this case?
```ruby
  @query = @query.unscope(:order)
```

Thanks again!
Thanks you very much!

mihaimdm22:
Hi @luonet 

Sorry for my late reply!

I pushed the change you suggested regarding scoping.

Now looking into issue #439.

Thanks for your patience!

luonet:
Hi @mihaimdm22 

Thank you very much!

Got the latest version and tested in my environment.

I think for most cases, it perfectly works;
but in some extreme edge cases, it may have a very subtle issue(NOT a bug).

Suppose we have a Movie model with default_scope:
```ruby
class Movie < ApplicationRecord
  default_scope { order(release_date: :desc) }
```

In the avo movie resource index page, if no sortable field clicked, the expected SQL should be:
```SQL
[...]

PR #438, opened 2021-06-17
fix: allow sorting columns on index when default_value or resolve_query_scope are set

mihaimdm22:
Fixes https://github.com/avo-hq/avo/issues/436

### DIFF
--- a/app/controllers/avo/base_controller.rb
+++ b/app/controllers/avo/base_controller.rb
@@ -38,6 +38,9 @@ module Avo
 
       # Sort the items
       if @index_params[:sort_by].present?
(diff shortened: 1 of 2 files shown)


app/controllers/avo/base_controller.rb
```diff
@@ -38,6 +38,9 @@ def index
 
       # Sort the items
       if @index_params[:sort_by].present?
+        if not @index_params[:sort_by].eql? :created_at
```
adrianthedev, line 41:
```suggestion
        unless @index_params[:sort_by].eql? :created_at
```

Issue #420, opened 2021-05-29
[BUG] readonly does nothing on :has_many


TheFehr:
### Describe the bug
I have ActionJobs that generate a lot of different model's which I would love to observe with avo. 
There is multiple instances where there is a :has_many relationship, that I don't want to be able to attach myself, since this is the ActionJob's job.
I was expecting the readonly flag to disable attachments on the :has_many fields.

### Steps to Reproduce
1. Add a :has_many field with readonly: true
2. Open that resource
3. See the attach button

### Expected behavior
3. Not see the attach button

### System configuration
**Avo version**:
1.7.2

**Rails version**:
6.0.3.6

**Ruby version**:
2.7.3p183

adrianthedev:
Yes. This is a valid bug. There's [a discussion](https://github.com/avo-hq/avo/discussions/398) about hiding those buttons.

We could go with `readonly: true` or with options like `disable_creation: true/false` and disable_attaching: true/false` so you can disable them however you need them.

PR #442, opened 2021-06-23
fix: allow readonly on resources

mihaimdm22:
[BUG] readonly does nothing on :has_many #420

### DIFF
--- a/app/components/avo/fields/has_one_field/show_component.html.erb
+++ b/app/components/avo/fields/has_one_field/show_component.html.erb
@@ -5,8 +5,10 @@
 <% else %>
   <%= render Avo::PanelComponent.new(title: @field.id.capitalize) do |c| %>
     <% c.tools do %>
-      <%= a_link helpers.resource_attach_path(@resource.model.model_name.route_key, @resource.model.id, @field.id), color: 'indigo', 'data-turbo-frame': 'attach_modal' do %>
-        <%= svg 'view-grid-add' %> <%= t('avo.attach_item', item: @field.id).capitalize %>
+      <% unless @field.readonly %>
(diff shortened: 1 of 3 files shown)


app/components/avo/views/resource_index_component.rb
```diff
@@ -45,18 +45,30 @@ def available_view_types
...
-    @reflection.present? && klass.is_a?(::ActiveRecord::Reflection::HasManyReflection)
+    @reflection.present? && klass.is_a?(::ActiveRecord::Reflection::HasManyReflection) && !is_field_read_only
   end
 
   def can_detach?
-    @reflection.present? && @reflection.is_a?(::ActiveRecord::Reflection::HasOneReflection) && @models.present?
+    @reflection.present? && @reflection.is_a?(::ActiveRecord::Reflection::HasOneReflection) && @models.present? && !is_field_read_only
+  end
+
+  def is_field_read_only
```
adrianthedev, line 62: In the context of `resource_index` there is no one `field`.

```suggestion
  def is_a_has_many_and_is_read_only
```

app/components/avo/fields/has_one_field/show_component.html.erb
```diff
@@ -5,8 +5,10 @@
 <% else %>
   <%= render Avo::PanelComponent.new(title: @field.id.capitalize) do |c| %>
     <% c.tools do %>
-      <%= a_link helpers.resource_attach_path(@resource.model.model_name.route_key, @resource.model.id, @field.id), color: 'indigo', 'data-turbo-frame': 'attach_modal' do %>
-        <%= svg 'view-grid-add' %> <%= t('avo.attach_item', item: @field.id).capitalize %>
+      <% unless @field.readonly %>
```
adrianthedev, line 8: How about using the same pattern with a `can_attach?` method?

PR #454, opened 2021-06-29
Use same resource name as show component

### LINKED ISSUE #453: [BUG] Action page not sending `resource_ids` on `index` page
### Describe the bug
Performing actions on 'index' page has no effect on ticked rows

### Steps to Reproduce

Steps to reproduce the behaviour:
1. Create any action
2. Attempt to use that action from the `index` page through selecting a couple of rows

### Expected behavior
The action should be called with the relevant model data

### Actual behavior
Any empty array is passed to the `handle` method. On further inspection it looks like no `resource_ids` are being passed, so the bug is likely to be in the front end:

Screenshot 2021-06-29 at 14 07 23

The 'show' action is working as expected.

### System configuration
**Avo version**: 1.10.1

**Rails version**: 6.1.3.2

**Ruby version**: 2.6.6

chrismdp:
Fixes #453 if the correct approach (see details there). Grateful for feedback on the change!

### DIFF
--- a/app/components/avo/views/resource_index_component.html.erb
+++ b/app/components/avo/views/resource_index_component.html.erb
@@ -24,7 +24,7 @@
 
     <% c.body do %>
       <div class="flex justify-between pt-2 pb-2 min-h-16"
-        data-selected-resources-name="<%= @resource.plural_name.downcase %>"
+        data-selected-resources-name="<%= @resource.model.model_name.route_key %>"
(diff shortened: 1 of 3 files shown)


adrianthedev (approved):

adrianthedev:
Released in https://rubygems.org/gems/avo/versions/1.10.2

Issue #449, opened 2021-06-25
[BUG] Detach button should not be visible when user is not authorized to delete


santhanakarthikeyan:
### Describe the bug
On the resource show page, `Detach` button is visible when there is an association. Suppose, current_user not authorized to delete that resource then we should not show detach button. 

Ref: https://github.com/avo-hq/avo/blob/master/app/components/avo/views/resource_show_component.html.erb:L13

### Steps to Reproduce
1. Go to the show view of a resource with a has one relation
2. "Detach ..." button is visible
3. Create a policy with delete? set as false
4. "Detach ..." button is still visible

### Expected behavior
When current_user is not authorized to delete then `Detach` button should not be visible, similar to the edit or delete button on the same page. 

Note: Even `GO back` button should not be visible. 

### Actual behavior
current_user is not authorized to delete. He can see `Detach` button. 

### System configuration
**Avo version**: 1.10.1

**Rails version**: 6.0

**Ruby version**: 2.7.2

### Screenshots
<img width="1196" alt="Screenshot 2021-06-25 at 6 03 53 PM" src="https://user-images.githubusercontent.com/35058248/123425410-be2d5f80-d5df-11eb-8038-144c5aaf52d6.png">

mihaimdm22:
Hi @santhanakarthikeyan 

With release 1.11 we released [Associations policies](https://docs.avohq.io/1.0/authorization.html#associations) that solve this bug. Refer to documentation and let me know if you have additional questions/requests.

Cheers!

santhanakarthikeyan:
Hi @mihaimdm22 

Thanks for solving it on short notice. Kudos to your dedication! 👍

PR #457, opened 2021-06-30
feature: relationships authorization

mihaimdm22:
[BUG] Detach button should not be visible when user is not authorized to delete #449

### DIFF
--- a/app/components/avo/resource_component.rb
+++ b/app/components/avo/resource_component.rb
@@ -7,12 +7,12 @@ class Avo::ResourceComponent < ViewComponent::Base
     @resource.authorization.authorize_action(:destroy, raise_exception: false)
   end
 
-  def get_association_policy policy_method
+  def authorize_association_for(policy_method)
     association_policy = true
     if @reflection.present?
       reflection_resource = ::Avo::App.get_resource_by_model_name(@reflection.active_record.name)
       if reflection_resource.present?
-        method_name = (policy_method + '_' + @reflection.name.to_s + '?').to_sym
+        method_name = ("#{policy_method}_#{@reflection.name.to_s}?").to_sym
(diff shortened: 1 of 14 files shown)


app/components/avo/resource_component.rb
```diff
@@ -7,6 +7,22 @@ def can_delete?
     @resource.authorization.authorize_action(:destroy, raise_exception: false)
   end
 
+  def get_association_policy policy_method
```
adrianthedev, line 10:
```suggestion
  def get_association_policy(policy_method)
```

app/components/avo/resource_component.rb
```diff
@@ -7,6 +7,22 @@ def can_delete?
...
+    association_policy = true
+    if @reflection.present?
+      reflection_resource = ::Avo::App.get_resource_by_model_name(@reflection.active_record.name)
+      if reflection_resource.present?
+        method_name = (policy_method + '_' + @resource.model_class.model_name.singular_route_key + '?').to_sym if @resource
```
adrianthedev:
```suggestion
        method_name = ("#{policy_method}_#{@resource.model_class.model_name.singular_route_key}?").to_sym if @resource
```

app/components/avo/views/resource_index_component.rb
```diff
@@ -52,11 +52,11 @@ def can_attach?
     klass = @reflection
     klass = @reflection.through_reflection if klass.is_a? ::ActiveRecord::Reflection::ThroughReflection
 
-    @reflection.present? && klass.is_a?(::ActiveRecord::Reflection::HasManyReflection) && !has_reflection_and_is_read_only
+    @reflection.present? && klass.is_a?(::ActiveRecord::Reflection::HasManyReflection) && !has_reflection_and_is_read_only && get_association_policy('attach')
```
adrianthedev, line 55: You might need to rename this one. I was expecting it to return a policy.

PR #473, opened 2021-07-12
feature: Select all on controller

chrismdp:
Simple Stimulus controller which uses JS to click all select buttons on the page at once, and additionally ensures the status of the checkbox is consistent with the row checkboxes.

The implementation isn't efficient and might get slow with very large pages, but felt like the simplest option for now that could work.

First proper PR - feedback very welcome! Happy to work on improving this as needed.

Closes #452.

### DIFF
--- a/app/components/avo/index/resource_table_component.html.erb
+++ b/app/components/avo/index/resource_table_component.html.erb
@@ -1,6 +1,6 @@
 <div class="w-full">
   <% if @resources.present?%>
-    <table class="w-full px-4 overflow-hidden">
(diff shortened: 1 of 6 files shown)


adrianthedev:
👀 Checking it right now.

chrismdp:
The tests seemed to fail for unrelated reasons 🤔

adrianthedev:
Yup. Totally unrelated. There's an issue with the `ViewComponent` library.

adrianthedev:
Released in `1.11.5`

PR #522, opened 2021-10-15
Ensure fields array is initialized for heading

calebuharrison:
This fixes a bug wherein the fields array is not initialized if a heading is used before any fields in a resource definition.


adrianthedev:
Thank you for the contribution. Cutting a release tomorrow.

PR #535, opened 2021-11-08
I18n resources

yorch:
* Supports new Resources' attribute for i18n called `translation_enable` to use translations key using convention `avo.resource_translations` and `avo.field_translations`. The goal is to avoid having to declare a translation key for each resource and field, and just use a convention.
* Fixes i18n on Fields as it's not currently working.

### DIFF
--- a/lib/avo/base_resource.rb
+++ b/lib/avo/base_resource.rb
@@ -16,6 +16,7 @@ module Avo
     class_attribute :includes, default: []
     class_attribute :model_class
     class_attribute :translation_key
+    class_attribute :translation_enabled
     class_attribute :default_view_type, default: :table
     class_attribute :devise_password_optional, default: false
     class_attribute :actions_loader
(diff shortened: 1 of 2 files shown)


lib/avo/base_resource.rb
```diff
@@ -16,6 +16,7 @@ class BaseResource
     class_attribute :includes, default: []
     class_attribute :model_class
     class_attribute :translation_key
+    class_attribute :translation_enable
```
adrianthedev, line 19: This being a boolean, wouldn't a better name be `translation_enabled` (ending in `d`)?

adrianthedev (approved):
Pretty good to go! 

Do you think `translation_enabled` would be a better name for that switch?

Thanks for the contribution!

lib/avo/base_resource.rb
yorch: Good call! made the change

yorch:
Thanks for the reviews!

yorch:
@adrianthedev gently reminder :)

adrianthedev:
Apologies for the delay. I was a bit worried about having the `attr_reader`s and the methods in the same class.

adrianthedev:
Would you mind adding a few words in the [docs](https://github.com/avo-hq/docs/tree/master/src/1.0) about this change?

adrianthedev:
Released in `1.16.4`

PR #543, opened 2021-11-25
Issue-533 Fix the TextFilter bug with Unicode string

luonet:
The TextFilter could not search Unicode string.

It's caused by the `btoa` method,
which will throw an error when encontering the Unicode string.

Rewrite the Base64 encode/decode method suggested by Stackoverflow:
https://stackoverflow.com/questions/30106476/using-javascripts-atob-to-decode-base64-doesnt-properly-decode-utf-8-strings


adrianthedev:
Releasing in 1.17.1 in a few minutes

PR #544, opened 2021-11-29
Fix: Rename boolean filter argument `value` => `values`

vfonic:
NOTE: I didn't manage to get the tests running locally!

Boolean filter receives a hash of key-values: keys being names/strings/features/flags and values being `true` or `false` (selected or not)

Example `values`:

```rb
> values
{
      "Cart Button Effects": true,
  "Customer Scarcity Timer": false,
             "Discount Pop": false,
                "IQ Slider": false
}
```


adrianthedev:
Because they are more than one, isn't `values` more correct than `value`? Doesn't `values["Cart Button Effects"]` look better than `value["Cart Button Effects"]`?

adrianthedev:
Lol. I misread it. It totally makes sense 😛 

Thanks!

PR #550, opened 2021-12-08
chore: update view_component

adrianthedev:


adrianthedev:
Closing this bc we don't see that issue anymore. The issue stopped manifesting itself when we removed a few Ruby versions from the testing env.

PR #551, opened 2021-12-13
Add direct_upload boolean to file field

jony1993:
This PR adds a `direct_upload` boolean to file fields.
By setting `direct_upload: true` you can use the active storage direct_upload feature ([https://edgeguides.rubyonrails.org/active_storage_overview.html#direct-uploads](https://edgeguides.rubyonrails.org/active_storage_overview.html#direct-uploads)) e.g. to prevent timeouts for large files when you host your application on heroku ([https://devcenter.heroku.com/articles/request-timeout#uploading-large-files](https://devcenter.heroku.com/articles/request-timeout#uploading-large-files))

Example:
`  field :cover_photo, as: :file, is_image: true, as_avatar: :rounded, direct_upload: true`


adrianthedev:
Hey. Thanks for the contribution.

I don't see the AWS upload in my network requests. Do you see it with this update?

image

app/components/avo/fields/common/multiple_file_viewer_component.rb
```diff
@@ -3,10 +3,11 @@
 class Avo::Fields::Common::MultipleFileViewerComponent < ViewComponent::Base
   include Avo::ApplicationHelper
 
-  def initialize(id:, file:, is_image:, resource:, button_size: :md)
+  def initialize(id:, file:, is_image:, direct_upload:, resource:, button_size: :md)
```
adrianthedev, line 6: We need a default here otherwise we'll get a crash.

```suggestion
  def initialize(id:, file:, is_image:, direct_upload: false, resource:, button_size: :md)
```

app/components/avo/fields/common/single_file_viewer_component.rb
```diff
@@ -3,10 +3,11 @@
 class Avo::Fields::Common::SingleFileViewerComponent < ViewComponent::Base
   include Avo::ApplicationHelper
 
-  def initialize(id:, file:, is_image:, resource:, button_size: :md)
+  def initialize(id:, file:, is_image:, direct_upload:, resource:, button_size: :md)
```
adrianthedev, line 6:
```suggestion
  def initialize(id:, file:, is_image:, direct_upload: false, resource:, button_size: :md)
```

blacklist-arcc:
Maybe missing something. We currently working on a new web app using rails 6, webpacker, latest avo and google cloud with google storage backend. The direct upload function need the `activestorage.js` as described on: https://edgeguides.rubyonrails.org/active_storage_overview.html#direct-uploads

As we need this feature right now we implement the direct upload using a custom field and included the `activestorage.js` as shown in: https://docs.avohq.io/1.0/custom-asset-pipeline.html#webpacker

ATM we get an alert for the enpoint "direct_upload" saying that the blob can not be create but i think the `activestorage.js` is needed in order to work anyway.

sheff3rd:
@blacklist-arcc 
I'm helping with PR and I'm not sure onto next steps. Correct me if I'm wrong:
I should implement a new field using custom asset piplines for direct upload, and change how urls are generated for it

blacklist-arcc:
Hi, thanks for your reply. I noticed avo using a different asset pipeline and ofc not including the active storage js stuff. That should be included in avo as shown in the rails documentation. The other problem is that avo using his own scope - by default /avo/ and direct upload paths are created using rails form helper and current scope. But direct links are mapped by the active record gem not using the avo scope so the generated links all go to 404. Removing the /avo/ prefix will make the links working, so the link generation for direct uploads has to be changed also. Alternatively you can use the path variable for the direct links and generate the file field without the form helpers.

blacklist-arcc:
The current PR is fine for the `direct_upload` boolean variable in the form builder. But if this should work you have to include the active_record js in avo directly like:

```ruby
<%= javascript_pack_tag 'avo_application' %>
```

`avo_application.js`
```ruby
import * as ActiveStorage from "@rails/activestorage"
ActiveStorage.start()
```
The link problem can be solved in two ways. First, using a different scope for the file field form helper. Second, add another route inside the avo scope:

```ruby
post "/avo/rails/active_storage/direct_uploads" => "active_storage/direct_uploads#create"
```

I am sure that there are still much more clever and more general solutions available but this will do the job at the moment. It would be nice to see a progress bar instead disabled upload buttons.

adrianthedev:
Hey,

Adrian, maintainer of Avo here. Apologies for the radio silence. We're AFK with family, until the 5th. 
Indee we are not including Active Storage JS. I can't remember why we went that route but we'll look into it. 
If someone wants to add it to this  PR, then that's fine as well.

blacklist-arcc:
Hey, thanks for your reply - hope you have a great time with your family - it's more important than ever. I think all important information are shared in this discussion here and I will wait for @sheff3rd changes as the problem can be solved temporary with avo custom fields, a custom route and custom asset inclusion.

Leave a comment if I should go into this deeper and create a PR on my own.

adrianthedev:
That's the perfect response! Thank you for that 🙏

sheff3rd:
@blacklist-arcc hello again. I've added a custom route and custom asset for active storage, and tests seem to be passing for me locally. Can you please review it? Thanks in advance!

blacklist-arcc (commented):
The custom route is now "hardcoded" and may conflict with https://docs.avohq.io/1.0/customization.html#home-path or nested paths.

With default configuration the changes working as expected. If a new avo version is published i will test this on production on our current google cloud project.

config/routes.rb
```diff
@@ -2,6 +2,7 @@
   root "home#index"
 
   get "resources", to: redirect("/admin")
+  post "/rails/active_storage/direct_uploads", to: "/active_storage/direct_uploads#create"
```
adrianthedev, line 5: This was generating a weird route nested under `/avo`

```diff
--  post "/rails/active_storage/direct_uploads", to: "active_storage/direct_uploads#create"
++  post "/rails/active_storage/direct_uploads", to: "/active_storage/direct_uploads#create"
```

image

adrianthedev: Also, we don't need the `/avo` route because it will all be nested under `/avo` (or whatever `/admin` route is used).

adrianthedev (approved):
I'd love to see [progress bars](https://guides.rubyonrails.org/active_storage_overview.html#example) when uploading. Maybe in a future release.

blacklist-arcc:
Looks fine to me now. Maybe I will go into the progress bar tomorrow and share the js code or open a PR ;)

Thanks for your effort!

adrianthedev:
Thank you @jony1993, @sheff3rd, and @blacklist-arcc for the contribution! You made Avo a bit more awesome today!

blacklist-arcc:
> I published [1.18.0.pre.1](https://rubygems.org/gems/avo/versions/1.18.0.pre.1) with these changes. Can you try it in your app @blacklist-arcc?
> 
> Thank you!
> 
> `gem 'avo', '1.18.0.pre.1'` in your Gemfile

I will go into this tomorrow right away and share my feedback!

adrianthedev:
Cool. If all works out I'll push to `main` and release.

blacklist-arcc:
Ahhhh, the progress bar is incoming :) Thanks for your effort on this topic. Can you publish a new pre release version so that I can test it on our live instance tomorrow?!

blacklist-arcc:
I have tested the 1.18.0.pre.1 in our live environment. The direct_upload works fine! Progressbar is untestet as it's not included in this version.

adrianthedev:
Published `1.18.0.pre.2` now @blacklist-arcc.

Can you also try uploading and deleting some files? I found [a bug](https://github.com/avo-hq/avo/pull/551/commits/490c8870cf5dfdf77e066bf65bcb8915562a5b3e) and want to make sure everything works as expected.

blacklist-arcc:
Using 1.18.0.pre.2, direct upload works with progressbar, but deleting files is not possible. 404 =>
https://example.com/avo/resources/RESOURCE/ID/active_storage_attachments/FILENAME/ID

adrianthedev:
Checking right now.

adrianthedev:
This is weird and I don't know why it worked the first time. 
[Here](https://github.com/avo-hq/avo/blob/main/app/components/avo/fields/common/files_list_viewer_component.html.erb#L2) we loop through the attachments, [here](https://github.com/avo-hq/avo/blob/main/app/components/avo/fields/common/multiple_file_viewer_component.html.erb#L26) we're assigning the `signed_attachment_id` param as the attachment `id` and [here](https://github.com/avo-hq/avo/blob/main/app/controllers/avo/attachments_controller.rb#L24) we're searching for the blob (old way). We were supposed to search for the `ActiveStorage::Attachment` from the start.

Do you see anything wrong?

adrianthedev:
I reverted that change and going to tackle it in a different PR. Will merge and release after checks are done

blacklist-arcc:
> I reverted that change and going to tackle it in a different PR. Will merge and release after checks are done

Will be there a pre.4 release for testing with these changes?

adrianthedev:
Nope. Releasing `1.18.0` directly.

blacklist-arcc:
> Nope. Releasing `1.18.0` directly.

Works fine: delete and direct upload! Thanks for your effort on this.

Issue #559, opened 2022-01-10
[BUG] Avo doesn't handle STI very well


adrianthedev:
### Describe the bug
The way Rails handles STI breaks Avo's `Model` -> `Resource` association.

### Steps to Reproduce
1. Create two models `class Person < ApplicationRecord` and `class Spouse < Person` and resources for both
2. Add two records. One with type `Spouse`
2. Go to the index page for people
3. Click the Spose record

### Expected behavior
It should take you to a `PersonResource` page

### Actual behavior
Takes you to a `SpouseResourse` page.

### System configuration
**Avo version**: all

### Additional context
When you query `Person.all` you get `[Person, Spouse]`.

### Approach
We should handle STI somehow differently in Avo. Maybe a `PersonResource` with an option `acts_as_sti_for Spouse`.

PR #560, opened 2022-01-10
fix: use base class for link helpers with STI models

adrianthedev:
Fixes https://github.com/avo-hq/avo/issues/559 and https://github.com/avo-hq/avo/discussions/549


adrianthedev:
@rctneil can you try [1.18.2.pre.0](https://rubygems.org/gems/avo/versions/1.18.2.pre.0) and test it thoroughly?

In your testing, you should remove the `SpouseResource` and `SpouseController`. If all is ok, I'll merge it and release it.

rctneil:
@adrianthedev Hi Adrian,

Many thanks for putting this pre-release out.   I've done some testing in my app today and it does appear to be working so far.  I will raise an issue if I hit any issues regarding that aspect.

One quick question, is there a way to get a text field to link to another record?  Eg, a text field on a show view to link to a corresponding show view?

adrianthedev:
Awesome!

I just pushed [a PR](https://github.com/avo-hq/avo/pull/561) allowing `as_html` to text fields.

adrianthedev:
Actually a better option for that is the `format_using` [option](https://docs.avohq.io/1.0/field-options.html#fields-formatter) setting. You should try that one instead.

Issue #576, opened 2022-01-18
URL helper methods not available in computed field blocks `format_using` and lambdas


adrianthedev:
This is something that has been surfaced a few times.

Let's take a computed field where you need to pass in a link to a different resource. 

```ruby
  field :partner_home, as: :text, as_html: true do |model, *args|
	hey_url
  end
```

You would like to be able to use the URL helpers that you have available to your app, but because this block is not executed in a view, but in Avo's `BaseField` [here](https://github.com/avo-hq/avo/blob/main/lib/avo/fields/base_field.rb#L124) it does not have access to the regular URL helpers.

Same if you want to use `link_to` in a `format_using` lambda executed [here](https://github.com/avo-hq/avo/blob/main/lib/avo/fields/base_field.rb#L128).

```ruby
  field :website_url, as: :text,
                      required: true,
                      name: "Website",
                      as_html: true,
                      format_using: -> (url) { link_to(url, url, target: "_blank") }
```

The routes file for reference.

<details>
<summary>`routes.rb`</summary>

```ruby
get 'hey', to: 'home#hey'
```

</details>

### Question

How could we overcome this issue and give the user access to the URL helpers and other "regular" view methods? 
Is there a way to run that code in the "regular Rails context"?

We're looking to make this as painless as possible for the user. We're ok with having the user use the `main_app.` prefix for accessing the apps routes.

adrianthedev:
I was on this path with `view_context`. I haven't worked with it before, but this link to `view_component` definitely helps. I'll check it out and see how they use it.

I think I have a solution for the other issue with `link_to`. Will have a PR up tomorrow.

Thanks for the help.

PR #590, opened 2022-01-18
feature: get access to view_context from blocks

adrianthedev:
Fixes https://github.com/avo-hq/avo/issues/576

### DIFF
--- a/app/controllers/avo/application_controller.rb
+++ b/app/controllers/avo/application_controller.rb
@@ -19,7 +19,7 @@ module Avo
     add_flash_types :info, :warning, :success, :error
 
     def init_app
-      Avo::App.init request: request, context: context, root_path: avo.root_path.delete_suffix("/"), current_user: _current_user
+      Avo::App.init request: request, context: context, root_path: avo.root_path.delete_suffix("/"), current_user: _current_user, view_context: view_context
 
       @license = Avo::App.license
     end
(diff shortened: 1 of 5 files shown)


adrianthedev:
This is what I came up with so quickly @rickychilcott. 

I'm passing the `view_context` from the application controller on init and set it as a `class_attribute` on `App` to have access to it in other places. Not sure about the performance implications of doing that everytime. What do you think?

https://github.com/avo-hq/avo/pull/590/files#diff-04b4150f5f611929955839a39ee76ec0e57c2fb96b241aebcacfbf5ce3e5162aR22

Then, I'm delegating some things to it [from the field class](https://github.com/avo-hq/avo/pull/590/files#diff-f76981801622e99a4aa2ab1f99d911852978a891ca6cc34c8fc7d6e0496f15aaR10).

Finally, instead of doing `block.call` I [changed it](https://github.com/avo-hq/avo/pull/590/files#diff-f76981801622e99a4aa2ab1f99d911852978a891ca6cc34c8fc7d6e0496f15aaR130) to `instance_exec`.

rickychilcott:
Thank you @adrianthedev and sorry for the delay on tracking this out.`link_to` and `main_app.route_url` are both working well.

Thank you so much!

adrianthedev:
No worries! Glad it worked out.

Issue #588, opened 2022-01-18
Allow actions to be hidden in the index view


adrianthedev:
### Discussed in https://github.com/avo-hq/avo/discussions/575

<div type='discussions-op-text'>

<sup>Originally posted by **nulldotpro** January 17, 2022</sup>
I'd like to prevent some actions from showing up in the index view, only allowing them to appear in the show view. This could be as simple as an option set on the action class.</div>

PR #594, opened 2022-01-18
feature: actions visibility

adrianthedev:
Fixes https://github.com/avo-hq/avo/issues/588

### DIFF
--- a/spec/system/avo/actions_spec.rb
+++ b/spec/system/avo/actions_spec.rb
@@ -1,187 +1,211 @@
-# require 'rails_helper'
+require 'rails_helper'
 
-# RSpec.describe 'Actions', type: :system do
-#   let!(:roles) { { admin: false, manager: false, writer: false } }
-#   let!(:user) { create :user, active: true, roles: roles }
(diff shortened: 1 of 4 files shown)


spec/system/avo/actions_spec.rb
```diff
@@ -1,187 +1,211 @@
...
-#       let!(:roles) { { admin: false, manager: false, writer: false } }
-#       let!(:second_user) { create :user, active: true, roles: roles }
-#       let(:url) { '/admin/resources/users' }
+    context 'show' do
+      it 'does not find an action on show' do
+        visit "/admin/resources/users/#{user.id}"
 
-#       it 'displays the actions button disabled' do
-#         visit url
+        click_on 'Actions'
```
nulldotpro, line 21: When there aren't any actions visible for a view, I don't think the `Actions` button should render at all.

adrianthedev: It [should be hidden](https://github.com/avo-hq/avo/blob/main/app/components/avo/views/resource_index_component.html.erb#L4) when no actions.

PR #603, opened 2022-01-22
Translate key value field and base controller messages

yorch:


adrianthedev:
Thanks for your PR. This will probably land in the next release. You missed this one by a few minutes.

Is the [placeholder](https://github.com/avo-hq/avo/pull/603/files#diff-f76981801622e99a4aa2ab1f99d911852978a891ca6cc34c8fc7d6e0496f15aaR60) change necessary?

yorch:
@adrianthedev awesome!

If I remember correctly when I did it on my fork several weeks ago, was to fix i18n on placeholders, without, they would not get translated

yorch:
Later today I'll look at the test errors, seems related to placeholder, so need to fix those

adrianthedev:
So, we can revert those? Do they work now?

yorch:
No, what I meant was that the placeholder is a fix for i18n, otherwise, it won't get translated like this:

image

adrianthedev:
Ok, I see, I'll test it out.

adrianthedev:

PR #612, opened 2022-01-31
chore: set model before authorize action

adrianthedev:
Based on [this](https://discord.com/channels/740892036978442260/740893011994738751/935108679848566844) discussion.

image

### DIFF
--- a/app/components/avo/views/resource_edit_component.html.erb
+++ b/app/components/avo/views/resource_edit_component.html.erb
@@ -10,7 +10,7 @@
               <%= helpers.svg 'arrow-left' %> <%= t('avo.cancel').capitalize %>
             <% end %>
 
-            <% if @resource.authorization.authorize_action :update, raise_exception: false %>
+            <% if can_see_the_save_button? %>
               <%= a_button color: 'green', variant: 'outlined', spinner: true, type: :submit do %>
                 <%= helpers.svg 'save' %> <%= t('avo.save').capitalize %>
               <% end %>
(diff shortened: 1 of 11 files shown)


jdbann:
@adrianthedev I've just been updating to use the actual release version and I think something may have changed? It looks like you populate `@model_to_fill` in the `before_action` but don't actually assign it to `@model` until the actual controller action. This means `authorize_action` still sees an empty `@model` in `:create` actions so it's not possible to authorize based on the provided params.

adrianthedev:
@jdbann I see what you mean. I probably introduced that bug when I added the `fill_model` method. Can you try [`1.20.2.pre.1`](https://rubygems.org/gems/avo/versions/1.20.2.pre.1) and see if it fixes it?

jdbann:
> Can you try ~[`1.20.2.pre.1`](https://rubygems.org/gems/avo/versions/1.20.2.pre.1)~ [`1.20.2.pre.2`](https://rubygems.org/gems/avo/versions/1.20.2.pre.2) and see if it fixes it?

@adrianthedev Yeah, that seems to work a treat! Thanks 🥳

adrianthedev:
Cool! Thanks for bringing that up.

jdbann:
@adrianthedev Is it possible to apply this approach to the `Avo::RelationsController` as well? The `@model` passed in to the policy is always the class at the moment. The `fill_model` action [specifically excludes attach actions](https://github.com/avo-hq/avo/blob/62fb00996b6dde1b6a2e4018f2cccbeb76117876/app/controllers/avo/application_controller.rb#L126-L127) so I might be missing some alternative approach?

adrianthedev:
Hey @jdbann. This slipped through the cracks.

Can you post an example on where the `@model` is a class? In the policy under `attach_SOMETHING?` the `record` is actually an instantiated model, not a class.

adrianthedev:
Also, it would probably be best to open a new ticket.

PR #686, opened 2022-02-19
fix: Support older version of Pundit

rickychilcott:
# Description
Pundit 2.2.0 changes the Module that is required to be included, which is a breaking change. Since Pundit 2.1.x is still very viable for most apps, support both options.

# Checklist:

### DIFF
--- a/app/controllers/avo/application_controller.rb
+++ b/app/controllers/avo/application_controller.rb
@@ -1,6 +1,11 @@
 module Avo
   class ApplicationController < ::ActionController::Base
-    include Pundit::Authorization
+    if defined?(Pundit::Authorization)
+      include Pundit::Authorization
+    else
+      include Pundit
+    end
+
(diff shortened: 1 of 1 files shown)


rickychilcott:
Not sure what you think of this one @adrianthedev. 

I'm surprised that https://github.com/varvet/pundit/commit/4d9b5849f41ae540cd48b323e100fe74d76ded92 wasn't considered a breaking change (i.e. bumping to Pundit 3.0), but either way, I can't currently fix all of my OTHER dependencies to jump to Pundit v2.2 yet.

Does this simple version work or would you rather a Pundit version check?

adrianthedev:
Hey @rickychilcott. Yeah, this might work. It seems I haven't thought about this change as being a breaking change, but what you're saying makes sense.

I'll check it out and merge this before Monday's release.

PR #704, opened 2022-02-25
Hide header, sidebar and footer when print in browser

yorch:
# Description

Not sure if it's worth adding this at this point as I think you are working on a UI revamp, but thought I would share it regardless in case you want to merge it.

When printing through the browser, it will hide the header, sidebar and footer as they are usually not required / useful.

# Checklist:

## Manual review steps

Description...

1. Step 1
1. Step 2


adrianthedev:
That's not a bad idea. Good job!
I'd like to make this toggle-able with a configuration option to prevent unwanted behavior to current users.
Would you be comfortable adding it to the PR? 
If not, I can help with that.

These files would be impacted (from what I can think of)

 - the configuration flag something like `hide_layout_when_printing` with a default to `false`. 
 - initializer template https://github.com/avo-hq/avo/blob/main/lib/generators/avo/templates/initializer/avo.tt
 - configuration file https://github.com/avo-hq/avo/blob/main/lib/avo/configuration.rb
 - all the templates that you already added the classes

Thank you!

yorch:
@adrianthedev updated the PR to be driven by config

PR #722, opened 2022-03-09
Helper `turbo_frame_wrap` will delegate to turbo_frame_tag

rajaravivarma-r:
# Description

* Instead of creating the `turbo-frame` component ourself,
  it can be delegated to the `turbo_frame_tag` helper.
* Because of this change, we will be able to provide all arguments
  to `turbo_frame_wrap`, that are accepted by `turbo_frame_tag`.
  For example, `src` to load the `turbo-frame` in a separate request.
* Issue Fixed: Previously it was not possible to supply arguments like `src`,
  now it is possible.

Fixes # (issue)

# Checklist:

## Manual review steps

1. Create a controller
`HomeController`
```ruby
class HomeController < ApplicationController
  def index
  end

  def show
    flash.now[:info] = 'Show loaded'
    sleep(4) # Simulate slow operation
  end
end
```
2. Add routes.rb
`routes.rb`
```ruby
Rails.application.routes.draw do
  root to: 'home#index'
  get 'show', 'home#show'
end
```
3. Add views4.
`index.html.erb`
```erb
<%= turbo_frame_wrap('show', src: show_path) do %>
  Loading show
<% end %>
```
`show.html.erb`
```erb
<%= turbo_frame_wrap('show') do %>
  Show loaded from turbo frame...
<% end %>
```

Verification: After visiting the root page, the `Loading show...` will get replaced by `Show loaded from turbo frame...` after a few seconds, along with a flash notification.

### DIFF
--- a/app/components/avo/turbo_frame_wrapper_component.html.erb
+++ b/app/components/avo/turbo_frame_wrapper_component.html.erb
@@ -1,11 +1,11 @@
-<% if name.present? %><turbo-frame id="<%= name %>"><% end %>
+<%= turbo_frame_tag(*args, **kwargs) do %>
   <%
     # When rendering the frames the flashed content gets lost.
     # By including the alerts partial, the stimulus will pick them up and display them to the user.
   %>
   <% if helpers.turbo_frame_request? %>
-    <%= render Avo::AlertsComponent.new if helpers.flash.present? && name.present? %>
+    <%= render Avo::AlertsComponent.new if helpers.flash.present? %>
   <% end %>
 
   <%= content %>
-<% if name.present? %></turbo-frame><% end %>
+<% end %>
--- a/app/components/avo/turbo_frame_wrapper_component.rb
+++ b/app/components/avo/turbo_frame_wrapper_component.rb
@@ -1,9 +1,15 @@
 # frozen_string_literal: true
 
 class Avo::TurboFrameWrapperComponent < ViewComponent::Base
-  attr_reader :name
+  # Refer: https://github.com/github/view_component/issues/1099
+  # delegating `turbo_frame_tag` to helpers doesn't capture the
+  # content in the block.
+  include Turbo::FramesHelper
 
-  def initialize(name = nil)
-    @name = name
+  attr_reader :args, :kwargs
+
+  def initialize(*args, **kwargs)
+    @args = args
+    @kwargs = kwargs
   end
 end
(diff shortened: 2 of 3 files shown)


rajaravivarma-r:
Hi @adrianthedev,
This was needed for our production app. So I fixed it.
As a part of this change, I thought the `Avo::TurboFrameWrapperComponent`, did not add any value. So I removed it and followed your pattern of directly rendering a partial, as layout, with a block.

Let me know if the work needs some change.

adrianthedev:
Hey @rajaravivarma-r. I'll have a look in the next couple of days.

One thing I can tell you for sure is that [I switched](https://github.com/avo-hq/avo/pull/706) the Rails partial for a view component for performance reasons. View components render much faster and that particular one is used in a lot of places in Avo. 
So, if you could refactor the PR to feature the view component instead of the partial, that would be greatly appreciated. Thanks!

adrianthedev:
Also, I think it's amazing that you're using that API `turbo_frame_wrap` 💪

rajaravivarma-r:
That makes sense. I will switch that to a ViewComponent.

> Also, I think it's amazing that you're using that API turbo_frame_wrap 

I had to, because I wanted to render turbo frames with `flash` alerts.

rajaravivarma-r:
@adrianthedev I have updated it to use ViewComponent again. I tried delegating `turbo_frame_tag` to `:helpers` but it didn't capture the content we pass in block. I think the issue is equivalent (but not same) to this one, https://github.com/github/view_component/issues/1099.

Like it is mentioned in that issue, I included `include Turbo::FramesHelper` and it worked fine.

adrianthedev:
I noticed you removed the `<% if name.present? %>` conditional. What if it's not meant to be wrapped in a turbo-frame?

rajaravivarma-r:
I'm sorry, I don't get what I'm missing here, but why would someone call `turbo_frame_wrap` method and not expect it be wrapped in a `turbo-frame`? 

IMO, it doesn't convey the intention well, which leads us to methods like `turbo_frame_wrap(if: proc { true })`, or even more explicitly call the `turbo_frame_wrap` inside an `<% if name.present? %> ... <% end %>`.

**Alternatives: 
'** I just tested and found that I can write a my own `turbo_frame_wrap` because the `Avo::AlertsComponent` is available from the main app.

rajaravivarma-r:
Hi @adrianthedev, any thoughts on my above comment?

adrianthedev:
Closing this for now. Feel free to open it if it's still of interest

rajaravivarma-r:
Sure thing.

PR #852, opened 2022-04-17
Feature/customizable sorting

MrJoy:
# Description

Implementation for: https://github.com/avo-hq/avo/discussions/783

If `sortable` is given a `Proc`, then use that to adjust the query and apply whatever custom sorting behavior is desired.

# Checklist:

## Manual review steps

1. Bring up dummy Rails app
2. Navigate to `Users` resource
3. Sort by "First Name" column, and observe that rows are ordered not by first name by ID (descending).
4. Sort by "First Name" again, and observe that rows are ordered by ID but in ascending order.


adrianthedev (approved):
Amazing! Thank you for this PR! 

I moved the custom sorting from the `first_name` to the `is_writer` attribute to avoid confusion. People might click on the `first_name` column (because it's among the first ones) and resources would get sorted differently than they would think.

Again, Thanks! This is really useful!

PR #889, opened 2022-05-09
feature: live resources

adrianthedev:
# Description

Fixes https://github.com/avo-hq/avo/discussions/466

I got action cable working and a channel to send events through. At first, I thought I would be able to use Turbo frames, but I'm not sure that's possible in an easy manner. My initial strategy was to wrap the `ResourceShow` component in a `turbo-frame` and trigger an update through a dumb stimulus controller when a web-sockets event came in. 

That's not so easy because when you wrap something in a turbo-frame, all the links inside are "hijacked" to run as turbo. 
For all the content that comes back from those links, turbo looks for the same frame that the `ResourceShow` component is wrapped in. 
We can force all of our links and add the `target="_top"` attribute (although that's not pleasant and prone to human error), but there are going link from custom pieces of content (custom fields, and tools).

Now I am trying to dynamically wrap the component in a turbo-frame, trigger a reload and remove the frame on each websockets event. It's kinda hacky, but it could do the job.

We're open to suggestions on this one.

# Checklist:

## Manual review steps

Description...

1. Step 1
1. Step 2

### DIFF
--- a/app/components/avo/turbo_frame_wrapper_component.html.erb
+++ b/app/components/avo/turbo_frame_wrapper_component.html.erb
@@ -1,4 +1,4 @@
-<% if name.present? %><turbo-frame id="<%= name %>"><% end %>
+<% if name.present? %><turbo-frame id="<%= name %>" <%= "target=\"#{target}\"" if target.present? %>><% end %>
 <%
   # When rendering the frames the flashed content gets lost.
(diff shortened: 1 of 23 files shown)


adrianthedev:
After some ore fiddling, I think we don't have to wrap the resource view in a turbo_frame, but just give it an id and turbo streams will replace it.
This might be easier than thought.

PR #896, opened 2022-05-12
feature: customize the model name of a resource

jclarke:
# Description
This simple PR adds the ability to customize a resources name in the navigation links, page headers, breadcrumbs, etc by specifying the model_name in the resource class.  

# Checklist:

## Manual review steps
In a Resource class, specify a model name, e.g.
self.model_name =  "Location Search"

This model name specified is then used for navigation links, page headers, breadcrumbs instead of the class name.

### DIFF
--- a/lib/avo/base_resource.rb
+++ b/lib/avo/base_resource.rb
@@ -26,6 +26,7 @@ module Avo
     class_attribute :search_query_help, default: ""
     class_attribute :includes, default: []
     class_attribute :model_class
+    class_attribute :model_name
     class_attribute :translation_key
     class_attribute :translation_enabled, default: false
     class_attribute :default_view_type, default: :table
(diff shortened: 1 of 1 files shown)


adrianthedev:
Hey @jclarke, the same behavior can be achieved by using the localization (I18n) [feature](https://docs.avohq.io/2.0/localization.html#localizing-resources). Is there a reason not to use that feature?

I'm a bit hesitant about adding year another mechanism to change the resource's name. It increases the complexity and may add new edge-cases and maintenance work in the future.

jclarke:
Makes sense, I was just trying to be lazy and define everything in the resource object :)

adrianthedev:
All good! Thank you for understanding!

The complexity level is relatively high in some areas. I'm trying to keep things manageable.

Issue #904, opened 2022-05-16
Searchable belongs_to resources don't apply attach_scope on new


davidlormor:
### Describe the bug

The `attach_scope` doesn't appear to work as expected on create when search is enabled for a belongs to relationship. Edit does work. From comparing logs, it looks like the via_relation  params passed to the search controller don't have enough context to properly apply the scope.

For clarity, this is for a "join table" model in a "has many through" relationship. In my example case - I have an `AdminAccount` model that belongs to a `Person` and an `Industry`.

When creating a new admin account via the industry relation the `AdminAccountsController#new` method receives the following params: `{"via_relation"=>"industry", "via_relation_class"=>"Industry", "via_resource_id"=>"[the selected industry id]"}`, which behaves as expected.

When `searchable: true` is enabled, the params passed to `Avo::SearchController#show` are `{"q"=>"[search text]", "via_association_id"=>"person", "via_reflection_class"=>"AdminAccount", "via_reflection_id"=>"", "resource_name"=>"people"}`. This doesn't behave as expected - looks like due to the fact that the record hasn't been persisted, so it doesn't have a reference to the industry object from the parent (i.e. the new admin account).

For reference, my attach scope looks something like this: `-> { query.potential_industry_admins(parent.industry).order_by_name(:asc) }`.

Due to the multi-tenant nature of our app we have several models that are scoped like this, where the model being manipulated is scoped to some industry-related resource (the tenant config) and a user resource (which should be scoped the current tenant).

Anyways, we can probably get by without the search for now - in the example I shared, we currently have 100s of people in our system, but we'll likely be into the thousands by the end of summer and search will become much more critical.

### Steps to Reproduce
Given the following setup (simplified for readability):
```ruby
class Industry
  has_many :admin_accounts
  scope :available_to_admin, ->(person) { where.not(id: person.admin_accounts.pluck(:industry_id) }
end

class Person
  has_many :admin_accounts
  scope :potential_admins, ->(industry) { where.not(id: industry.admin_accounts.pluck(:person_id) }
end

class AdminAccount
  belongs_to :person
  belongs_to :industry
end

class IndustryResource
  field :admin_accounts, as: :has_many
end

class PersonResource
  field :admin_accounts, as: :has_many
end

class AdminAccountResource
  field :person, as: :belongs_to, searchable: true, attach_scope: -> { query.potential_admins(parent.industry) }
[...]

PR #905, opened 2022-05-16
feature: initialize the parent and attach the grandparent in attach_scope

adrianthedev:
# Description

Fixes https://github.com/avo-hq/avo/issues/904

# Checklist:

## Manual review steps

Description...

1. Step 1
1. Step 2

### DIFF
--- a/app/controllers/avo/search_controller.rb
+++ b/app/controllers/avo/search_controller.rb
@@ -62,7 +62,7 @@ module Avo
     end
 
     # Figure out if it's a belongs to search and if it has the attach_scope block present.
-    # If so, set the parent for those edit screens and the grandparent for those new screens.
+    # If so, set the parent for those edit view and the parent with the grandparent for the new view.
     def apply_attach_scope(query)
       return query if params[:via_reflection_class].blank?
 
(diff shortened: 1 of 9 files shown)


app/controllers/avo/search_controller.rb
```diff
@@ -72,6 +61,31 @@ def search_resource(resource)
...
+
+      # Try to fetch the parent.
+      if params[:via_reflection_id].present?
+        parent = params[:via_reflection_class].safe_constantize.find params[:via_reflection_id]
+      end
+
+      # Try to fetch the grandparent for the new views where the parent is nil.
+      if params[:via_parent_resource_id].present? && params[:via_parent_resource_class].present?
+        grandparent = params[:via_parent_resource_class].safe_constantize.find params[:via_parent_resource_id]
+      end
```
davidlormor, line 83: Hmmm...I wonder if it would be feasible to use `find_or_initialize_by` to pass in the association on the parent vs. introducing another variable? Would look something like:

```ruby
grandparent = params[:via_parent_resource_class].safe_constantize.find params[:via_parent_resource_id]

parent = params[:via_reflection_class].safe_constantize.find_or_initialize_by(
  "id" => params[:via_reflection_id],
  params[:via_parent_resource_relation] => grandparent
)
```

I'm not familiar enough with Avo's internals to know where the grandparent relation might come from, so not really sure on the variable naming...

Another thought might be to use [the `GlobalID` feature](https://github.com/rails/globalid/blob/main/README.md) introduced to Rails alongside ActiveJob to serialize the parent into a URI perhaps?

davidlormor (commented):
@adrianthedev this appears to introduce a breaking change when `searchable: false`. I get the following error: `Avo::Hosts::AssociationScopeHost: option 'grandparent' is required`. What's also strange is I don't even see the `attach_scope` being applied when `searchable: true` is applied...if I try to `binding.pry` in the attach scope I'm never receiving a breakpoint to inspect what's going on? Really weird, since I'm pretty sure that I was able to do it yesterday...

Also, I agree the `grandparent` naming is a bit "awkward"...see my thoughts on how to possibly avoid introducing a new variable by utilizing `find_or_initialize_by`.

davidlormor:
Also should've mentioned - thanks for looking into this! Keep up the amazing work!

app/controllers/avo/search_controller.rb
adrianthedev: Hey @davidlormor.

TL;DR: Try `2.6.1.pre.2` and use `parent.industry` in the attach scope. You'll have access to the `Industry` record.

Long version👇

In the scenario where you'd edit the `AdminAccount`, and try to change the `belongs_to` of the `person`, the parent is a record of `AdminAccount` class. So, in the same manner, when creating an `AdminAccount`, when setting the `person`, the `parent` will still have to be an `AdminAccount` instance, but there isn't one right now because you're just creating it.

I followed your suggestion and initialized the parent model `AdminAccount` with the grandparent attached. So in the attach scope you can use `parent.industry`. `parent` will be an `AdminAccount` without an ID, but with the `industry_id` attribute populated so you have all the info needed to scope the query.

I fixed the `searchable: false` bug too. Thanks for the shoutout.

davidlormor:
@adrianthedev `2.6.1.pre.2` appears to work as expected 🎉

Thanks for the quick fixes - it's going to clean up both our codebase and admin user experience significantly!

adrianthedev:
Coolio! I'm glad it's fixed. 
I'll merge and release it tomorrow. 

Thanks for the suggestions.

PR #926, opened 2022-05-27
fix: allow to put singular name on has fields

Paul-Bob:
# Description
Now, with ```singular_name: true``` it's possible to not pluralize associations. 
Example:
I have ```field :links, as: :has_many```
If I want to have singular name "Links List" for this field I only need to add  ```, name: "Links List", singular_name: true```
# Checklist:

- [ x ] I have performed a self-review of my own code

### DIFF
--- a/app/components/avo/views/resource_index_component.rb
+++ b/app/components/avo/views/resource_index_component.rb
@@ -32,7 +32,7 @@ class Avo::Views::ResourceIndexComponent < Avo::ResourceComponent
 
   def title
     if @reflection.present?
-      return field.plural_name if field.present?
+      return field.custom_name if field.present?
 
       reflection_resource.plural_name
     else
--- a/app/components/avo/views/resource_index_component.rb
+++ b/app/components/avo/views/resource_index_component.rb
@@ -32,7 +32,7 @@ class Avo::Views::ResourceIndexComponent < Avo::ResourceComponent
 
   def title
     if @reflection.present?
-      return field.custom_name if field.present?
+      return field.name if field.present?
 
       reflection_resource.plural_name
     else
--- a/app/components/avo/views/resource_index_component.rb
+++ b/app/components/avo/views/resource_index_component.rb
@@ -32,7 +32,7 @@ class Avo::Views::ResourceIndexComponent < Avo::ResourceComponent
 
   def title
     if @reflection.present?
-      return field.name if field.present?
+      return name if field.present?
 
       reflection_resource.plural_name
     else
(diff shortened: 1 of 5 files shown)


adrianthedev:
I'm curious. Doesn't `name: "Links list"` do that? That should be the whole job of the name property. 
If it doesn't change those labels, than it's a bug (by me TM) and needs to be fixed.

I think that in the `ResourceIndexCompoment`, on the `name` method, if `@reflection` is present, we should fetch the `name` off the field. 
Could you check to see if this is feasable?

Paul-Bob:
> I'm curious. Doesn't `name: "Links list"` do that? That should be the whole job of the name property. If it doesn't change those labels, than it's a bug (by me TM) and needs to be fixed.

Actually the argument `name` is always pluralize because we call `return field.plural_name` if the `@reflection.present?` and if the `field.present?`

> I think that in the `ResourceIndexCompoment`, on the `name` method, if `@reflection` is present, we should fetch the `name` off the field. Could you check to see if this is feasable?

Sure we can return field.name by default, i thanked you make it pluralized by default on porpoise :D

I'll change that behavior and this will be a bug fix and not a feature anymore

adrianthedev:
What I mean is, the field will get the `name` from the `name: "Links list"` option. Sounds like in `ResourceIndexComponent` we need to get the `field` (which is available from `Avo::BaseComponent`) and get the name from there if it's present.

Sorry, I'm AFK right now and haven't checked the source files.

app/components/avo/views/resource_index_component.rb
```diff
@@ -32,7 +32,7 @@ def initialize(
 
   def title
     if @reflection.present?
-      return field.plural_name if field.present?
+      return name if field.present?
```
adrianthedev, line 35: Can't we return `field.name` here? It seems that it fixes everything.

```ruby
      def name
      	# the default is pluralized by the declaration
        default = @id.to_s.humanize(keep_id_suffix: true)

				# returns the name if it's set
        return @name if @name.present?

				# get's it from the translation if that's set
        return t(translation_key, count: 1, default: default).capitalize if translation_key

        default
      end
```


I think that fixes it without requiring changes to the other files.


```suggestion
      return field.name if field.present?
```

Paul-Bob: That was my [first try](https://github.com/avo-hq/avo/pull/926/commits/e9f17a4611d83086d412462cddbc6f14461cf058) but unfortunately that have an issue. Imagine we have a field without a custom name but it has a translation_key configured. In this case we want to return ... `translation_key, count:2` .... by default. That's why i make that verify, if name is present return singular name, else return plural default

adrianthedev: Got it! Makes sense!

Thank you!

PR #928, opened 2022-05-29
feature: custom blank text on select filter

Paul-Bob:
# Description
Now it's possible to customize the default blank text on a select filter using the option `blank: "custom text"` or let it by default ( "-" )

# Checklist:

- [ x ] I have performed a self-review of my own code

## Manual review steps
1. Go to file `published_filter.rb` and add  `blank: "Select status"` to `def options{ ... }`
2. Go to `/admin/resources/posts` open filters and confirm if blank changed
image

### DIFF
--- a/app/views/avo/base/_select_filter.html.erb
+++ b/app/views/avo/base/_select_filter.html.erb
@@ -4,9 +4,9 @@
   data-select-filter-keep-filters-panel-open-value="<%= @resource.keep_filters_panel_open %>"
 >
   <%= filter_wrapper name: filter.name do %>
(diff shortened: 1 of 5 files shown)


adrianthedev:
What if the user actually has a `blank` option and needs to display it? 
I see there's an `empty_message` option on the `BaseFilter` already. Could we use that?

Paul-Bob:
> What if the user actually has a `blank` option and needs to display it? I see there's an `empty_message` option on the `BaseFilter` already. Could we use that?

I thanked about that after but never changed ='))
Sure we can, it's even better to use that empty_message, I'll change it today's evening

adrianthedev:
Could you also add it as a placeholder to the `TextFilter` too, so we can have it in the same PR? Thank you!

PR #933, opened 2022-06-02
Access resource.model in action visibility

marschattha:
Moving set actions after the resource is hydrated with the model allows access to the model in self.visible lambda via resource.

Also few of these lines seemed like unnecessary old code, model is available via `@model` instance variable.

# Description
Could not access the model in the visible lambda of action when determining whether an action should be shown on a particular page. With this you should be able to access it using `resource.model`.

# Checklist:

- [ x] I have performed a self-review of my own code

### DIFF
--- a/app/controllers/avo/base_controller.rb
+++ b/app/controllers/avo/base_controller.rb
@@ -75,9 +75,9 @@ module Avo
     end
 
     def show
-      set_actions
-
       @resource.hydrate(model: @model, view: :show, user: _current_user, params: params)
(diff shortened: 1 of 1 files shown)


adrianthedev:
Now you can using the `visible` block on the action


```ruby
  self.visible = -> (resource:, view: ) {
    # the resource now contains the params and all hydrated items
    puts ["resource->", resource].inspect
  }
```

adrianthedev:

PR #944, opened 2022-06-07
fix: don't open filters after filters reset + f5

Paul-Bob:
# Description
Excluding :keep_filters_panel_open param when reset filters make user able to reset the page or navigate throw table pages without having filters page open on every move

# Checklist:

- [ x] I have performed a self-review of my own code

## Manual review steps
Go to courses page, play around with filters, press reset filters and then f5. Filters page should not open

### DIFF
--- a/app/helpers/avo/url_helpers.rb
+++ b/app/helpers/avo/url_helpers.rb
@@ -7,7 +7,7 @@ module Avo
       if keep_query_params
         begin
           existing_params =
-            Addressable::URI.parse(request.fullpath).query_values.symbolize_keys
+            Addressable::URI.parse(request.fullpath).query_values.symbolize_keys.except(:keep_filters_panel_open)
         rescue
         end
       end


adrianthedev:
From what I understand, after you make some selection with the filters, when a user clicks the `reset filters` button, the filters panel should be closed, not open like it currently happens on courses.

If so (I might be wrong), I think it's the intended behavior. I mean, you already enabled the `keep_filters_panel_open` option. So it will stay open until the user closes it.

I see an unwanted behavior that after the `reset filters` button was pressed when the user refreshes the page, the panel stays open. Then, yes, that would be a bug. Or not... depends on how you look at it 😛. If the panel is set to stay open, it mostly does.

TBH, I'm not sure how to fix this. Removing the `keep_filters_panel_open` from the params isn't the best approach IMHO. That param might be used sometime in the future by us or by a developer using Avo. We'd remove it arbitrarily.

I'm open to suggestions though

Paul-Bob:
Maybe i wasn't clear enough. The only problem what I'm trying to solve is: When the filter panel it's closed and the user refresh page it should still closed, if it's open it should stay open.

Actually removing this from prams is not the best approach i agree with you but i don't have more ideas =')) it's not a big issue so maybe we can close this PR until someone come up with something better.

There are more strange behaviors to fix, for example if you set the keep_open true and change page it will close and turn it off

adrianthedev:
I agree, this is not a biggie. I'd leave it like that.

I tried the other scenario. I couldn't replicate it.

CleanShot 2022-06-11 at 17 39 20

adrianthedev:
Closing for now, but let's re-open it if we do find a bug here.

PR #965, opened 2022-06-16
Removed the custom course resource controller

iainbeeston:
# Description
The custom course resource controller (which is an example from the docs) is included in every install of avo right now. It seems odd that this would be available in every install of avo (perhaps it's needed for the docs or the example site?), so I think it should be removed. There might be a better way to do it, but this is my suggestion.

Fixes # (issue)

# Checklist:


adrianthedev:
You're right. It shouldn't be present in every install. 
I left it there as a demo. I need to move it to the dummy app. Will do that on this PR if you don't mind.

Thank you!

iainbeeston:
I thought there might be more work needed, thanks!

PR #985, opened 2022-06-22
fix: namespaced resources broken

adrianthedev:
# Description

Fixes an issue that occurred with namespace resources like `Super::Project` would have issues generating and finding routes.

# Checklist:

## Manual review steps

1. Go to the `Super project resource`
2. Make sure all CRUD operations work
3. Do the same for a resource that's not namespaced


adrianthedev:
I'm going to pause work on this PR for now. It's definitely something I want to fix in the future.

I know that not being able to namespace resources is not the best experience, but in the end, it's not a deal-breaker for using Avo.

I explained a bit [here](https://youtu.be/IAxHvhwzeOU?t=643) why this refactor is a PITA and why it takes so long.

PR #991, opened 2022-06-24
fix: hard-coded policy action called on ResourceEditComponent

bear-in-mind:
# Description
Following the recent refactoring, a little authorization bug slipped through : the `can_see_the_save_button?` would call the `:edit` action, even on `:new` action. Therefore the `save` button would not appear despite correct authorization rules.


bear-in-mind:
Mmh the failed spec doesn't seem related to my code change 🤔

PR #993, opened 2022-06-25
feature: Controller hooks

### LINKED ISSUE #986: [Feature] Provide hooks in controller actions
### Feature
Provide hooks on `create` or `update` after save to perform some actions. Here's what I'm trying to achieve:
```ruby
class Avo::EmailsController < Avo::ResourcesController
  def create
    super
    # This is not being called
    @model.send_as_reply
  end
end
```
Here's how it could work:
```ruby
class Avo::EmailsController < Avo::ResourcesController
  def create
    super do |resource|
      resource.send_as_reply
    end
  end
end
```

### Current workarounds
Right now, we can hook onto these actions through ActiveRecord callbacks, but it's not always convenient.

### Additional context
A simple and good implementation is the pattern used in the `devise` controllers, where the resource is being `yield` if a block is passed : https://github.com/heartcombo/devise/blob/f8d1ea90bc328012f178b8a6616a89b73f2546a4/app/controllers/devise/registrations_controller.rb#L20

bear-in-mind:
# Description
Provide controller hooks on `create`, `update` and `destroy` actions, to be able to easily perform some action on the resources after successful save/destroy like so:
```ruby
class Avo::UsersController < Avo::ResourcesController
  def create
    super do |record|
      record.notify("User created")
    end
  end

  def update
    super do |record|
      record.notify("User updated")
    end
  end

  def destroy
    super do
      puts "User destroyed"
    end
  end
end
```

Fixes #986 

# Checklist:

## Manual review steps
I added `create`, `update` and `destroy` methods in the `Avo::UsersController` in the dummy app. I performed some action on the yielded resource and ensured the code was executed properly.

### DIFF
--- a/app/controllers/avo/base_controller.rb
+++ b/app/controllers/avo/base_controller.rb
@@ -121,9 +121,8 @@ module Avo
     def create
       # model gets instantiated and filled in the fill_model method
       # Upon success, if a block is supplied to the inherited controller action, yield the resource to perform some actions
-      if (saved = save_model)
-        yield(@model) if block_given?
-      end
+      saved = save_model
+      yield(@model) if saved && block_given?
       @resource.hydrate(model: @model, view: :new, user: _current_user)
 
       # This means that the record has been created through another parent record and we need to attach it somehow.
@@ -172,9 +171,8 @@ module Avo
     def update
       # model gets instantiated and filled in the fill_model method
       # Upon success, if a block is supplied to the inherited controller action, yield the resource to perform some actions
-      if (saved = save_model)
-        yield(@model) if block_given?
-      end
+      saved = save_model
+      yield(@model) if saved && block_given?
       @resource = @resource.hydrate(model: @model, view: :edit, user: _current_user)
 
       respond_to do |format|
@@ -188,7 +186,8 @@ module Avo
     end
 
     def destroy
-      yield @model if (destroyed = destroy_model) && block_given?
+      destroyed = destroy_model
+      yield if destroyed && block_given?
       respond_to do |format|
         if destroyed
           format.html { redirect_to params[:referrer] || resources_path(resource: @resource, turbo_frame: params[:turbo_frame], view_type: params[:view_type]), notice: t("avo.resource_destroyed", attachment_class: @attachment_class) }
--- a/app/controllers/avo/base_controller.rb
+++ b/app/controllers/avo/base_controller.rb
@@ -120,7 +120,10 @@ module Avo
 
     def create
       # model gets instantiated and filled in the fill_model method
-      saved = save_model
+      # Upon success, if a block is supplied to the inherited controller action, yield the resource to perform some actions
+      if (saved = save_model)
+        yield(@model) if block_given?
+      end
       @resource.hydrate(model: @model, view: :new, user: _current_user)
 
       # This means that the record has been created through another parent record and we need to attach it somehow.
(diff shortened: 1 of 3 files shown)


adrianthedev:
This is an interesting take on the matter (having a block to do things in).

Let me test it out for a few days please and get back to you.

adrianthedev:
Hey man, look what an interesting idea here https://github.com/avo-hq/avo/issues/1008. So cool to have this comment so close to your proposal.

Let me think a bit more about what else we could add to make this a more complete feature and tackle it in the next release. WDYT?

adrianthedev:
Closed by mistake 🤦‍♂️

bear-in-mind:
Sure thing !

dhnaranjo:
It seems like the behavior added here can be implemented with something like:

```
class Avo::UsersController < Avo::ResourcesController
  after_action -> { @model.notify("User created" }, only: :create
  after_action -> { @model.notify("User updated" }, only: :update
  after_action -> { puts "User destroyed }, only: :destroy
end
```

Maybe just some clearer documentation around how Avo controllers work so folks know they can access the model n instance resources directly??

bear-in-mind:
Well no, the whole point is to only trigger action upon success of the request. With the implementation you propose, callbacks would be triggered even in case of failure

dhnaranjo:
Aright, I missed that, but I think the API proposed might not be sufficiently expressive, or I wouldn't have. Maybe an `after_success_action`, which suggests an `after_failure_action` as well. It doesn't need to be in this controller action callback style, but having the action yield to a block doesn't really communicate WHEN it will yield. It'd be completely valid for it to yield on failure, or in all cases, nah mean?

bear-in-mind:
That's very true ! I'm used to using it this way with `devise`, but I agree that it doesn't convey much meaning as to when it occurs. @adrianthedev what do you think ? Would you prefer it refactored as a more meaningful callback ?

adrianthedev:
Thank you for brainstorming this here in public! I appreciate your support!

The way I approached most features in Avo is with multiple iterations. It seems to me to be the gentlest way of building advanced features. It's better than building the whole thing and finding out that the users might not need it or might use it differently.

So there are a few things that I see need achieving here and in #1008 (pls let me know if I missed any):

1. Customize the message after save/create/update/delete
2. Customize the path where the user is redirected after
3. Add hooks to run arbitrary code within those methods. 

Regarding number 3

1. Should the hooks receive data like the `@model`/`@resource`/`params`? My thoughts are yes. You need the context to do the action. You will have access to them if they are regular methods inside the controller.
2. Should the hooks be allowed to change that data? This is tricky, but probably yes. If they are methods inside the controller, they will be able to do that.
3. Should the hooks be before and/or after the destructive action (save/create/update/delete) happened? I'd say before and after. This will bring overhead for us on maintenance and some overhead for support (will always have to ask if those methods are overwritten, etc.), but that will bring on a lot of flexibility for Avo developers.

So I'd use the following approach:

- [ ] Create a PR for the first issue (Customize the message) similar to the devise approach.
- [ ] Create a PR for the second issue (Customize the path) similar to the devise approach. Must be careful with the [current way](https://docs.avohq.io/2.0/resources.html#customize-what-happens-after-record-is-created-edited) of doing things.
- [ ] Create a PR for the third issue after we answer those questions.

adrianthedev:
I talked with @Paul-Bob, and we're going to prioritize this feature like so:

- We're going to implement items 1. and 2. similar to how devise implemented their callbacks (overridable methods)
- We're going to add a response interceptor so the user can respond in their own way. Example below.
- We'll document how you can already run before & after hooks with Rails' built-ins (`before_action`, `after_action`).

```ruby
      if before_response.presenet?

      else
        respond_to do |format|
          if saved
            format.html { redirect_to after_create_path, notice: "#{@resource.name} #{t("avo.was_successfully_created")}." }
          else
            flash.now[:error] = t "avo.you_missed_something_check_form"
            format.html { render :new, status: :unprocessable_entity }
          end
        end
      end
```

adrianthedev:
Closing this in favor of https://github.com/avo-hq/avo/pull/1139

Thank you for spearheading this initiative!

bear-in-mind:
This is great, thank you guys !

Issue #994, opened 2022-06-25
Using custom `model_class` with classes using Rails 7 encryption fails


damianlegawiec:
### Describe the bug
When using custom `model_class` (as described here: https://docs.avohq.io/2.0/resources.html#custom-model-class) with a class with encrypted attributes, eg

```ruby
encrypts :name, deterministic: true, ignore_case: true
encrypts :email, deterministic: true, downcase: true
```

it fails on test/production environment (with class caching on):

as it tries to connect to the DB during class autoloading

### Expected behavior
What I would suggest is to have the `model_class` as a string and use `safe_constantize` when needed to access the class itself. 

### System configuration
**Avo version**: 2.9

**Rails version**: 7.0

**Ruby version**: 3.0

**License type** Community

adrianthedev:
Hey @damianlegawiec. Thanks for getting back to me with this. 
I haven't used encrypted attributes with Avo, but that makes sense. 

I'm AFK now, but I'll push out a pre-release tomorrow with that fix applied so you can test it too.

damianlegawiec:
@adrianthedev thanks for the quick reply. Take your time, no pressure!

adrianthedev:
Hey @damianlegawiec I just published [`2.9.1.pre1`](https://rubygems.org/gems/avo/versions/2.9.1.pre1).


I wasn't able to replicate the bug on my end (the app booted ok), but there might be differences in configuration.

 Please try it out and let me know if that fixes the issue.

damianlegawiec:
Thanks @adrianthedev , that indeed helped, I've found one more issue and created a PR with a fix - https://github.com/avo-hq/avo/pull/997

adrianthedev:
Hey @damianlegawiec could you test [2.9.1.pre2](https://rubygems.org/gems/avo/versions/2.9.1.pre2) as well when you have the time. I did a tiny refactor and want to make sure it still supports your use case.

I'll merge that PR and come back to it before release if necessary.

PR #997, opened 2022-06-26
Ensure we're autoloading rails app in development only

damianlegawiec:
Previous check also included environments such as staging or test which wasn't intended

# Description

Fixes https://github.com/avo-hq/avo/issues/994

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/lib/avo/dynamic_router.rb
+++ b/lib/avo/dynamic_router.rb
@@ -1,7 +1,7 @@
 module Avo
   module DynamicRouter
     def self.routes(router)
-      Rails.application.eager_load! unless Rails.env.production?
+      Rails.application.eager_load! if Rails.env.development?
 
       BaseResource.descendants
         .select do |resource|


adrianthedev:
TBH, I don't think that will help.

I'm triggering the eager load to pick up on all the descendants of `Base` classes like `BaseResource` (`class UserResource < Avo::BaseResource`). Then, after it identifies the descendants, it can generate the routes for those resources. I placed that condition because I know the eager loading happens in `production` by default.

If we don't eager_load them, the routes aren't available and Avo won't work. Which is what we see in the failed tests.

I'm curious, what problem does this PR fix for you. What error do you get before this change?

adrianthedev:
I see what you mean. Not sure how to handle this.

My recommendation is to look more into this "rails encryption fails to create database" or something similar. My gut is telling me that this is something related to your particular setup and there should be something that you could do to mitigate it. 

Are you encrypting the credentials for the database somehow? I don't see how it could get to Avo from the `db:create` method.

I tried it with a few apps where I added `Rails.application.eager_load!` in `application.rb` and the `db:setup` process went on fine. I don't think it's an Avo issue TBH. I'm happy to go on a call next week and troubleshoot this together.

damianlegawiec:
@adrianthedev so we kinda bypassed this by explicitly adding avo to development/production gemfile group, to avoid loading avo code in test env

PR #1049, opened 2022-07-13
Fix Resource Search when Overriding Model

rickychilcott:
# Description
Use Resource's `model_key` when generating search url

# Checklist:

## Manual review steps
Works locally with a namespaced model class, in addition to normal resources.

`self.model_class = SalesList::Organization`

### DIFF
--- a/app/components/avo/views/resource_index_component.html.erb
+++ b/app/components/avo/views/resource_index_component.html.erb
@@ -34,7 +34,7 @@
       >
         <% if @resource.search_query.present? %>
           <div class="flex items-center px-4 w-64">
-            <%= render partial: 'avo/partials/resource_search', locals: {resource: @resource.model_name.collection} %>
+            <%= render partial: 'avo/partials/resource_search', locals: {resource: @resource.model_key} %>
           </div>
         <% else %>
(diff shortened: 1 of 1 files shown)


rickychilcott:
I know #985 is happening, but It's not clear to me that that particular PR will fix this issue.

adrianthedev:
I think this change requires a bit more attention. I made this change https://github.com/avo-hq/avo/pull/936/files#diff-9b73f266a2b6d218a572aeb5bc7f34c546c74dba11346830d91a0c24ba20bf26R34 a week ago. 

I'll have a look and make sure it's fixed in the next release.

rickychilcott:
Sounds good. Search for a resource with a namespaced model is broken, and likely is broken for ALL overridden model classes.

The code you referenced in https://github.com/avo-hq/avo/pull/936/files#diff-9b73f266a2b6d218a572aeb5bc7f34c546c74dba11346830d91a0c24ba20bf26R34 only works if the model class is the same name as the Resource (i.e. `UserResource` resource and `User` model)

adrianthedev:
Yes. I understand this is pressing and I'm on it!

rickychilcott:
same as #1050

Issue #1009, opened 2022-07-01
Add an option to hide the pagination information for few records


adrianthedev:
### Feature

What I would do is to set a `hide_pagination_when_one_page_present` (I'm open to a better name) option on the `has_many` field and pick thst up in the pagination [component](app/components/avo/paginator_component.html.erb). If that's turned on in the pagination component, hide the information You don't need. 

I believe you can get that option from the `@field` instance variable in the `ResourceIndexComponent`. You can check in the pagination component before and see if it's available there as well. 

The `@field` variable will have that option `hide_pagination_when_one_page_present`. 

From here: https://discord.com/channels/740892036978442260/740893011994738751/992484340057509989

> Is it possible to hide pagination for associations on a Resources#show page? I have a has_many association on a model, but it is usually only 1-2 associations. I don't need it to show the "8 per page" or "Displaying 2 items".

<img width="721" alt="CleanShot 2022-07-01 at 22 44 41@2x" src="https://user-images.githubusercontent.com/1334409/176959707-98b5d333-6916-410b-886e-eb0330b37a57.png">

PR #1106, opened 2022-07-26
improvement/hide_pagination_when_few_items

Paul-Bob:
# Description

Allow to hide pagination elements when only one page is available

<details>
<summary>field :comments, as: :has_many</summary>

</details>

<details>
<summary> field :comments, as: :has_many, discreet_pagination: true</summary>

</details>

Fixes # (issue)
https://github.com/avo-hq/avo/issues/1009

# Checklist:


app/components/avo/views/resource_index_component.html.erb
```diff
@@ -63,7 +63,7 @@
       <% if view_type.to_sym == :table %>
         <% if @models.present? %>
           <div class="mt-4">
-            <%= render Avo::PaginatorComponent.new pagy: @pagy, turbo_frame: @turbo_frame || 'none', index_params: @index_params, resource: @resource, parent_model: @parent_model %>
+            <%= render Avo::PaginatorComponent.new pagy: @pagy, turbo_frame: @turbo_frame || 'none', index_params: @index_params, resource: @resource, parent_model: @parent_model, field: field %>
```
adrianthedev, line 66: Would it be a bit cleaner to pass `discreet_pagination` to the component?
The pagination component shouldn't really know about the `@field`
```suggestion
            <%= render Avo::PaginatorComponent.new pagy: @pagy, turbo_frame: @turbo_frame || 'none', index_params: @index_params, resource: @resource, parent_model: @parent_model, discreet_pagination: true/false %>
```

Paul-Bob: make sense, already changed it

adrianthedev (commented):
The PR looks perfect!

Do you mind adding in a test or two? We could add the `discreet_pagination` option to the `teams` field on the `UserResource` (open to suggestions).
Then write a spec that would check for when there is one page and when there are more.

Paul-Bob:
For some odd reason i had some struggle creating a spec for teams and users, so I opted to make it for courses and links.

spec/system/avo/discreet_pagination_spec.rb
```diff
@@ -0,0 +1,32 @@
+require "rails_helper"
+
+RSpec.describe "DiscreetPagination", type: :system do
```
adrianthedev, line 3: Can we make this a `feature` test? This will make it way faster.
Sorry I haven't mentioned it before.

System tests will not work with this approach because Turbo is loading the frame in the browser.
In order to test it as a `feature` test, we should load the actual frame and test our assertion there. Something similar to [this](https://github.com/avo-hq/avo/blob/main/spec/requests/avo/authorization/associations_spec.rb#L14).

Paul-Bob: Sure! System type took 4 seconds on this test and feature type only 1 second, great improvement!

PR #1108, opened 2022-07-28
Support `date` and `date_time` to clear the value

rickychilcott:
# Description
While updating the date/date_time picking implementation, we lost the ability to set the value to empty.

Previously, there was an avo date_field_controller.js exception that was raise that would bail out of the `onChange` when the input field was empty.

-  [x] I have performed a self-review of my own code

## Manual review steps

1. Utilize a `date` or `date_time` field
2. Manually clear the input field
3. Check it works :)

### DIFF
--- a/app/controllers/avo/base_controller.rb
+++ b/app/controllers/avo/base_controller.rb
@@ -254,18 +254,19 @@ module Avo
     def cast_nullable(params)
       fields = @resource.get_field_definitions
 
-      nullable_fields = fields.filter do |field|
-        field.nullable
-      end
+      nullable_fields = fields
+        .filter do |field|
+          field.nullable
+        end
         .map do |field|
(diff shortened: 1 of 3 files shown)


app/controllers/avo/base_controller.rb
```diff
@@ -254,18 +254,19 @@ def permitted_params
...
+          [field.id, field.null_values]
+        end
         .to_h
 
       params.each do |key, value|
-        nullable = nullable_fields[key.to_sym]
+        nullable_values = nullable_fields[key.to_sym]
 
-        if nullable.present? && value.in?(nullable)
+        if nullable_values.present? && value.in?(nullable_values)
```
adrianthedev, line 269: minor formatting here

adrianthedev:
Thank you for the fix @rickychilcott

Issue #1103, opened 2022-07-26
Add the resource option to has_many/has_and_bleogns_to_many associations


adrianthedev:
### Feature

Example: 

`field :comments, as: :has_many, resource: CommentResource`

the `resource` option can take the resource class or a string that can later be constantized.

Example: `ModelClassConstantized`

PR #1109, opened 2022-07-28
improvement/allow_to_specify_resource_on_has_many_fields

Paul-Bob:
# Description

By default, `has_many` will use the field resource. 

For example `field :comments, as: :has_many` will use `CommentResource`.

If you want to use a custom resource you need to add the **`use_resource: YourCustomResource`** option.

Make sure to have the custom resource prepared to be used by adding `self.model_class = ::YourModel`

Fixes https://github.com/avo-hq/avo/issues/1103

# Checklist:


app/controllers/avo/application_controller.rb
```diff
@@ -207,6 +207,12 @@ def resource
     end
 
     def related_resource
+      # Find the field from the parent resource
+      field = @resource.get_field params[:related_name]
+
+      # Get the related resource from the field
+      return field.resource if field.present?
```
adrianthedev, line 214: This will set the proper resource for the association view.

spec/dummy/config/initializers/avo.rb
```diff
@@ -59,6 +59,7 @@
       group "Blog", collapsable: true do
         resource :posts
         resource :comments
+        resource :tests
```
adrianthedev, line 62: Do we need this to be visible?

spec/dummy/app/avo/resources/test_resource.rb
```diff
@@ -0,0 +1,11 @@
+class TestResource < Avo::BaseResource
```
adrianthedev, line 1: Can we change the name to something more expressive like `TinyComment` (I'm open to something else)?

adrianthedev (changes requested):
I pushed a few changes:

 - changed the `resource` field option to `use_resource`. I did that because in the `BaseField` class we already have a `@resource` instance variable and we might get some clashes
 - I changed `related_resource` in `ApplicationController` to take into account the `use_resource` option from the field. That's why you couldn't see the changes from the `HasBaseField`

Next up:

- [ ] we need to make sure the links in the turbo frame take the users to the proper locations.
For example, when you click on the `Attach comment` button it takes you to a page where that new resource is used. Same for the rest of the buttons and links.
- [ ] make sure the page is translatable using the translation tags
- [ ] other misc things like search, attach/detach/create

CleanShot 2022-07-29 at 12 20 09@2x

spec/dummy/config/initializers/avo.rb
Paul-Bob: no, already deleted

spec/dummy/app/avo/resources/test_resource.rb
Paul-Bob: changed to PhotoComment because now have a photo attached

adrianthedev:
Getting back to you with some things that were "broken"

- the search is not working because it's not using the proper resource. `Comment` does not have `search_query` and `PhotoComment` has it. But it seems to be working on my end (screenrec below)
- the pagination seems to be working too

Let me know if I'm looking in the wrong place or I'm not reproducing the bugs properly.


CleanShot 2022-07-31 at 20 16 23


CleanShot 2022-07-31 at 20 14 14

lib/avo/fields/base_field.rb
```diff
@@ -256,6 +256,14 @@ def model_or_class(model)
           "model"
         end
       end
+
+      def name_from_id
+        @id.to_s.humanize(keep_id_suffix: true)
+      end
+
+      def name_from_use_resource
```
adrianthedev, line 264: `use_resource` is not a `BaseField` property. Can we create a `name` method in `HasBaseField` that checks for `use_resource` and uses that or falls back to `super`?

lib/avo/fields/has_base_field.rb
```diff
@@ -5,6 +5,7 @@ class HasBaseField < BaseField
       attr_accessor :scope
       attr_accessor :attach_scope
       attr_accessor :description
+      attr_accessor :use_resource
```
adrianthedev, line 8: We have a method that retrieves the instance so this is not used anymore.
```suggestion
```

lib/avo/fields/has_base_field.rb
```diff
@@ -15,19 +16,24 @@ def initialize(id, **args, &block)
...
+        App.get_resource @use_resource
+      end
+
       def resource
         Avo::App.get_resource_by_model_name @model.class
       end
 
       def turbo_frame
-        "#{self.class.name.demodulize.to_s.underscore}_#{display}_#{id}"
+        "#{self.class.name.demodulize.to_s.underscore}_#{display}_#{frame_id}"
```
adrianthedev, line 36: GG!

spec/dummy/app/avo/resources/post_resource.rb
```diff
@@ -34,7 +34,7 @@ class PostResource < Avo::BaseResource
   end
   field :user, as: :belongs_to, placeholder: "—"
   field :status, as: :select, enum: ::Post.statuses, display_value: false
-  field :comments, as: :has_many
+  field :comments, as: :has_many, use_resource: PhotoCommentResource, searchable: true
```
adrianthedev, line 37: Let's make this unsearchable please.
```suggestion
  field :comments, as: :has_many, use_resource: PhotoCommentResource
```

spec/dummy/app/models/comment.rb
```diff
@@ -15,6 +15,9 @@ class Comment < ApplicationRecord
   validates :body, presence: true
   validate :body_different
 
+  has_one_attached :photo
+  # acts_as_list
```
adrianthedev, line 19:
```suggestion
```

adrianthedev (commented):
Some minor tweaks here and there.

- [ ] docs needed too

lib/avo/fields/has_base_field.rb
Paul-Bob: easy! :D

spec/dummy/app/avo/resources/post_resource.rb
Paul-Bob: agree

lib/avo/fields/base_field.rb
Paul-Bob: That approach is kinda tricky because we always want to give priority to custom name or to localization... We can have the name method overridden in HasBaseField but there we need to check for custom name and translation too so that would be a duplicated method. Check my last approach for name

lib/avo/fields/has_base_field.rb
```diff
@@ -91,6 +96,16 @@ def authorized?
           true
         end
       end
+
+      def name
+        super(from_use_resource: use_resource&.name)
+      end
```
adrianthedev, line 102: I was thinking of a different route where we didn't have to introduce `from_use_resource` to the `BaseField.name` method.

```suggestion
      def name
        return use_resource&.name if use_resource.present?
      
        super
      end
```

adrianthedev (changes requested):
There's one thing I'd like us not to push with this PR. The `name(from_use_resource: nil)` override in `BaseField`.

Also, are all items from [this](https://github.com/avo-hq/avo/pull/1109#pullrequestreview-1055252069) comment checked?

Otherwise, we're good to go! Good job!

lib/avo/fields/has_base_field.rb
Paul-Bob: That approach have an issue, if use resource is present the `name: "CustomName"` and translation will be unconsidered. I don't think we want it, we want the name to have the flowing priority:

1. Custom name, using name option
2. Translation keys
3. Name from use resource if present, otherwise
4. Name from id

I understand that my first approach and second one are kinda tricky but that was the ways i found to maintain this naming priority to be the same without duplicating the code and the guards

Paul-Bob:
> Next up:
> 
>     * [ ]  we need to make sure the links in the turbo frame take the users to the proper locations.
>       For example, when you click on the `Attach comment` button it takes you to a page where that new resource is used. Same for the rest of the buttons and links.

✅ HalfCheked, it keeps the "old" path but actually use the new resource so it's fine

>     * [ ]  make sure the page is translatable using the translation tags
Can you test this one please?

>     * [ ]  other misc things like search, attach/detach/create
✔️   Checked, attach and create are even in the spec

lib/avo/fields/has_base_field.rb
Paul-Bob: Check the last approach based on what you suggested

adrianthedev (approved):
Perfect!

Thank you!

Amazing job!

Issue #1070, opened 2022-07-19
Scoped search in has_many associations


adrianthedev:
### Feature

From [this](https://discord.com/channels/740892036978442260/740893011994738751/998947654560141362) thread. 

This PR still needs a bit of research to make sure it's feasible to automatic usage of `search_query` on the parent resource.

### To do

- [ ] Avo should have an easy way to scope out searches for `has_many` associations.
- [ ] The search box should be hidden in the association screen if the search hasn't been configured.

### Approach

- The resource index component will detect if it has a `reflection` present and use a different route for search
- Let's create a new nested route similar to `get "/:resource_name/search", to: "search#show"` that should catch those requests
- in the `SearchController` detect if it's from a nested resource and apply the search on the scoped

### Current workarounds

Use a text filter.

PR #1122, opened 2022-08-04
feature/scoped_search_in_has_many_associations

Paul-Bob:
# Description

Avo now only shows parent's records in a has_many association search results, but if the associated resource have no search_query configured the search box don't even appear. 

Fixes https://github.com/avo-hq/avo/issues/1070

# Checklist:

# Breaking change

We made some improvements on how we run the search_query. It's more flexible now and future-proof. In order to improve that process we introduced a breaking change and removed the params keyword.

**Before**
```ruby
self.search_query = ->(params:) do
  scope.ransack(id_eq: params[:q], m: "or").result(distinct: false)
end
```

**After**
```ruby
self.search_query = -> do
  scope.ransack(id_eq: params[:q], m: "or").result(distinct: false)
end
```

To make it easier for you to migrate, we made this ruby script
```ruby
Dont_touch = ['.', '..', $0]
Old_search_query = "self.search_query = ->(params:) do"
New_search_query = "self.search_query = -> do"

def remove_params_keyword(file_name)
  content = File.read file_name
  content.gsub!(Old_search_query, New_search_query)

  File.open(file_name, "w") { |file| file << content }
end

Dir.foreach(".") {|file_name| remove_params_keyword file_name unless Dont_touch.include? file_name}

File.delete($0)

```

**Usage**
- Create a ruby file on your **resources folder** (ex: `remove_params_keyword.rb`)
- Copy the above script and paste it on that file.
- Execute the script: `$ ruby script_name.rb` (ex: `$ ruby remove_params_keyword.rb`)

**Expected behavior**
The script should replace all `self.search_query = ->(params:) do` with `self.search_query = -> do` and then remove himself.


app/controllers/avo/search_controller.rb
```diff
@@ -67,35 +67,52 @@ def search_resource(resource)
...
+    end
+
+    def apply_has_many_attach_scope(query)
+      # Try to fetch the parent.
+      fetch_parent
+
+      # Fetch the field
+      field = fetch_field
+
+      # please don't judge the next to lines, =')) it is work in progress, i'm testing some use cases.
```
Paul-Bob, line 109: two*

app/controllers/avo/search_controller.rb
```diff
@@ -67,35 +67,52 @@ def search_resource(resource)
...
+      # we need to know the relation with the inverse_of reflection here in order to know how to handle the query
```
Paul-Bob, line 110: we need to know if the inverse relation of this field is also a has_many or a has_one/belongs_to , we also need to know the foreign type in case of polymorphic association

adrianthedev (changes requested):
The search works as expected, so good job on that!

Let's take this opportunity to refactor this controller a bit and make it easier to scan.
It's my fault to have piled on unnecessary functionality in the `apply_attach_scope` in the past.
I left some comments on how to split that method up and apply the scopes as we need them in different methods. Separate methods for each scope.

app/components/avo/views/resource_index_component.rb
```diff
@@ -153,4 +153,22 @@ def name
...
+        id: @parent_model.id
+      }
+    else
+      {
+        association: '',
+        association_id: '',
+        class: '',
+        id: ''
+      }
+    end
```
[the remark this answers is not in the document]
Paul-Bob: sure, that's fixed

app/components/avo/views/resource_index_component.rb
```diff
@@ -153,4 +153,22 @@ def name
...
   end
 
+  def via_reflection
+    if @reflection.present?
+      {
+        association: 'has_many',
+        association_id: @reflection.name,
+        class: reflection_model_class,
+        id: @parent_model.id
+      }
```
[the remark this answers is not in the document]
Paul-Bob: i think we need, let me know if you find one we don't use or we can omit

app/controllers/avo/search_controller.rb
```diff
@@ -44,7 +44,11 @@ def search_resources(resources)
...
     def search_resource(resource)
-      query = resource.search_query.call(params: params).limit(8)
+      query =  Avo::Hosts::SearchScopeHost.new(
+        block: resource.search_query,
+        params: params,
+        scope: resource.class.scope
+      ).handle
 
       # Figure out if this is a belongs_to search
       query = apply_attach_scope query
```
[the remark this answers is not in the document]
Paul-Bob: i tried to make that logic flow readable

app/controllers/avo/search_controller.rb
```diff
@@ -67,35 +71,40 @@ def search_resource(resource)
...
+      fetch_parent
+
+      if should_apply_has_many_scope?
+        apply_has_many_scope(query)
+      elsif should_apply_attach_scope?
+        apply_attach_scope(query)
       end
+    end
 
+    # Set the parent for those edit view and the parent with the grandparent for the new view.
```
adrianthedev, line 86: This comment does not reflect what the method does.

app/controllers/avo/search_controller.rb
```diff
@@ -118,9 +127,36 @@ def apply_search_metadata(models, avo_resource)
       end
     end
 
-    def belongs_to_field
+
+    private
+
+    def fetch_parent
```
adrianthedev, line 133: Maybe `set_parent` is a better name for this method? Because it sets the `@parent` instance var.

app/controllers/avo/search_controller.rb
```diff
@@ -118,9 +127,36 @@ def apply_search_metadata(models, avo_resource)
...
+        @parent = params[:via_reflection_class].safe_constantize.find params[:via_reflection_id]
+      end
+    end
+
+    def should_apply_has_many_scope?
+      @should_apply_has_many_scope ||= params[:via_association] == 'has_many' && @resource.search_query.present?
+    end
+    
+    def should_apply_attach_scope?
+      @should_apply_attach_scope ||= params[:via_association] == 'belongs_to' && field_have_attach_scope?
```
adrianthedev, line 144:
```suggestion
      params[:via_association] == 'belongs_to' && field_have_attach_scope?
```

app/controllers/avo/search_controller.rb
```diff
@@ -118,9 +127,36 @@ def apply_search_metadata(models, avo_resource)
...
+      @should_apply_any_scope ||= should_apply_has_many_scope? || should_apply_attach_scope?
+    end
+
+    def field_have_attach_scope?
+      fetch_field
+      @attach_scope = @field.attach_scope
+      @attach_scope.present?
+    end
+
+    def fetch_field
```
adrianthedev, line 157:
```suggestion
    def set_field
```

app/javascript/js/controllers/search_controller.js
```diff
@@ -100,6 +106,19 @@ export default class extends Controller {
         via_relation: this.dataset.viaRelation,
       }
     }
+    else if (this.isHasManyToSearch) {
```
adrianthedev, line 109:
```suggestion
    } else if (this.isHasManyToSearch) {
```

app/controllers/avo/search_controller.rb
```diff
@@ -118,9 +127,36 @@ def apply_search_metadata(models, avo_resource)
...
+    def fetch_parent
+      if params[:via_reflection_id].present?
+        @parent = params[:via_reflection_class].safe_constantize.find params[:via_reflection_id]
+      end
+    end
+
+    def should_apply_has_many_scope?
+      @should_apply_has_many_scope ||= params[:via_association] == 'has_many' && @resource.search_query.present?
+    end
+
```
adrianthedev, line 142:
```suggestion

```

app/controllers/avo/search_controller.rb
```diff
@@ -118,9 +127,36 @@ def apply_search_metadata(models, avo_resource)
       end
     end
 
-    def belongs_to_field
+
```
adrianthedev, line 130:
```suggestion
```

adrianthedev (commented):
The controller looks much better! Good job!

I left a few nitpicks and style tweaks.
If we refactor the tests, we're good to merge.

app/controllers/avo/search_controller.rb
```diff
@@ -118,9 +127,36 @@ def apply_search_metadata(models, avo_resource)
...
+    private
+
+    def fetch_parent
+      if params[:via_reflection_id].present?
+        @parent = params[:via_reflection_class].safe_constantize.find params[:via_reflection_id]
+      end
+    end
+
+    def should_apply_has_many_scope?
+      @should_apply_has_many_scope ||= params[:via_association] == 'has_many' && @resource.search_query.present?
```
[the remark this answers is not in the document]
Paul-Bob: no we don't, actually that is for time optimization, first time we use the method he "does the math" and assign the result to the variable, second time the result is returned from the variable and the math is not done twice...  Maybe here it's not worth it, should we take it out?

app/controllers/avo/search_controller.rb
Paul-Bob: Here i think we should maintain the var because we don't want the field to get fetched more than once

app/controllers/avo/search_controller.rb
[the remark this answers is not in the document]
adrianthedev: I think we don't need it now

app/controllers/avo/search_controller.rb
adrianthedev: ok. let's keep it

app/controllers/avo/search_controller.rb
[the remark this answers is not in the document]
adrianthedev: actually, if we kept it on the other one, let's keep it here too. for more familiarity

app/controllers/avo/search_controller.rb
```diff
@@ -67,22 +70,20 @@ def search_resource(resource)
...
+    # This is also used when looking for `belongs_to` associations, and this method applies the parents `attach_scope` if present.
+    def apply_scope(query)
+      if should_apply_has_many_scope?
+        apply_has_many_scope
+      elsif should_apply_attach_scope?
+        apply_attach_scope(query, parent)
       end
+    end
 
+
```
adrianthedev, line 83:
```suggestion
```

app/controllers/avo/search_controller.rb
```diff
@@ -118,9 +124,41 @@ def apply_search_metadata(models, avo_resource)
       end
     end
 
-    def belongs_to_field
+    private
+
+    def should_apply_has_many_scope?
+      params[:via_association] == 'has_many' && @resource.search_query.present?
```
adrianthedev, line 130:
```suggestion
      params[:via_association] == "has_many" && @resource.search_query.present?
```

app/controllers/avo/search_controller.rb
```diff
@@ -118,9 +124,41 @@ def apply_search_metadata(models, avo_resource)
...
+    end
+
+    def should_apply_attach_scope?
+      params[:via_association] == 'belongs_to' && attach_scope.present?
```
adrianthedev, line 134:
```suggestion
      params[:via_association] == "belongs_to" && attach_scope.present?
```

adrianthedev (approved):
There's just one thing I'd like us to try and rename. It's doing some weird `true`/`false` switching logic.

If we fix that, we're good to merge!

lib/avo/fields/has_base_field.rb
```diff
@@ -101,6 +102,10 @@ def default_name
         use_resource&.name || super
       end
 
+      def show_search_box?
+        !@hide_search_box
+      end
```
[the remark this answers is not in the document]
Paul-Bob: logic refactored

PR #1127, opened 2022-08-09
Fix sidebar default collapsed state

OlexYakov:
# Description
The deafault `collapsed` option for menu sections was not working. The default state is correctly set, but incorrectly read in the Stimulus controller, the bug was introduced in #941.

Basically, on the first load (and if the user never interacted with the collapse feature) the `this.userState` will return `null`, because there is no such key in the local storage. `this.userState === 'collapsed'` will always return false if the user never toggled the collapse feature, making the default option useless.

The fix is using the defaultState if userState is unefined/null.

# Checklist:

## Manual review steps

1. Create a section with  `collapsable: true, collapsed: true`
2. Delete local storage (clean cookies) to ensure there are no user preferences overriding the default
2. Check if the section appears collapsed.

### DIFF
--- a/app/javascript/js/controllers/menu_controller.js
+++ b/app/javascript/js/controllers/menu_controller.js
@@ -22,8 +22,8 @@ export default class extends Controller {
   }
 
   get initiallyCollapsed() {
-    if (this.defaultState === 'collapsed') {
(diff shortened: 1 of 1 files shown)


adrianthedev:
Thanks for the fix!

PR #1151, opened 2022-08-18
feature: Utilize required attribute in HTML to have browser-side requires

rickychilcott:
# Description
Use the HTML attribute of `required` and sync it up with the `required` status on the field so that the client-side knows it's required therefore disallowing the browser to submit the form.

This also adds a custom matcher for `have_required_field` and cleans up the alphabetical ordering of some keys in the HTML.

With support from @ncsmith24

# Checklist:

### DIFF
--- a/app/components/avo/fields/boolean_field/edit_component.html.erb
+++ b/app/components/avo/fields/boolean_field/edit_component.html.erb
@@ -5,6 +5,7 @@
       class: "text-lg h-4 w-4 #{@field.get_html(:classes, view: view, element: :input)}",
       data: @field.get_html(:data, view: view, element: :input),
       disabled: @field.readonly,
+      required: @field.is_required?,
       style: @field.get_html(:style, view: view, element: :input)
     %>
   </div>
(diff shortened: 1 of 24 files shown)


adrianthedev:
This looks good @rickychilcott! Let me have a go through the fields tomorrow and merge it.

Thanks for the alphabetical check and `have_required_field`.

adrianthedev:
I had a hunch there was a reason why client side validation was not added.

With the tabs feature, I added [`novalidate: true`](https://github.com/avo-hq/avo/blob/main/app/components/avo/views/resource_edit_component.html.erb#L14) on the form. 
The reason behind this is that when you have an input that is hidden on the `Edit` view, you get an error from the browser that the field is hidden.
I updated the user resource on this PR as an example.

Using `novalidate: true` allows for the field to be hidden but removes the client-side validation. I figured it was ok because we have server-side validation.

CleanShot 2022-08-19 at 18 35 17

I'm not sure how to fix this issue to allow for both hidden fields and client-side validation.

rickychilcott:
Not a problem. I'll close this PR and have a think about it.

I'm kinda in a weird space with _needing/wanting_ this because I have a very non-standard use case. I have a non-AR backed model that has a KeyValue field and occasionally the person who uses it doesn't submit a valid form. I'll have to investigate if I can get server-side validations to properly work with my non-AR backed form and then this PR will be moot.

That said, it does feel a little weird to not have at least basic client-side validations, but I understand the complexities here.

Thank you for reviewing!

adrianthedev:
Anytime @rickychilcott 👍 

I'm curious about the non-AR resource. How did you implement it on the "model" side? Can you share a snippet?

Issue #1119, opened 2022-08-01
Avo crashes if the validator fails and the user attached a file


adrianthedev:
### Describe the bug

It only happens sometimes.

### Steps to Reproduce and brake
1. Go to projects
2. click on create project
3. click attach a file and don't fill in any fields
4. click on submit and see it crash

<details>
<summary>Fails with</summary>

```
No route matches {:action=>"show", :controller=>"avo/projects", :id=>#<Project id: nil, name: "", status: nil, stage: "Discovery", budget: nil, country: "", users_required: nil, started_at: nil, meta: nil, created_at: nil, updated_at: nil, description: "", progress: 50>}, possible unmatched constraints: [:id]
Did you mean?  resources_project_url
               resources_projects_url
               resources_projects_path
               new_resources_project_url

```
![CleanShot 2022-08-01 at 22 41 17](https://user-images.githubusercontent.com/1334409/182231943-4523f182-647a-43fb-9afe-fe3380aadf0b.gif)

![CleanShot 2022-08-01 at 22 42 14](https://user-images.githubusercontent.com/1334409/182232005-73d10d2f-b85c-4203-b6e4-d5043a6f90a1.jpg)

</details>

### Steps to Reproduce and succeed
1. Go to a user
2. click create new user
3. select a file to upload and don't fill any other fields
4. click submit
5. observe it's not crashing

<details>
<summary>Details</summary>

![CleanShot 2022-08-01 at 22 42 44](https://user-images.githubusercontent.com/1334409/182232145-205d4785-3948-4c8a-9a14-5207932296d4.gif)

</details>

### System configuration
**Avo version**: 2.11

**Rails version**:

**Ruby version**:

#### Impact

 - [ ] High impact
 - [ ] Medium impact
 - [x] Low impact

Adeynack:
It happens systematically when a file is attached. Not "sometimes". And this has quite a high impact in our case at [nerdgeschoss](https://github.com/nerdgeschoss) since the user expects to be reminded of the required fields instead of seeing the application crash. 😐 

We would appreciate if someone could look quickly into it. 🙏🏼

Paul-Bob:
@Adeynack thank you for the details, we are working on fixing this issue

PR #1173, opened 2022-08-25
fix: crash when attach files and validator render errors

Paul-Bob:
# Description

Fixes https://github.com/avo-hq/avo/issues/1119

# Checklist:

### DIFF
--- a/app/components/avo/fields/common/single_file_viewer_component.html.erb
+++ b/app/components/avo/fields/common/single_file_viewer_component.html.erb
@@ -33,7 +33,7 @@
         <% end %>
       </div>
       <div>
-        <% if @resource.authorization.authorize_action(:delete_attachments?, raise_exception: false) %>
+        <% if @file.id.present? && @resource.authorization.authorize_action(:delete_attachments?, raise_exception: false) %>
           <%= a_link destroy_path,
             icon: 'heroicons/outline/trash',
             method: :delete,


app/components/avo/fields/common/single_file_viewer_component.html.erb
```diff
@@ -33,7 +33,7 @@
         <% end %>
       </div>
       <div>
-        <% if @resource.authorization.authorize_action(:delete_attachments?, raise_exception: false) %>
+        <% if @file.id.present? && @resource.authorization.authorize_action(:delete_attachments?, raise_exception: false) %>
```
adrianthedev, line 36: Doesn't the Rails model have something like `persisted?`?

Paul-Bob: yes! thanks for remembering me, that was my first thought actually, don't remember how it ends like this ='))

Paul-Bob:
I just committed a new approach. In order to prevent inaccessible lost files on storage we delete them if validator render errors.

app/components/avo/fields/common/single_file_viewer_component.rb
```diff
@@ -40,4 +40,12 @@ def is_video?
   rescue
     false
   end
+
+  # If model is not persistent blob is automatically destroyed otherwise it can be "lost" on storage
+  def persistent_record?
```
adrianthedev, line 45:
```suggestion
  def record_persisted?
```

app/components/avo/fields/common/single_file_viewer_component.html.erb
```diff
@@ -1,3 +1,4 @@
+<% return unless persistent_record? %>
```
adrianthedev, line 1: We can use the built-in `render?` function on `single_file_viewer_component.rb`

https://viewcomponent.org/guide/conditional_rendering.html

app/components/avo/fields/common/single_file_viewer_component.rb
```diff
@@ -40,4 +40,12 @@ def is_video?
...
+    return true if @resource.model.persisted?
+
+    ActiveStorage::Blob.destroy(file.blob_id) if file.blob_id.present?
```
adrianthedev, line 48: Will this have no side-effects when updating an existing record?

app/controllers/avo/base_controller.rb
```diff
@@ -397,6 +397,7 @@ def create_success_action
 
     def create_fail_action
       respond_to do |format|
+        flash.now[:warning] = reupload_files_message if blobs_present?
```
adrianthedev, line 400: Let's not do this. They should figure it out from the form. It adds extra complexity.

adrianthedev (changes requested):

app/components/avo/fields/common/single_file_viewer_component.rb
Paul-Bob: No. On update `@resource.model.persisted?` returns `true`

app/components/avo/fields/common/single_file_viewer_component.html.erb
Paul-Bob: Didn't know that exists, just applied!

app/controllers/avo/base_controller.rb
Paul-Bob: Agree

app/components/avo/fields/common/single_file_viewer_component.rb
adrianthedev: Now it's `file.blob_id.present?` not `@resource.model.persisted?`, so I'm a bit confused. Is it ok like this?

CleanShot 2022-08-29 at 17 13 17

PR #1200, opened 2022-09-01
feature: custom policies

### LINKED ISSUE #1085: Custom Pundit policy per resource
### Feature

Add the ability to select which policy a resource should use. 

From [here](https://discord.com/channels/740892036978442260/740893011994738751/999699188004692009). 

### Approach
_These thoughts come to my mind but feel free to use your own approach._

Add a `authorization_policy` `class_attribute` to `BaseResource`.

Right now, each resource has an `authorization` method applied to it. This will create a new `AuthorizationService` which, by default, tries to figure out the policy by itself from the model you're trying to authorize.
I think that's where we need to hook in and pass it a policy class (around here).
Then, add a `policy` method to the `AuthorizationService` which will default to figuring it by itself from the `@model` or use the one you give it.

bear-in-mind:
# Description
Adding the possibility to specify a custom policy class for a given resource.

It's now easy as:

```ruby
class PhotoCommentResource < Avo::BaseResource
  self.model_class = ::Comment
  self.authorization_policy = PhotoCommentPolicy
  # ...
end
```

Fixes #1085

### About `policy_class` option

From pundit's documentation:
```ruby
# You can pass an argument to override the policy class if necessary. For example:
def create
  @publication = find_publication # assume this method returns any model that behaves like a publication
  # @publication.class => Post
  authorize @publication, policy_class: PublicationPolicy
  @publication.publish!
  redirect_to @publication
end
```
**Note**

I wish I could just pass the wanted policy once to Pundit via our service, instead of calling and passing `policy_class` in so many locations. Unfortunately, due to the way both `pundit` and our own `AuthorizationService` behave (by infering the policy class from the model of the `record` being passed), this is hardly possible. Thus, the fallback solution of forcing the `policy_class` at the last step of the process, which is the call to the `authorize` method.

** Another note**

While trying out the feature in the wild, I stumbled upon a very annoying bug, fixed in #1202

# Checklist:

## Manual review steps
To manually test that this is working, using the dummy app:

1. I added `self.authorization_policy = PhotoCommentPolicy` to the `PhotoCommentResource` just like above
2. I created a `PhotoCommentPolicy` and made `index?` return `false`
3. Ensured visiting `admin/resources/photo_comments/` would redirect me and prevent access
4. Turn `index?` back to `true`, refresh, and check that we can now access the index


adrianthedev:
The failing test is flaky and does that from time to time.

bear-in-mind:
Just realized I forgot to deal with scopes. Turning this back to draft while I fix this :)

app/controllers/avo/application_controller.rb
```diff
@@ -257,7 +258,13 @@ def render_unauthorized(exception)
     end
 
     def set_authorization
-      @authorization = Services::AuthorizationService.new _current_user
+      # We need to set @resource_name for the #resource method to work properly
+      @resource_name = resource_name
```
Paul-Bob, line 262: this exact line is done on `before_action :set_resource_name`

lib/avo/base_resource.rb
```diff
@@ -92,6 +93,7 @@ def find_scope
       end
 
       def authorization
+        # Is there a way to pass the authorization_policy here as well ?
```
Paul-Bob, line 96: you have acces to `authorization_policy`, `model_class` and `Avo::App.current_user` here

app/controllers/avo/application_controller.rb
bear-in-mind: ... but it doesn't work :( When I tried debugging it at this step, `@resource_name` would always be nil, despite the `set_resource_name` being called before `set_authorization`. Hence the duplication

Paul-Bob: image

Paul-Bob: how? :O

Paul-Bob: yes, i'll investigate that later, for now can we use the `set_resource_name` instead of `@resource_name = resource_name` ? It's easier to maintain this way

bear-in-mind:
Scopes working ! Was not exactly straightforward to implement, so I'm open to suggestions @Paul-Bob

lib/avo/services/authorization_service.rb
```diff
@@ -87,13 +95,14 @@ def defined_methods(user, record, **args)
         end
       end
 
-      def initialize(user = nil, record = nil)
+      def initialize(user = nil, record = nil, policy_class: nil)
         @user = user
         @record = record
+        @policy_class = policy_class
```
Paul-Bob, line 101: what do you say if we  initialize this this way, `@policy_class = policy_class || Pundit.policy(user, record).class`. I think this can make all other logic simpler

lib/avo/services/authorization_service.rb
```diff
@@ -68,12 +70,18 @@ def authorized_methods(user, record)
           end.to_h
         end
 
-        def get_policy(user, record)
-          Pundit.policy user, record
+        def get_policy(user, record, policy_class: nil)
```
Paul-Bob, line 73: if we initialize the way i suggested this method can be deleted and wher it's called we can do something like `policy_class&.new(user, record)`

lib/avo/services/authorization_service.rb
Paul-Bob: this way we can use`policy_class.new`everywhere instead of `if policy_class this else that end`

Paul-Bob:
> Scopes working ! Was not exactly straightforward to implement, so I'm open to suggestions @Paul-Bob

Great work, i tested some basic use cases and it seems to work. Maybe we can make the authorization service cleaner with the suggestions i let above

bear-in-mind:
These are great suggestions, thanks @Paul-Bob ! I'll try these today when I get a moment :)

bear-in-mind:
@Paul-Bob I want to bang my head on the wall so hard right now

Screenshot 2022-09-12 at 16 20 23

adrianthedev:
Could it be because it's `nil` @bear-in-mind? Try and `puts` the `policy_class` var to see what it is.

bear-in-mind:
Yes, I added this silly statement to see why the safe navigation operator was being ignored. I'm suspecting some kind of shadow variable. Anyway my head is kind of fuzzy ATM, I'll give it another try later

adrianthedev:
The error trace `authorization_service:13` is in a different place than the displayed code.

Paul-Bob:
> @Paul-Bob I want to bang my head on the wall so hard right now (trying to debug WHY THE HELL would `policy_class&.new(...)` return an error)
> 
> Screenshot 2022-09-12 at 16 20 23

That is weird but let me know, are you using the `policy_class` always as a method or you use the exact same name for a local variable?

bear-in-mind:
@adrianthedev I'm not sure what you're saying ; the error line is exactly the one mentionned by the error trace. But this code is not yet pushed, I'm trying to refactor as per @Paul-Bob's suggestions.

@Paul-Bob I'm not using `policy_class` as a method (I haven't set any `attr_accessor` on it), and the error happens in the singleton class anyway. Here's the complete code for the method :
```ruby
  def authorize(user, record, action, policy_class: nil, **args)
    return true if skip_authorization
    return true if user.nil?

    begin
      if policy_class.present? && policy_class&.new(user, record)
        Pundit.authorize user, record, action, policy_class: policy_class
      end

      true
    rescue Pundit::NotDefinedError => e
      return false unless Avo.configuration.raise_error_on_missing_policy

      raise e
    rescue => error
      if args[:raise_exception] == false
        false
      else
        raise error
      end
    end
  end
```

bear-in-mind:
@Paul-Bob @adrianthedev 
AAAAAAAAH I get it ! It's not returning an instance of `NilClass`, it's returning `NilClass` itself, thus a class, and not nil !!!

adrianthedev:
GG! It's a tough one to spot.

adrianthedev:
I can't figure out why `set_resource_name` isn't running before the `set_authorization` method. let's leave it like so for now.

Paul-Bob (approved):

lib/avo/services/authorization_service.rb
```diff
@@ -5,13 +5,13 @@ class AuthorizationService
       attr_accessor :record
 
       class << self
-        def authorize(user, record, action, **args)
+        def authorize(user, record, action, policy_class: nil, **args)
           return true if skip_authorization
           return true if user.nil?
```
adrianthedev, line 11: Maybe we should add this here too as a safeguard?

```suggestion
          policy_class ||= Pundit.policy(user, record)&.class

```

adrianthedev (commented):
I tried pushing changes to this branch too but was unable to do so.

I left a suggestion on the static method. In the scenario where someone uses the static method and doesn't pass in a `policy_class` that will fail. Maybe we should safeguard it by falling back to Pundits search.

Also, I think you merged another branch into this PR that shouldn't be there. Could you revert that, please? We have a different PR with conversations on it.

CleanShot 2022-09-28 at 11 20 37

bear-in-mind:
Yes I'm so sorry, didn't see my `git checkout` failed before I commited changes  🤦🏻‍♂️

adrianthedev:
Thanks for the contribution @bear-in-mind! Awesome PR!

Issue #1189, opened 2022-08-30
Sidebar for show & edit views


adrianthedev:
### Context

Traditionally, Avo is displaying fields stacked one on top of another. Now that's ok for most fields, but some have less information to display and don't need to take up all that space.

We want to introduce a sidebar on the `Show` and `Edit` views. The sidebar would show a compressed field version better suited for a narrower container.

#### V1

![V1](https://user-images.githubusercontent.com/1334409/187537221-1ce66b1e-d3cc-4bab-ad4f-a2a33a35e85b.jpg)

#### V2

![V2](https://user-images.githubusercontent.com/1334409/187537356-e7f2d1f3-d3b6-40be-960f-7b9721fec18d.jpg)

PR #1205, opened 2022-09-02
feature: resource_sidebar

Paul-Bob:
# Description

Fixes https://github.com/avo-hq/avo/issues/1189

# Checklist:

## Screenshots

CleanShot 2022-10-10 at 17 18 25
CleanShot 2022-10-10 at 17 18 51
CleanShot 2022-10-10 at 17 18 59


Paul-Bob:
@adrianthedev classes `class Avo::SidebarBuilder` and `class Avo::Sidebar` was inspired from tab classes, maybe there are some things we don't need for sidebar and others that we need and are not there

app/components/avo/panel_component.html.erb
```diff
@@ -23,7 +23,23 @@
...
+          <div class="text-2xl tracking-normal font-semibold text-gray-800 pb-2 mb-6 md:pb-0">
+            <span><%= @sidebar_name %></span>
+          </div>
+        <% end %>
+        <div class="h-full <%= white_panel_classes %>">
+          <%= sidebar %>
+        </div>
+      </div>
+    </div>
+  <% elsif body? %>
```
Paul-Bob, line 42: i don't like this complexity here, this should be refactored

app/components/avo/panel_component.rb
```diff
@@ -7,14 +7,16 @@ class Avo::PanelComponent < ViewComponent::Base
 
   renders_one :tools
   renders_one :body
+  renders_one :sidebar
```
Paul-Bob, line 10: Not sure if we should have a slot for sidebar here on the panel component or render it some other way

app/components/avo/panel_component.rb
```diff
@@ -7,14 +7,16 @@ class Avo::PanelComponent < ViewComponent::Base
...
   renders_one :bare_content
   renders_one :footer_tools
   renders_one :footer
 
-  def initialize(name: nil, description: nil, body_classes: nil, data: {}, display_breadcrumbs: false, index: nil, classes: nil, **args)
+  def initialize(name: nil, description: nil, body_classes: nil, data: {}, display_breadcrumbs: false, index: nil, classes: nil, sidebar_name: nil, **args)
     # deprecating title in favor of name
     @title = args[:title]
     @name = name || title
+    @sidebar_name = sidebar_name
```
Paul-Bob, line 19: same for sidebar_name

Paul-Bob: name was removed

Issue #968, opened 2022-06-16
Add the ability to skip the Show view for a resource


adrianthedev:
### Feature

Some users don't need the `Show` screen for resources and only need the `Edit` one. It would be nice to be able to completely remove it.

#### Some things that came out from our brainstorming:

 - [ ] ensure all current controller paths (after create/update)
 - [ ] ensure all current resource paths lead to resource edit path
 - [ ] ensure all fields that support "link_to_resource" do the same
 - [ ] ensure actions redirects to edit
 - [ ] remove show link from resource controls
 - [ ] make it testable
 - [ ] add the redirect from routes

hassyyy:
Would be great if we can have the ability to hide the show button in list view. 
Justification: It is not needed when `link_to_resource` is set

adrianthedev:
Some things that came out from our brainstorming:

# Skip show view

 - [ ] ensure all current controller paths (after create/update)
 - [ ] ensure all current resource paths lead to resource edit path
 - [ ] ensure all fields that support "link_to_resource" do the same
 - [ ] ensure actions redirects to edit
 - [ ] remove show link from resource controls
 - [ ] make it testable
 - [ ] add the redirect from routes

Paul-Bob:
@adrianthedev i suggest to no do the last step (add redirect from routes) yet. We expect all system to work fine without the forced redirect and if we add that redirect we kinda lost track if something is not working as expected

PR #1209, opened 2022-09-03
feature: skip_show_view

Paul-Bob:
# Description

Fixes https://github.com/avo-hq/avo/issues/968#event-7290444871

Avo now can be configured to ignore the resources show page and to use only edit views instead!

For example, lets say, you just created some random resource, Avo's default behavior would redirect you to that resource show right? Now you can change that behavior and be redirected directly to the edit view, and there you'll find all the associations tables, where you can attach, detach associations and you also can edit or delete that parent record.

To change that behavior add this simple line of code `config.skip_show_view = true` in `avo.rb` file 
# Checklist:


app/components/avo/views/resource_edit_component.html.erb
```diff
@@ -22,6 +22,24 @@
         icon: 'arrow-left' do %>
           <%= t('avo.cancel').capitalize %>
         <% end %>
+        <% #TODO: for some reason new button act as delete button when delete button is present %>
```
Paul-Bob, line 25: @adrianthedev I'm having some troubles here, when i put `method: :delete` on destroy button, both, save and destroy buttons acts as destroy button, if I remove `method: :delete`, both buttons acts like edit. I don't know how to make 2 separate actions for the same form without some kind of param and in the routes check if param is present to decide if we go to update or to destroy route.

Paul-Bob: Maybe a data action and an AJAX request solve this, i don't know what is the best approach

Paul-Bob: actually changing it from button to link solved this issue

adrianthedev (commented):
The PR works as expected in manual tests. Great job!

I'd like to make sure we extract & group the show/edit functionality properly and with the least duplication possible.

I'll have another look this evening and maybe come up with some solutions for that.

app/components/avo/fields/belongs_to_field/index_component.rb
```diff
@@ -1,4 +1,13 @@
...
 class Avo::Fields::BelongsToField::IndexComponent < Avo::Fields::IndexComponent
+  def resource_path
+    return edit_path if Avo.configuration.skip_show_view
+
+    helpers.resource_path(model: @field.value, resource: @field.target_resource)
+  end
+
+  def edit_path
+    helpers.edit_resource_path(model: @field.value, resource: @field.target_resource)
+  end
```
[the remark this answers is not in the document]
adrianthedev: How about we add this method to `url_helpers.rb` and use it everywhere? Would that be pheasable?

```ruby
    def show_or_edit_path(**args)
      if Avo.configuration.skip_show_view
        edit_resource_path(**args)
      else
        resource_path(**args)
      end
    end
```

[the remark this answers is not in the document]
Paul-Bob: I agree that we need this, we have some edge cases where the model or some args are different from `edit_resource_path` and for `resource_path`. I'll refactor it and implement this method, then we can discuss later about that particular cases. Btw, i don't think `show_or_edit_path` is the most appropriated name, can we use something like `default_resource_path` or `resource_default_view_path`?

app/controllers/avo/associations_controller.rb
```diff
@@ -65,13 +65,18 @@ def create
...
         if @model.save
-          format.html { redirect_back fallback_location: resource_path(model: @model, resource: @resource), notice: t("avo.attachment_class_attached", attachment_class: @related_resource.name) }
+          format.html { redirect_back fallback_location: resource_default_view_after_path, notice: t("avo.attachment_class_attached", attachment_class: @related_resource.name) }
         else
           format.html { render :new }
         end
       end
     end
 
+    def resource_default_view_after_path
```
Paul-Bob, line 75: @adrianthedev here and in the `base_controller.rb`i had to change this method name to prevent helper'se method overriding, i'm open to name suggestions, maybe `resource_default_view_response_path` ??

adrianthedev:
I've been looking at this for a while now and can't seem to find a way of wrapping this up in a nice way.

1. I'm afraid that at one point someone will forget to call the `resource_default_view_path` and will call the `resource_path` or something and that will break it for some people.
2. I don't like the `ResourceEditPath` Concern. It does not feel like the right abstraction.
3. There are a lot of `if skip_show_view` statements

I'm still thinking about it.

Paul-Bob:
1 - I don't know how to prevent that, we have the spec that will test some cases but in future features maybe this spec will not cover them... We can kinda block that resource_path for using or protect the resource_path to call the edit_path if it's called with skip_show_view on

2 - Actually i don't know what is the best approach to abstract those 2 methods, they are used for 2 different view components without connection between them.

3 - I'm not sure if we can take them out, we need them on initializers, in `can_view?`, on after CRUD paths, on after Association action (attach / detach) and to decide where to go on `resource_default_view_path`. I'm trying to find out some unnecessary if statement

Paul-Bob:
Hei @adrianthedev, after a major refactor I can say it looks better now. The concern no longer exists because it's not needed and I think we reduced the if statements. About the `I'm afraid that at one point someone will forget to call the resource_default_view_path and will call the resource_path or something and that will break it for some people.` I don't know what is the best approach

adrianthedev (approved):
The refactor definitely looks better!

There's one thing that I'm not sure about. The name of the method `resource_default_view_path`. 
First of all, it was a mistake to call it `resource`, when it should be `record`. It's like that because the name of the "model" in routes is `resources`.
Second, it's a bit verbose `resource_default_view_path`.

How about calling it `resource_view_path`?

Paul-Bob:
> The refactor definitely looks better!
> 
> There's one thing that I'm not sure about. The name of the method `resource_default_view_path`. First of all, it was a mistake to call it `resource`, when it should be `record`. It's like that because the name of the "model" in routes is `resources`. Second, it's a bit verbose `resource_default_view_path`.
> 
> How about calling it `resource_view_path`?

I totally agree with `resource_view_path` instead `resource_default_view_path`. About the `record...path` instead `resource...path` maybe we should do a separate PR with that refactor

app/components/avo/views/resource_edit_component.rb
```diff
@@ -16,16 +16,28 @@ def title
...
-      helpers.resource_path(model: model.safe_constantize, resource: relation_resource, resource_id: params[:via_resource_id])
-    elsif via_index?
-      helpers.resources_path(resource: @resource)
-    elsif is_edit? # via resource show page
-      helpers.resource_path(model: @resource.model, resource: @resource)
-    else
-      helpers.resources_path(resource: @resource)
+    return resource_default_view_path if via_resource?
+    return resources_path if via_index?
+
```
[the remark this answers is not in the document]
Paul-Bob: I think it's more readable that way

PR #1206, opened 2022-09-03
feature: add ability to make use of belongs_to associations in actions

ckmorris:
# Description

Feature: Add ability to utilize `belongs_to` associations in Avo Actions

# Checklist:
- [X] I have performed a self-review of my own code
- [-] I have commented my code, particularly in hard-to-understand areas
- [-] I have made corresponding changes to the [documentation](https://github.com/avo-hq/docs)
- [-] I have added tests that prove my fix is effective or that my feature works

## Manual review steps
1. Add an action to an Avo resource
2. Add a belongs_to field to the Avo resource
3. Set the belongs_to field as `actionable: true` and `actionable_class: ParentClass` 
4. `handle` the `fields['reflection_name']` that occurs as a result
5. ???
6. Profit

If you need me to write the specs and documentation, I will, just let me know. I wanted to get this out to you sooner rather than later, though, so that you can review my approach.

Manual reviewer: please leave a comment with output from the test if that's the case.

PR #1211, opened 2022-09-05
Made sure model is set on the resource for actions

iainbeeston:
# Description
I'd like to use the trix field in an action, but it does not work right now because the resource does not have a model set by the actions controller. I've changed it to explicitly hydrated the resource like the resources controller does.

Fixes # (issue)

# Checklist:

## Manual review steps

1. Check that a pre-existing action still works correctly
2. Create a new action that has a trix field and confirm that this works without errors.

### DIFF
--- a/app/controllers/avo/actions_controller.rb
+++ b/app/controllers/avo/actions_controller.rb
@@ -4,11 +4,15 @@ module Avo
   class ActionsController < ApplicationController
     before_action :set_resource_name
     before_action :set_resource
+    before_action :set_model, only: :show, if: ->(request) do
+      # Try to se the model only if the user is on the record page.
+      # set_model will fail if it's tried to be used from the Index page.
+      request.params[:id].present?
+    end
     before_action :set_action, only: [:show, :handle]
 
     def show
-      model = @resource.class.find_scope.find params[:id]
-      @resource.hydrate(model: model, view: :show, user: _current_user, params: params)
+      @resource.hydrate(model: @model, view: :show, user: _current_user, params: params)
       @model = ActionModel.new @action.get_attributes_for_action
     end
 
@@ -46,11 +50,7 @@ module Avo
     def set_action
       action_class = params[:action_id].gsub("avo_actions_", "").camelize.safe_constantize
 
-      if params[:id].present?
-        model = @resource.class.find_scope.find params[:id]
-      end
-
-      @action = action_class.new(model: model, resource: resource, user: _current_user, view: :edit)
+      @action = action_class.new(model: @model, resource: @resource, user: _current_user, view: :edit)
     end
 
     def respond(response)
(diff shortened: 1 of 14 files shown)


iainbeeston:
There might be a better way to do this, I'm guessing based on what other controllers do...

app/controllers/avo/actions_controller.rb
```diff
@@ -7,6 +7,8 @@ class ActionsController < ApplicationController
     before_action :set_action, only: [:show, :handle]
 
     def show
+      model = @resource.class.find_scope.find params[:id]
```
Paul-Bob, line 10: I think `params[:id]` here is always nil, am I wrong?

Paul-Bob: I mean, in show view it's ok but in other use cases this will break, maybe we should protect it

iainbeeston: Yes quite possibly! This is just some code i threw together quickly and I don’t have many examples of actions to test it on, it may well have bugs

adrianthedev: I think there's value in this approach. There's [this](https://github.com/avo-hq/avo/pull/1206) PR that tackles the same issue in a different way. 
Yes, it will work only for one record (from show view), but that's ok. We'll soon introduce row-only actions so this will come in handy. 

I'd like to have a test here to protect this behavior in the future.

adrianthedev:
I had a few free minutes last night and pushed a few more changes to correctly allow the model to be attached to the resource on the record pages.

Fixes https://github.com/avo-hq/avo/pull/1206

app/controllers/avo/actions_controller.rb
```diff
@@ -4,9 +4,15 @@ module Avo
   class ActionsController < ApplicationController
     before_action :set_resource_name
     before_action :set_resource
+    before_action :set_model, only: :show, if: ->(request) do
```
Paul-Bob, line 7: [Link to commit discussion](https://github.com/avo-hq/avo/commit/eccb22901a7e3cd369cb46b4b25c6d5793b447d5#r83390129)

iainbeeston:
This PR works for my use case 👍

adrianthedev:
Got it. I'll do some digging for file fields on actions.

adrianthedev:
@iainbeeston I tried your config on an action and I don't get the same results. I see the field  visible.

Can you try this config on a different action?

But before of that check if you have a policy for that resource and see what the `upload_attachments?` method returns. It might be that the authorization is preventing that part of the field, because we can see the label.

```ruby
  field :message, as: :textarea, help: "Tell the fish something before releasing."
  field :user, as: :belongs_to, searchable: true
  field :image, as: :file, is_image: true, direct_upload: true, accept: "image/*"
```

CleanShot 2022-09-08 at 19 00 38@2x

adrianthedev:
Got it. I'll definitely release this. It's got a lot of value.

PR #1215, opened 2022-09-06
fix: Ensured @parent_resource was being passed down to grid items controls

bear-in-mind:
# Description
Here is a video description of the bug at hand : https://www.loom.com/share/2eb63bcb559347949d61e3f7f868cebf
Basically, when displaying a `has_many` association as a grid, the original `parent_resource` would be lost further down the component tree. This does not occur in list mode.

The problem is that the behaviour of the `get_resource_by_model_name` method is unreliable as it currently is: if several resources share the same `model_class`, the method will return the first of these resources in alphabetical order.

The fix consists in passing the `@parent_resource` from the upper component to the lower level ones.

# Checklist:

## Manual review steps
Trying out the #1200 feature, I was able to restrict association actions through a custom policy after the fix was done. Before, it would fetch the default policy of the wrong resource, since the `get_resource_by_model_name` would not return the actual parent resource.


app/components/avo/index/resource_controls_component.rb
```diff
@@ -65,8 +65,8 @@ def singular_resource_name
   end
 
   def parent_resource
+    return @parent_resource unless @parent_resource.blank?
```
Paul-Bob, line 68: Can we use `if @parent_resource.present?` ? I think it would make it more readable this way

app/components/avo/index/resource_controls_component.rb
```diff
@@ -65,8 +65,8 @@ def singular_resource_name
...
     return nil if @parent_model.blank?
-
```
Paul-Bob, line 69: It's a common and good practice to insert a newline after a guard clause to increase readability

app/controllers/avo/associations_controller.rb
```diff
@@ -23,7 +23,6 @@ def index
       @parent_resource.hydrate(model: @parent_model)
       @query = @authorization.apply_policy @parent_model.public_send(params[:related_name])
       @association_field = @parent_resource.get_field params[:related_name]
-
```
Paul-Bob, line 26: newline here increase readability too

lib/avo/concerns/fetches_things.rb
```diff
@@ -51,6 +51,9 @@ def get_resource_by_name(name)
         #
         # get_resource_by_name('User') => UserResource
         # get_resource_by_name(User) => UserResource
+
+        # FIXME: This query doesn't necessarily get the right resource.
+        # From all the resources that use the same model_class, it will fetch the first one in alphabetical order
```
Paul-Bob, line 56: `"it will fetch the first one in alphabetical order"`

It happens when `model_resource_mapping` is missing.  https://docs.avohq.io/2.0/resources.html#use-multiple-resources-for-the-same-model 

Anyway, i think `model_resource_mapping` don't solve the problem addressed in this PR because it only allows you to choose 1 default resource per model.

Paul-Bob:
Hi @bear-in-mind , thanks for the effort putted into this PR. I think you figured it out, we basically do the same on the table components! @adrianthedev I think this can be a valid solution for now. Maybe in the future we can try to figure it out how to map the correct resource to the used model according to each situation without having to pass the `@parent_resource` to everywhere

Paul-Bob:
@bear-in-mind i think this configuration should fix the problem exposed on your video

```ruby
# config/initializers/avo.rb
Avo.configure do |config|
  config.model_resource_mapping = {
    'Ticket': 'DocumentAnalysisResource'
  }
end
```

lib/avo/concerns/fetches_things.rb
bear-in-mind: Exactly ; I saw this configuration, unfortunately it won't be of help in our case. Our use case of Avo is building multiple and specific resources around one main model, which is tied to probably a dozen resources...

bear-in-mind:
> @bear-in-mind i think this configuration should fix the problem exposed on your video
> 
> ```ruby
> # config/initializers/avo.rb
> Avo.configure do |config|
>   config.model_resource_mapping = {
>     'Ticket': 'DocumentAnalysisResource'
>   }
> end
> ```

Thanks @Paul-Bob for the thorough review, sorry for messing with newlines. Your suggestion would unfortunately not solve the issue for us, as the main resource for our `Ticket` model is in fact the `TicketResource` ; it's just that lots of other resources are baed on the same model, which is probably not a scenario that you guys had in mind originally.

Paul-Bob:
> > @bear-in-mind i think this configuration should fix the problem exposed on your video
> > ```ruby
> > # config/initializers/avo.rb
> > Avo.configure do |config|
> >   config.model_resource_mapping = {
> >     'Ticket': 'DocumentAnalysisResource'
> >   }
> > end
> > ```
> 
> Thanks @Paul-Bob for the thorough review, sorry for messing with newlines. Your suggestion would unfortunately not solve the issue for us, as the main resource for our `Ticket` model is in fact the `TicketResource` ; it's just that lots of other resources are baed on the same model, which is probably not a scenario that you guys had in mind originally.

Understood. Do you have any clue or suggestion on how to make Avo choose the appropriate resource for each situation?

adrianthedev:
I think the way I always looked at it was like this

1. you start with one resource per model. And it's the first one in line in `get_resource_by_model_name`
2. you add a resource for an association `belongs_to`, `has_many` with `use_resource`
3. now that resource might not be the first one in `get_resource_by_model_name` and you use `model_resource_mapping` for setting the default one.

I think the best way of looking at things is "you have a default resource per model, and if you want to use a different one in a field, you just tell the field". We don't want to make the `model_resource_mapping` very complex.
Now, we just have to make sure we have all the outlets necessary to specify the resource to use, like `use_resource` for association fields.

Does that sound right to you guys?

bear-in-mind:
I would tend to agree, unfortunately I experienced several times the fact that at some point in the code, the `field` loses the reference to its original resource, thus resorting to calling this getter method, which would always return some random resource. Even if a _default_ resource is specified in the `model_resource_mapping` (which by the way seems to not be documented anymore), it won't help fetching the _right_ one at a given time.

In our scenario, our `Ticket` model is bound to at least 10 different resources, each with a specific configuration and authorizations. And quite often, pages would blow up expecting some fields or attributes that don't exist, wrong policies are being looked up, links would break or take us to a different resource. I still think the need for a solid lookup method is missing. Maybe a `use_resource` param ? Or maybe some key-value in the `session` object (although that would prove more complicated across navigation) ?

app/components/avo/index/grid_item_component.rb
```diff
@@ -2,7 +2,6 @@
 
 class Avo::Index::GridItemComponent < ViewComponent::Base
   include Avo::ResourcesHelper
-
```
adrianthedev, line 5: Can we leave this empty space please? It sets a boundary between the different types of declarations.

```suggestion

```

lib/avo/concerns/fetches_things.rb
adrianthedev, line 56: Could we make the whole comment block like so?

```ruby
        # Returns the Avo resource by singular snake_cased name
        # From all the resources that use the same model_class, it will fetch the first one in alphabetical order
        #
        # get_resource_by_name('User') => UserResource
        # get_resource_by_name(User) => UserResource
```

The method isn't really broken. It will still fetch the resource as requested. If the resource is the wrong one, that needs to be fixed in its proper place.

adrianthedev (commented):
Hey @bear-in-mind. Thanks for the PR! I wanted to push some changes but failed to do so. Not sure why.

Could you apply the feedback from the comments? It's ready to merge after that.

Thanks ✌️

adrianthedev:
> I would tend to agree, unfortunately I experienced several times the fact that at some point in the code, the field loses the reference to its original resource, thus resorting to calling this getter method, which would always return some random resource. Even if a default resource is specified in the model_resource_mapping (which by the way seems to not be documented anymore), it won't help fetching the right one at a given time.
> 
> In our scenario, our Ticket model is bound to at least 10 different resources, each with a specific configuration and authorizations. And quite often, pages would blow up expecting some fields or attributes that don't exist, wrong policies are being looked up, links would break or take us to a different resource. I still think the need for a solid lookup method is missing. Maybe a use_resource param ? Or maybe some key-value in the session object (although that would prove more complicated across navigation) ?

Regarding this scenario @bear-in-mind, I wouldn't want to have to **map everything in one place**. The mapping config would become gnarly really fast. Imagine having to map a `has_many-Ticket-for-ClientResource`. The "DSL" for that will become very difficult to learn and scan at a glance.

Having query params fixes the issue but from my experience, relying on them is quite fragile.

I strongly believe that if we take it one by one, like in the case you pointed out here, we'll be able to send the parent downstream through the components. It's just a matter of finding the edge-cases.
So, please send reports on where Avo picks up the wrong resource, and we'll fix them one by one.

[Docs on the mapping](https://docs.avohq.io/2.0/resources.html#model-resource-mapping).

bear-in-mind:
@adrianthedev done !

app/components/avo/index/grid_item_component.rb
```diff
@@ -4,6 +4,8 @@ class Avo::Index::GridItemComponent < ViewComponent::Base
   include Avo::ResourcesHelper
   attr_reader :parent_resource
 
+  attr_reader :parent_resource
+
```
adrianthedev, line 8:
```suggestion
```

adrianthedev: These lines are extra @bear-in-mind.

bear-in-mind: Gooood my editor's killing me

PR #1220, opened 2022-09-08
Make the trix toolbar responsive

iainbeeston:
# Description

By default trix toolbox does not wrap no matter what the size of the container is. That can be an issue when using a trix field in a smaller container like an action modal.

Here's trix in an action modal and on a resource form on 2.14.2 (see how the toolbar is wider than the container, which also hides the cancel and save buttons because they are aligned to the right):

Screenshot 2022-09-08 at 13 33 57

Screenshot 2022-09-08 at 13 35 13

Here is the same action modal and form on this branch:

Screenshot 2022-09-08 at 13 32 43

Screenshot 2022-09-08 at 13 34 15

This is a known [issue with trix](https://github.com/basecamp/trix/issues/524) and there's no plan to fix it.

Fixes # (issue)

# Checklist:

## Manual review steps

1. Add an action that has a trix field in it (`  field :content, as: :trix`)
2. Open the action in the browser and confirm that the trix toolbar and also the save cancel buttons are visible

### DIFF
--- a/app/assets/stylesheets/css/fields/trix.css
+++ b/app/assets/stylesheets/css/fields/trix.css
@@ -15,3 +15,13 @@
 .trix-content ol {
   @apply list-decimal;
 }
+
+trix-toolbar {
+  .trix-button-row {
+    @apply flex-wrap gap-x-5;
+  }
+
+  .trix-button-group:not(:first-child) {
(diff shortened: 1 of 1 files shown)


adrianthedev (approved):
Spot on! 👌

PR #1221, opened 2022-09-08
Removed unused parameter from `after_create_path`

iainbeeston:
# Description

AFAIK the `resource` parameter for `resource_path` is only used if the value is a model instance with an id attribute. Here it's a model class instead, so it'll be ignored and the `resource_id` parameter is used instead. I think it'd be simpler and easier to understand if this unnecessary parameter was removed?

Fixes # (issue)

This doesn't actually fix an issue, just cleans up the code a little

# Checklist:

## Manual review steps

1. Create a nested resource (eg. project comments in the demo app) and confirm that after creation you are redirected back to the show page for the parent resource (eg. project)


iainbeeston:
OK I just realised that the model param is required 🤦 It'd be better if the `resource_id` param was removed and `model` (for the parent model) was passed instead but it's not clear how to load the parent here

Issue #1172, opened 2022-08-25
add the ability to delete the selected resources from the index page


adrianthedev:

adrianthedev:
THis has been discussed here https://github.com/avo-hq/avo/pull/1224.
Closing for now.

PR #1224, opened 2022-09-10
feature: batch delete

Paul-Bob:
# Description

Fixes https://github.com/avo-hq/avo/issues/1172

batch delete

# Checklist:


Paul-Bob:
@adrianthedev i don't know if any documentation is needed for this one, it's intuitive and for now we don't have any configuration to enable/disable this feature

lib/generators/avo/templates/locales/avo.nb-NO.yml
```diff
@@ -16,6 +16,15 @@ nb-NO:
...
     resource_destroyed: 'Ressurs slettet'
+    record: # HELP NEEDED FOR TRANSLATIONS
+      zero: '%{count} records'
+      one: '%{count} record'
+      other: '%{count} records'
+    destroyed: # HELP NEEDED FOR TRANSLATIONS
+      zero: 'destroyed'
+      one: 'destroyed'
+      other: 'destroyed'
+    resource_batch_destroyed: '%{records} %{destroyed}'
```
Paul-Bob, line 27: Can't find nobody to help us on this, want to translate it online or let it like that and wait for someone to translate it?

app/avo/actions/pre_made_action_batch_delete.rb
```diff
@@ -0,0 +1,23 @@
...
+module Actions
+  class PreMadeActionBatchDelete < Avo::BaseAction
+    self.name = I18n.t('avo.delete').capitalize
+    self.message = I18n.t('avo.are_you_sure').capitalize
+    self.confirm_button_label = I18n.t('avo.delete').capitalize
+
+    def handle(**args)
+      models, = args.values_at(:models, :fields, :current_user, :resource)
+
+      models.each(&:destroy!)
```
veelenga, line 12: This should probably be wrapped into transaction

Paul-Bob: Actually i thanked about 3 approaches and i'm not sure which one is the better one.

1 - As you suggested wrap the delete into transaction and if one fails none is deleted
2 - Output the exception to the user for each model that fails 
3 - Delete all and stop when one of them fails

I think the better one is the one you suggested, @adrianthedev what do you think?

lib/generators/avo/templates/locales/avo.nb-NO.yml
adrianthedev: Try and find a similar project that has translations (think Wordpress, e-commerce websites, etc.) or use Google translate. We'll get suggestions from developers if they are wrong.

app/avo/actions/pre_made_action_batch_delete.rb
adrianthedev: I'm now thinking about exposing this action somehow to the public as it is and allowing the developer to choose the strategy.

I mean, when you run `avo:install` to generate this action in the Rails app and then enable the developer to add it to their resources and modify it as they wish.

The downside is that we should add the `index_controls` option too.

Paul-Bob:
The batch destroy logic approach should be decided by the developer for each project, closing this for now.

PR #1262, opened 2022-09-27
refactor: add reform

adrianthedev:
# Description

Exploring the option of using https://github.com/trailblazer/reform.

# Things to figure out

- [ ] How do we manually fill each form field? Avo requires that because some fields are filled in a different way. We'd like to support that. Ex: belongs_to fields are declared like `field :user` but the model key is different `user_id`
- [ ] How do we do and respond to validation?

# Checklist:

## Manual review steps

1. Step 1
1. Step 2


adrianthedev:
Closing for now

PR #1267, opened 2022-09-29
fix: norwegian translations

tersor:
# Description
Updated translations for norwegian bokmål and added translations for norwegian nynorsk.


lib/generators/avo/templates/locales/avo.nb.yml
```diff
@@ -1,4 +1,4 @@
-nb-NO:
```
adrianthedev, line 1: Is this the right way of having the translations in Norway `nb` vs. `nb-NO`? I'm not very familiar with localization.

adrianthedev (approved):

lib/generators/avo/templates/locales/avo.nb.yml
tersor: Yes it is. [Example reference](https://github.com/svenfuchs/rails-i18n/commit/196ea80372e08941f9e9e8899e3af7b0b86e703c).

adrianthedev: Thank you @tersor! I appreciate the help.

Could we, in the future when we add new strings, ping you to validate, and help with Norwegian translations?

tersor: Jep, sure!

Issue #1217, opened 2022-09-07
Filters button overlap


Paul-Bob:
### Describe the bug

Filters button overlap an already opened filters menu
![image](https://user-images.githubusercontent.com/69730720/188862787-06b4b47a-767b-415c-8319-0c190d452960.png)

### System configuration
**Avo version**:
2.14.2

#### Impact

 - [ ] High impact
 - [ ] Medium impact
 - [x] Low impact

#### Urgency

 - [ ] High urgency
 - [ ] Medium urgency
 - [x] low urgency

Paul-Bob:
Also, filters don't make sense on empty table

adrianthedev:
What if the table is empty from the filters selaection? How would one reset the filters?

PR #1290, opened 2022-10-06
Prevent filter buttons from overlapping filter popovers

koppen:
# Description

Buttons and related action components should inherit the z-index from their wrapping elements. If need be, the popups they open are the ones that should have a high z-index.

In the same vein, panels are intended to "hover" above other elements, so they should have a higher z-index than most other elements.

As part of debugging this issue I mapped the following z-index uses:

* `z-20`: Dividers, tabs, 
* `z-30`: selectAllOverlay
* `z-40`: Popovers, dropdowns
* `z-50`: Modals, dialogs, Navbar

Fixes https://github.com/avo-hq/avo/issues/1217

# Checklist:

## Manual review steps

1. Configure a resource to have multiple sub-resources that each have one or more filters.
2. Go to the resources show page.
3. Open the topmost filter and scroll down to the filter button below - it should not be visible through the open filter dropdown.

### DIFF
--- a/app/components/avo/actions_component.html.erb
+++ b/app/components/avo/actions_component.html.erb
@@ -1,4 +1,4 @@
-<div class="relative flex flex-col w-full sm:w-auto z-40 js-actions-dropdown"
+<div class="relative flex flex-col w-full sm:w-auto js-actions-dropdown"
   data-controller="toggle-panel actions-picker"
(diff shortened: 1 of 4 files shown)


adrianthedev:
Thank you for your contribution @koppen. Spot on! 👌

Issue #1253, opened 2022-09-23
Add a visibility option to filters


adrianthedev:
Add a `self.visible` option to filters similar to how we do it to actions.

Maybe even extract that piece of code somehow if that makes sense.

_Originally posted by @ckmorris in https://github.com/avo-hq/avo/discussions/838#discussioncomment-3675127_

PR #1298, opened 2022-10-09
feature: filter visibility

Paul-Bob:
# Description

Fixes https://github.com/avo-hq/avo/issues/1253

# To do:

 - [x] Update the docs on actions and filters visible block
 - [x] Update the docs in the upgrade guide warning about the breaking change
 - [x] Add the `visible` block to the filter templates (generators). Maybe like on `dashboard.tt`?

# Breaking change

Check the visible block args!

# Checklist:


Paul-Bob:
@adrianthedev i think we should use a dry initializer on this. I inspired myself on how action visibility is done but i think we should refactor that one too, with the dry initializer both, action and filter visibility is more scalable.

adrianthedev:
Ok @Paul-Bob. Let's do that then!

app/components/avo/actions_component.rb
```diff
@@ -17,9 +17,7 @@ def render?
   end
 
   def actions
-    @actions.select do |action|
-      !action.class.in?(@exclude)
-    end
+    @actions.reject { |action| action.class.in?(@exclude) }
```
Paul-Bob, line 20: unrelated

lib/avo/resources/controls/action.rb
```diff
@@ -14,9 +14,7 @@ def initialize(klass, model: nil, resource: nil, view: nil, **args)
         end
 
         def action
-          return @instance if @instance.present?
-
-          @instance = @klass.new(model: @model, resource: @resource, view: @view)
+          @instance ||= @klass.new(model: @model, resource: @resource, view: @view)
```
Paul-Bob, line 17: unrelated

spec/dummy/app/avo/actions/dummy_action.rb
```diff
@@ -1,7 +1,18 @@
 class DummyAction < Avo::BaseAction
   self.name = "Dummy action"
   self.standalone = true
-  self.visible = ->(resource:, view:) do
+  self.visible = -> do
```
Paul-Bob, line 4: warning users about this change

lib/avo/base_action.rb
```diff
@@ -56,8 +59,10 @@ def action_name
       self.class.to_s.demodulize.underscore.humanize(keep_id_suffix: true)
     end
 
-    def initialize(model: nil, resource: nil, user: nil, view: nil)
+    def initialize(model: nil, parent_model: nil, parent_resource: nil, resource: nil, user: nil, view: nil)
```
adrianthedev, line 62: I'm a bit weary of adding that many things (`parent_model` and `parent_resource`) into an action. I feel they are a bit too extra.

Can we try not initializing with these two and (optionally) pass them in the `visible_in_view` method, so we have them in the Host?

lib/avo/resources/controls/action.rb
adrianthedev: 💪

spec/dummy/app/avo/actions/dummy_action.rb
adrianthedev: Could you also update the docs and the upgrade guide?

adrianthedev (commented):
Amazing job with the refactor!

I think we can lighten the load on `BaseAction` and `BaseFilter` and pass the parent stuff in the `visible_in_view` method.

I'll update the PR description with things still needed to be done.

spec/dummy/app/avo/actions/dummy_action.rb
Paul-Bob: sure

lib/avo/base_action.rb
Paul-Bob: Great observation, do you think this apply to the filters too?

adrianthedev: Yes, definitely!

Paul-Bob: @adrianthedev refactor done and ready for last review and merge

adrianthedev (approved):
I left a comment about initializint the `BaseFilter` and these suggestions. Otherwise, the PR is good to go.

lib/avo/base_action.rb
Paul-Bob: sure, no problem

PR #1300, opened 2022-10-10
Extract pundit to client

dhnaranjo:
# Description
Auth changes are stressful so let's do it in pieces. The goal of this PR is to pull all direct interaction with Pundit into a single client class AND simplify the AuthorizationService.

I believe that we should only need two public methods on the auth service and clients, `authorize` and `apply_policy`, with errything else being kind of implementation details that are obscuring what you're really trying to get done.

# Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the [documentation](https://github.com/avo-hq/docs)
- [ ] I have added tests that prove my fix is effective or that my feature works

## Manual review steps

This is a WIP, I'm leaving comments about other things I'm thinking on to discuss. I'll squash the lil commits when it's time but for now it helps with following what I'm up to.

PR #1303, opened 2022-10-10
Fix/association policy

### LINKED ISSUE #1283: Wrong policy is applied in AssociationsController
FYI I just noticed that in 'has_many' frames, some resources wouldn't show up.

After digging a little, it turns out that Avo would keep using the `parent_resource` policy scope instead of the policy of the actual resource, and thus disregard any custom policy present in the current resource.

I already solved the bug, but am working on a batch of fixes related to `use_resource` options. Will push a PR soon :)

bear-in-mind:
# Description
As it stands, the associations controller fetches the parent policy instead of the actual associated resource's policy in association frames. 

This PR makes sure we use the actual resource's policy and its scopes instead of the parent's.
If a custom policy is specified in the resource as an `authorization_policy`, we'll use it ; otherwise we'll default on the associated resource's default policy.

Fixes #1283 

# Checklist:


adrianthedev:
Can you share an example in the dummy app where this happens and what should really happen?

I'm working on a parallel branch to fix this.

*BTW, I know you're on vacation, so don't worry about it now. We'll fix it when you come back.

adrianthedev:
This might be fixed by https://github.com/avo-hq/avo/pull/1300. Can you give it a try @bear-in-mind?

adrianthedev:
Closing for now.

PR #1305, opened 2022-10-10
feature: fail action and keep modal open

### LINKED ISSUE #1271: Prevent closing a standalone action if validation fails
### Feature
Instead of closing the action modal and then rendering an error, leave the modal open with the user's inputs and render the error toast notification.

### Current workarounds
Just warning the user in the `message` to fill out all required fields

### Screenshots

avo-action

Paul-Bob:
# Description

Fixes #1271

# Checklist:

## Manual review steps

1. Step 1
1. Step 2


lib/avo/base_action.rb
```diff
@@ -145,8 +145,8 @@ def succeed(text)
       self
     end
 
-    def fail(text)
-      add_message text, :error
+    def fail(text, keep_modal_open: false)
```
Paul-Bob, line 148: @adrianthedev I don't like this name, what do you think about `persistent_error`?

Paul-Bob: i mean, a different method called `def persistent_error`

Paul-Bob: and maybe we should duplicate that fail method to `def error` and start printing deprecation messages inside `def fail` method because it's a ruby reserved method.

lib/avo/base_action.rb
```diff
@@ -145,8 +145,21 @@ def succeed(text)
       self
     end
 
-    def fail(text, keep_modal_open: false)
-      add_message text, keep_modal_open ? :keep_modal_open : :error
+    def fail(text)
+      Rails.logger.warn "DEPRECATION WARNING: Action fail method is deprecated in favor of error method and will be removed from avo version 3.0.0"
```
adrianthedev, line 149: GG!

Paul-Bob:
closed in favor of v2

PR #1306, opened 2022-10-11
update rails pg port from 5432 to 5433

ericdain:
# Description
While setting up a dev environment, docker-compose.yml is currently configured to expose PG on port 5433, but the rails config is looking for PG on default of 5432. Instructions under "Getting your local environment set up" say no changes are needed to database.yml if using Docker, but that's not true. So various steps of the dev set up instructions will fail, for example during bin/init:

```
======================
Preparing the database
======================

connection to server at "::1", port 5432 failed: Connection refused
	Is the server running on that host and accepting TCP/IP connections?
```

Maybe you want to stay away from the default PG port to avoid conflicts, so I adjusted rails to look for 5433. **Could align on 5432 instead, though.**

Not the most glamorous change, but will hopefully save other newbies like myself a bunch of time trying to figure out what is going on. Thanks for the otherwise great dev env setup instructions.

# Checklist:

## Manual review steps

Follow CONTRIBUTING.MD -> Getting your local environment set up
1. Fork & clone repo
2. Run `bin/init`


adrianthedev:
@ericdain would you be able to use the docker way if you update the `POSTGRES_PORT` environment variable?

I think the better change here is to update the docs and keep local postgres as default and enable developers to switch to docker if they want to.

ericdain:
@adrianthedev thanks for taking a look so quickly! And yep, we can alternatively keep local rails PG default. In this case, can we just update the docker-compose.yml to expose the PG container on that default 5432 that rails is looking for? I don't think we need to update docs or add env variable in this case.

I've updated code. Let me know what you think!

ericdain:
@adrianthedev I'm not sure if I properly kept my fork updated with the newest changes from the upstream avo repo because there is now an extra merge commit from upstream back into my fork branch (which I don't think is the ideal way to handle... I think rebase is preferred to keep history clean).

At this point I'm not sure of the best way to fix it other than maybe redoing the PR. Let me know what you think

adrianthedev:
Hey @ericdain, I updated the docs to reflect the default on local postgres and to mention that if you do want to use the docker instance with the provided configuration, then you'd have to change the database port on the app.

If one decides to actually use the docker postgres instance and wants to use the default `5432` port, they can update the `docker-compose.yml` file. I think it's easier this way and it supports the default way much better.

What do you think?

ericdain:
Hi @adrianthedev, thanks for the feedback. I think I must be missing it, but is there a reason the PG port in the docker-compose.yml can't be configured to "just work" without modification (5432:5432)? I guess you want to avoid conflict if a user happens to already have PG running on the default port of 5432, but in that case I wouldn't think the user would use docker-compose anyway. 

Other than that question, the added clarity in the docs will definitely help steer new folks who opt to use docker, so that works for me.

adrianthedev:
>  I guess you want to avoid conflict if a user happens to already have PG running on the default port of 5432

Yes. That's precisely why!

> but in that case I wouldn't think the user would use docker-compose anyway.

Hmm... I guess you're right. I was thinking that maybe they have a local postgres running on 5432, and if they use docker in parallel, they would conflict.
But you're right. They would only use one of them at a time.

adrianthedev:
Thanks for brainstorming this with me.

ericdain:
Very welcome, and thanks for a pleasant discussion on one of my first PRs!

adrianthedev:
Likewise. Keep up the streak 💪

ericdain:
tested and working like a charm now 🚀

PR #1311, opened 2022-10-12
Feature: `Time` field

### LINKED ISSUE #1289: Time field or `noCalendar` option for date_time
### Context

Hi Avo team! 👋 

We are creating an event resource on Avo with a field which should allow us to change the **time** of an event. For this we are using `as: :date_time` like in [this part](https://docs.avohq.io/2.0/fields/date_time.html) of the docs.
This gives us a flatpickr calendar with the date and time as well.

However, we would like to **not display the calendar** and only show the time picker (the date doesn't matter and could be today's date by default), to have a display like this 👇 
Screenshot 2022-10-06 at 13 21 29

### Possible solution

Flatpickr offers [the option](https://flatpickr.js.org/examples/#:~:text=2016%2D10%2D20-,Time%20Picker,-%23) but we saw that it is not possible to pass the `noCalendar` value inside the [date_field_controller.js](https://github.com/avo-hq/avo/blob/main/app/javascript/js/controllers/fields/date_field_controller.js#L13-L22).

Is there any other way to just show the time or would it be possible to add this option? (Also happy to contribute) 

Thank you for your help! 🙌

sarahlafer:
# Description
In this PR I added a new field called `Time` which works with the [Time Picker](https://flatpickr.js.org/examples/#:~:text=2016%2D10%2D20-,Time%20Picker,-%23) part of flatpickr to add a time to a resource. The field is similar to the `DateTime` field, but it doesn't show the calendar, only the time.

By default it looks like this 👇 
Screenshot 2022-10-12 at 12 11 31

You can also add some options to the field to show a different format (for example to not show the seconds, or to have a 24-hour time picker) 👇 
Screenshot 2022-10-12 at 12 12 19

It looks like this on mobile 👇 
Screenshot 2022-10-12 at 12 56 11

Other than the `DateTime` field, this new field doesn't convert the time to the time of the browser, but instead leaves it to how it was created. So when a course starts at 6pm it will also say so if I am in a different timezone (I simulated this by changing my browser's timezone) 👇 
Screenshot 2022-10-12 at 20 56 03

I also added the `starting_at` attribute to the Course Factory and added some system tests. Let me know if I should any more.

Fixes #1289

# Checklist:

- [x] I have made corresponding changes to the [documentation](https://github.com/avo-hq/docs) -> updated here avo-hq/docs#48

## Manual review steps

1. You can test this by going to a course resource (`/admin/resources/courses`) and adding and editing the time in the field `starting at`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,3 +71,4 @@ avo*.gem
 app/assets/builds
 
 TAGS
+brakeman_results.html
--- a/app/components/avo/fields/date_field/edit_component.html.erb
+++ b/app/components/avo/fields/date_field/edit_component.html.erb
@@ -7,6 +7,8 @@
       date_field_picker_format_value: @field.picker_format,
       date_field_first_day_of_week_value: @field.first_day_of_week,
       date_field_disable_mobile_value: @field.disable_mobile,
+      date_field_field_type_value: "date",
+      date_field_picker_options_value: @field.picker_options,
   } do %>
     <%= datetime_field "fake_#{@field.id}", "fake",
       value: @field.edit_formatted_value,
--- a/app/components/avo/fields/date_field/index_component.html.erb
+++ b/app/components/avo/fields/date_field/index_component.html.erb
@@ -3,6 +3,7 @@
     controller: "date-field",
     date_field_view_value: @view,
     date_field_format_value: @field.format,
+    date_field_field_type_value: "date",
   } do %>
     <%= @field.formatted_value %>
   <% end %>
--- a/app/components/avo/fields/date_field/show_component.html.erb
+++ b/app/components/avo/fields/date_field/show_component.html.erb
@@ -3,6 +3,7 @@
     controller: "date-field",
     date_field_view_value: @view,
     date_field_format_value: @field.format,
+    date_field_field_type_value: "date",
   } do %>
     <%= @field.formatted_value %>
   <% end %>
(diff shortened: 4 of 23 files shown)


adrianthedev:
Hey @sarahlafer. Thanks a bunch for this! Let me have a look at it these next couple of days.

I think I'd like to keep the option of having relative time thought. Probably on a switch.

sarahlafer:
Hi @adrianthedev. Great, thank you!
Okay, so should the time by default stay the same and then if you pass an option to the field you can make it relative? Should we then also display the timezone?

adrianthedev:
Hey @sarahlafer. Thanks for that.

I hope you don't mind, I took the liberty to add some updates:

 - change `absolute` to `relative`
 - add `picker_options` to `date`, `date_time`, and `time` fields
 - cleanup a few things on those fields

I also changed the migration to actually add a `time` column instead of a `date_time`.

Can you please check that everything works on your end? If all is well, we only need to change [the docs](https://github.com/avo-hq/vitepress-docs). Do you want to give that a try?

sarahlafer:
Hi! Oh nice, I love the way you refactored! And it still works on my end, yes 👍 

Yes I can have a look at the docs. I thought I had to change these docs so I opened [a PR](https://github.com/avo-hq/docs/pull/48) there (which still has the first version of the PR so without the `relative` option). Shall I close that PR and open one only on [here](https://github.com/avo-hq/vitepress-docs)?

adrianthedev:
Ok. This looks good to go. Let me have another look later and push it to `main`.

Thank you for championing this feature and for being a part of this community 🙏

sarahlafer:
Awesome 🚀 
Thank you for your help and the refactoring! It was great getting to know your code base a bit better and contributing 😊

PR #1316, opened 2022-10-17
fix: Norwegian translations

tersor:
# Description

Small fixes in Norwegian translations


adrianthedev:
Are you sure you branched out from `main`?

I think the easiest way of re-doing this PR is to do a rebranch from `main`. We did some changes to the translations.

Issue #1249, opened 2022-09-21
Presence of Avo forces full Zeitwerk app reload on any change


dhnaranjo:
### Describe the bug
Having Avo gem in your gemfile causes Zeitwerk to unload and reload every Class/Module on change to any Class/Module, rather than only those that are influenced. This causes, in my actual job application, ~10 second hang in development server on any change to a Ruby file.

### Steps to Reproduce
1. `rails new test_reloading --minimal`
2. `rails g scaffold should_reload`
3. `rails g scaffold do_not_reload`
4. `rails db:create db:migrate`
5. Add `Rails.autoloaders.log!` to bottom of `application.rb`
6. Run your server. Your server log should contain a block of lines similar to `Zeitwerk@rails.main: autoload set for ApplicationController, to be loaded from...`
7. visit `/should_reloads`, you will see some Zeitwerk activity.
8. Reload the page,  you will not see Zeitwerk activity.
9. Make a change to `ShouldReloadController#index` (tho any file would do honestly). The change does not matter, you could add `one = 1`. It's just the presence of a change.
10. Reload the page. Your output should include (not necessarily in this order):
```
Zeitwerk@rails.main: autoload for DoNotReload removed <<< THIS IS THE SIGNIFICANT BIT
Zeitwerk@rails.main: ShouldReload unloaded
Zeitwerk@rails.main: autoload set for ShouldReload, to be loaded from /Blah/blah/blah
```
11. Add Avo to gemfile & `bundle install`
12. Uncomment `require "active_job/railtie"` in `application.rb` (required by Avo)
13. Run steps 6 through 10, only step 10 server output should include (not necessarily in this order):
```
Zeitwerk@rails.main: DoNotReload unload <<< THIS IS THE SIGNIFICANT BIT
Zeitwerk@rails.main: ShouldReload unloaded
Zeitwerk@rails.main: autoload set for DoNotreload, to be loaded from /Blah/blah/blah
Zeitwerk@rails.main: autoload set for ShouldReload, to be loaded from /Blah/blah/blah
```

### Expected behavior & Actual behavior
Expected: Any code unrelated to a change should have its Zeitwerk autoload removed prior to code unloading.

Actual: All application code is unloaded and reloaded.

### System configuration
**Avo version**: 2.15.3

**Rails version**: 7.0.4

**Ruby version**: 3.1.2

**License type**:

 - [ ] Community
 - [X] Pro

#### Context

[...]

PR #1333, opened 2022-10-19
refactor: zeiwerk eager load resource dir

adrianthedev:
# Description

Fixes https://github.com/avo-hq/avo/issues/1249

# Checklist:


dhnaranjo:
Alright so. I ran three scenarios:
- Current 2.17 and Zeitwerk 2.6.1
- This fresh PR
- Without Avo

Scenario was:
- Start app
- Load page
- make trivial change to controller associated with page
- reload page

Line count:
- Current: 3413
- Fresh: 2017
- No Avo: 610

We're no longer reloading classes that aren't associated with Avo. But we're still reloading what appears to be every Avo-related class, as well as the app classes referenced by Avo classes, like a resources model and any classes the model calls, and so on.

We're also seeing a sort of race condition on the reloading, where concurrent requests all force a reload. The difference between Fresh and No Avo's log line count is that the whole refresh happens twice, once for HTML, once for ActionCable. I suspect that someone doin frontend components that call back async would get absolutely slammed by this.

So like, don't get me wrong, this helps, but I think it's time we weight what the eager loading of Resources it getting us. If the alternative to this is just maintaining a hard-coded and maintained list of Resource constants sign me up for **that**.

adrianthedev:
Thanks for this analysis. But how about your concrete use case with the assets. Does it still take 10 seconds for the auto reload to happen?

Also, can you create a demo app so I can troubleshoot this against a real use case?

adrianthedev:
> The difference between Fresh and No Avo's log line count is that the whole refresh happens twice.

@dhnaranjo I pushed a change where we eager load only on the main autoloader.

Can you please try again and test if that lowers the line count, and hopefully, the loading time?

adrianthedev:
Any input here @dhnaranjo?

dhnaranjo:
My fault I been occupied this week. I have not got a chance to look at this change, but here's a bit more response to the rest of the convo:

The page refresh loading slowness did persist in all of those test scenarios I listed above, except of course the control w/o Avo installed. The log line length of the requests was intended to illustrate relative refresh delay, as more lines represented more instances of a class being reloaded.

I wouldn't be able to provide a demo app that shows the delay times I'm experiencing as, you know, it took me ~40 hours a week for bout a year to get this one where it is. I could knock one together that shows the same double-dip reload on both http and websocket request. Is that what you're hoping for?

I'd really like to know what you think about my pitch of just... not doing any of this custom reloading, instead having a manifest of resources. No matter what, a resource being reloaded will always reload the controller and model associated with it, that's how Zeitwerk works. Especially with the models this can easily cascade. I've got 22 resource classes, so that's bare minimum if I were to reference no other classes in the related models an additional 66 classes reloaded.

I'd love to get some input from other folks using Avo on non-trivial applications.

dhnaranjo:
Oh shoot I just had an idea. So let's say for some reason we want to keep doing this reloading, tho I'm still advocating for not doing that. Why does it happen on every page refresh? Why not just the refresh of Avo pages? What do I care what Avo resources exist when I'm not interacting with it?

adrianthedev:
Hey @dhnaranjo, I'm bringing my baby home today, and I'll probably not have time for this for the next couple of days. 

I think your premise is not correct.

> Why does it happen on every page refresh?

It doesn't. It refreshes only when Avo content is loaded.

> Why not just the refresh of Avo pages?

It refreshes what it's imported in those Avo resources AFAIK. If you use a module that sets fields or settings inside a resource, you'd want that to be refreshed in a local environment.

>  having a manifest of resources

That's probably not going to happen. If Rails and other libraries use autoloading, I will not go backward on a list of files to be included. That's just not going to scale, and it's not going to be beneficial to anyone (including yourself).

I need your help to fix this in a sustainable way.

TBH, the issue from https://github.com/avo-hq/avo/issues/1249 seems fixed. The logs are passing as they should with `autoload for DoNotReload removed`.

Don't get me wrong. I'm not in the "it's fixed on my end so I'm going to close the issue" mood. I truly want to help you fix this, but you're going to have to help.

If there isn't a demo app where we can both see what happens, I can't reproduce it and can't fix it.

dhnaranjo:
> Hey @dhnaranjo, I'm bringing my baby home today, and I'll probably not have time for this for the next couple of days.

Hey that's rad,congratulations. I hope the baby n mother are both doing well. Don't even worry bout all this.

adrianthedev:
We are all good. Thank you for asking!
I will chime in here between the feeding time and other chores around the home.

adrianthedev:
Hey, @dhnaranjo. Let's wait for the new zeitwerk release, merge this as is, and see what the results are.

dhnaranjo:
> Hey, @dhnaranjo. Let's wait for the new zeitwerk release, merge this as is, and see what the results are.
> 
> 

I'm good with that. This change is absolutely an improvement, in that current state reloads every Avo class and this branch reloads only Resources and related classes. It is a significant speedup.

I knocked together a lil test app with appraisal scenarios for a control, main, and this branch that could be a useful tool for future work on this problem, but I'll look into that on my own time .

adrianthedev:
Hey @dhnaranjo I released https://rubygems.org/gems/avo/versions/2.18.1.pre.1.eagerloaddirs to try out the new loader.

dhnaranjo:
It feels much faster. Below is a test scenario in my actual app. We load my login page, modify a file to force a reload, load it again. Count iterations in 60 seconds. Current `2.18` is ~1.6x slower than `2.18.1`, which is in turn ~1.4x slower than Avo-less control. Manual observations have reload refreshes taking 400-600ms instead of 800-1000ms.

I still encounter occasional individual page refreshes that are significantly slower but I sure can't figure out why. But this is a big improvement.

```
  require "benchmark/ips"
  Benchmark.ips do |x|
    x.warmup = 0
    x.time = 60
    x.report("Avo reloads") do
      visit_sign_in
      modify_unrelated_controller
      visit_sign_in
    end
  end

Calculating -------------------------------------
   control    2.898  (± 0.0%) i/s -    169.000  in  60.031160s
   2.18.1      2.084  (± 0.0%) i/s -    124.000  in  60.542400s
   2.18         1.306  (± 0.0%) i/s -     64.000  in  60.022292s
```

adrianthedev:
Please, let's keep this optimization train going. Send in any other suggestions to make this issue go away for good. 
I'll definitely keep an eye out and try to improve this as we go.

Thanks for this nudge @dhnaranjo!

adrianthedev:
> I'd really like to know what you think about my pitch of just... not doing any of this custom reloading, instead having a manifest of resources. No matter what, a resource being reloaded will always reload the controller and model associated with it, that's how Zeitwerk works. Especially with the models this can easily cascade. I've got 22 resource classes, so that's bare minimum if I were to reference no other classes in the related models an additional 66 classes reloaded.

@dhnaranjo I just published [this PR](https://github.com/avo-hq/avo/pull/1584) that adds manual registration of resources if you want to give it a try

Issue #1331, opened 2022-10-19
search input got rendered multiple time


elassadi:
### Describe the bug

Search input are wrongly rendered multiple times after you hit the browser back button.

<img width="890" alt="image" src="https://user-images.githubusercontent.com/1938264/196647178-37b7a954-09f5-426f-ace9-f428414c85f5.png">

### Steps to Reproduce
![avo-search-input-bug](https://user-images.githubusercontent.com/1938264/196648595-31fb884d-b717-433e-b4e8-80447c3a5c80.gif)

1. use latest Avo release or checkout: [avo-demo](https://github.com/elassadi/avo-demo)
2. Go to any resource with search enabled 
3. Click on browser back button 
4. a second search input will be rendered just below the existing one

### Expected behavior & Actual behavior

Only one search button schould be visible

### System configuration
**Avo version**: latest 2.17

**Rails version**: 7.0.4

**Ruby version**:3.0.1

**License type**:

 - [x] Community
 - [ ] Pro

**Are you using Avo monkey patches or overriding views or view components?**

 - [ ] Yes. If so, please post code samples.
 - [x] No

#### Impact

 - [ ] High impact
 - [x] Medium impact
 - [ ] Low impact

#### Urgency

 - [ ] High urgency
 - [x] Medium urgency
 - [ ] low urgency

adrianthedev:
Yup. It's a bug. We gotta clean up the artifacts somehow.

If you have any experience with stimulus JS and custom js plugins, we're open for PRs.

Thanks for reporting.

rajaravivarma-r:
@adrianthedev This bug is preventing us from going with the upgrade. Could you please point me to a place where the *artifacts* are, so that I can give a shot at fixing it?

adrianthedev:
### Approach

The issue is that the js plugin (`autocomplete` found in `search_controller.js`) is not properly cleaning up the input when the user navigates away from the page.

What I'd look into is running the [`destroy`](https://www.algolia.com/doc/api-client/getting-started/update/javascript/?client=javascript#the-destroy-method) method on the autocomplete js plugin. That should reset things to how they were (supposedly).

If that doesn't work, I would do something like this. On the [`date_field_controller`](https://github.com/avo-hq/avo/blob/main/app/javascript/js/controllers/fields/date_field_controller.js#L77) is to use the `disconnect` method to "reset" the state back to what it was.

[...]

PR #1339, opened 2022-10-20
Reset autocomplete container before set up

rajaravivarma-r:
# Description
Algolia autocomplete.js creates duplicate search containers when navigating back to the page.

Fixes https://github.com/avo-hq/avo/issues/1331

# Checklist:

## Manual review steps

1. Open a resource index page with resource search. 
2. Navigate to an individual resource. 
3. Go back to the index page, by clicking on the browser back button. 
4. There should be one more `search box`.
5. Repeat from step 2 to step 4 and there will be as many search boxes as the number of times you did this.

### DIFF
--- a/app/javascript/js/controllers/search_controller.js
+++ b/app/javascript/js/controllers/search_controller.js
@@ -73,6 +73,8 @@ export default class extends Controller {
       Mousetrap.bind(['command+k', 'ctrl+k'], () => this.showSearchPanel())
     }
 
+    this.autocompleteTarget.innerHTML = ""
+
     autocomplete({
       container: this.autocompleteTarget,
       placeholder: this.translationKeys.placeholder,
--- a/app/javascript/js/controllers/search_controller.js
+++ b/app/javascript/js/controllers/search_controller.js
@@ -73,7 +73,8 @@ export default class extends Controller {
       Mousetrap.bind(['command+k', 'ctrl+k'], () => this.showSearchPanel())
     }
 
-    this.autocompleteTarget.innerHTML = ""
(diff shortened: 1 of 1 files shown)


rajaravivarma-r:
Hi @adrianthedev, looks like this fixes it. I can't think of any possible problems with this approach. It fixes the issue seems simple enough.

Can you think of any case where this would fail to work?

app/javascript/js/controllers/search_controller.js
```diff
@@ -73,6 +73,8 @@ export default class extends Controller {
       Mousetrap.bind(['command+k', 'ctrl+k'], () => this.showSearchPanel())
     }
 
+    this.autocompleteTarget.innerHTML = ""
```
adrianthedev, line 76:
```suggestion
    // This line fixes a bug where the search box would be duplicated on back navigation.
    this.autocompleteTarget.innerHTML = ''
```

adrianthedev:
This was simpler than I imagined 😛 
Not sure if it will impact anything else. Works well in manual tests.
Let's merge it and monitor.

Thank you for this!

rajaravivarma-r:
Ha ha. Same, I went through the autocomplete js documentation a lot before it popped up. Could please you do a release so that I can use this in our app?

rajaravivarma-r:
Thank you!

adrianthedev:
Release in https://rubygems.org/gems/avo/versions/2.17.1.pre.3 @rajaravivarma-r

rajaravivarma-r:
Thank you! I'll try that.

elassadi:
nice one :) 
I just tested the Fix. looks good .
thanks @rajaravivarma-r

PR #1347, opened 2022-10-23
feature: Extend field API with link_to_child_resource attribute

elassadi:
# **WIP**

Single table inheritance  support. 
This PR implement a new field option `link_to_child_resource` to allow 

- configuration of which resource should be used for any has_many field.
- configuration at the resource level 

# Description

 set this field option to true to enable child resources for a has_many field 
 
```ruby 
field :relatives, as: :has_many, link_to_child_resource: true 
```

 set this field option to true to enable child resources for a resource
 
```ruby 

class PersonResource < Avo::BaseResource
  self.link_to_child_resource = true
  self.authorization_policy= SomePolicy
 ....
end
```

An example is available in the dummy app:
https://github.com/avo-hq/avo/blob/582caa726faf81bd2a7abccd61326ea34f5dd3c0/spec/dummy/app/avo/resources/person_resource.rb#L16

Assuming relatives is an association of peoples and people model is an STI model. 

_any record of types Sibling or Spouse, which they all inherit from people, will use the Sibling or Spouse resource  
instead of the parent resource people._

image

# Checklist:

## Manual review steps

### DIFF
--- a/app/components/avo/base_component.rb
+++ b/app/components/avo/base_component.rb
@@ -43,4 +43,21 @@ class Avo::BaseComponent < ViewComponent::Base
   rescue
     nil
   end
+
+  def parent_or_child_resource
+    return @resource unless link_to_child_resource_is_enabled?
+
+    ::Avo::App.get_resource_by_model_name(@resource.model.class).dup
+  end
+
+  def link_to_child_resource_is_enabled?
+    return field_linked_to_child_resource? if @parent_resource
+
+    @resource.link_to_child_resource
+  end
+
+  def field_linked_to_child_resource?
+    field&.respond_to?(:link_to_child_resource) && field.link_to_child_resource
+  end
+
 end
--- a/app/components/avo/base_component.rb
+++ b/app/components/avo/base_component.rb
@@ -57,7 +57,6 @@ class Avo::BaseComponent < ViewComponent::Base
   end
 
   def field_linked_to_child_resource?
-    field&.respond_to?(:link_to_child_resource) && field.link_to_child_resource
+    field.present? && field.respond_to?(:link_to_child_resource) && field.link_to_child_resource
   end
-
 end
--- a/lib/avo/base_resource.rb
+++ b/lib/avo/base_resource.rb
@@ -48,6 +48,7 @@ module Avo
     class_attribute :record_selector, default: true
     class_attribute :keep_filters_panel_open, default: false
     class_attribute :extra_params
+    class_attribute :link_to_child_resource, default: false
 
     class << self
       delegate :t, to: ::I18n
--- a/spec/dummy/app/models/sibling.rb
+++ b/spec/dummy/app/models/sibling.rb
@@ -10,9 +10,6 @@
 #  user_id    :bigint
 #  person_id  :bigint
 #
-class Person < ApplicationRecord
-  belongs_to :user, optional: true
+class Sibling < Person
   has_many :spouses, foreign_key: :person_id
-  has_many :relatives, foreign_key: :person_id, class_name: "Person"
-  has_many :peoples, foreign_key: :person_id, class_name: "Person"
 end
--- a/app/components/avo/fields/index_component.rb
+++ b/app/components/avo/fields/index_component.rb
@@ -1,18 +1,19 @@
 # frozen_string_literal: true
 
-class Avo::Fields::IndexComponent < ViewComponent::Base
+class Avo::Fields::IndexComponent < Avo::BaseComponent
   include Avo::ResourcesHelper
 
   attr_reader :parent_resource
   attr_reader :view
 
-  def initialize(field: nil, resource: nil, index: 0, parent_model: nil, parent_resource: nil)
+  def initialize(field: nil, resource: nil, reflection: nil, index: 0, parent_model: nil, parent_resource: nil)
     @field = field
     @resource = resource
     @index = index
     @parent_model = parent_model
     @parent_resource = parent_resource
     @view = :index
+    @reflection = reflection
   end
 
   def resource_view_path
@@ -25,7 +26,7 @@ class Avo::Fields::IndexComponent < ViewComponent::Base
       }
     end
 
-    helpers.resource_view_path(model: @resource.model, resource: @resource, **args)
+    helpers.resource_view_path(model: @resource.model, resource: parent_or_child_resource, **args)
   end
 
   def field_wrapper_args
--- a/app/components/avo/index/grid_item_component.rb
+++ b/app/components/avo/index/grid_item_component.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class Avo::Index::GridItemComponent < ViewComponent::Base
+class Avo::Index::GridItemComponent < Avo::BaseComponent
   include Avo::ResourcesHelper
 
   attr_reader :parent_resource
@@ -37,6 +37,6 @@ class Avo::Index::GridItemComponent < ViewComponent::Base
       }
     end
 
-    helpers.resource_view_path(model: @resource.model, resource: @resource, **args)
+    helpers.resource_view_path(model: @resource.model, resource: parent_or_child_resource, **args)
   end
 end
--- a/app/components/avo/index/resource_controls_component.rb
+++ b/app/components/avo/index/resource_controls_component.rb
@@ -41,7 +41,7 @@ class Avo::Index::ResourceControlsComponent < Avo::ResourceComponent
       }
     end
 
-    helpers.resource_path(model: @resource.model, resource: @resource, **args)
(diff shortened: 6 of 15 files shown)


adrianthedev:
How about instead of adding the configuration on a global level (in the initializer), we add it on the association field level?

```ruby
field :animals, as: :has_many, link_to_child_resource: true
```

This way, you can have some associations link to the parent and some link to the child.

elassadi:
Hi @adrianthedev 

Maybe I am missing something. My approach will be
- define an attribute in the  HasBaseField Class
`        @link_to_child_resource = args[:link_to_child_resource] || false
`
- detect and return the appropriate resource in the StiResourcesHelper according to the field attributes.


As far as I see I can't access the field attribute in my included helper.
Any idea suggestion or better approach  how to pass the field attribute down?

adrianthedev:
You can access the configuration using the [`field` getter](https://github.com/avo-hq/avo/blob/main/app/components/avo/base_component.rb#L13).

```ruby
if field.present?
  field.link_to_child_resource # do something with it
end
```

elassadi:
This was exactly  my initial, though. 

Unfortunately, the filed I receive in the resource_controller_component etc. is the ID field of the associated resource. 
the field id does not hold any information about the association it belongs to.

adrianthedev:
Ok. I see that.

Why do you need to patch that component? So the `link_to_resource` option to work, right?
Otherwise, only the resource controls would need to be patched, right?

adrianthedev:
We can do this:

 - make `Avo::Fields::IndexComponent` inherit from `Avo::BaseComponent`
 - make `Avo::Fields::IndexComponent` to take a `reflection` keyword arg
 - pass the `reflection` in the `table_row_component.html.erb`
 - now we have access to the `field`🙌

elassadi:
Hi @adrianthedev 
Sofar we are ready to got with this one. I think is flexible enough to allow setting the behavior of has many STI resources and a regular resource.

.gitignore
```diff
@@ -72,3 +72,4 @@ app/assets/builds
 
 TAGS
 brakeman_results.html
+/spec/dummy/.env.development
```
adrianthedev, line 75: Is this necesarry for this PR?

app/components/avo/base_component.rb
```diff
@@ -1,46 +1,66 @@
 # frozen_string_literal: true
 
-class Avo::BaseComponent < ViewComponent::Base
-  include Turbo::FramesHelper
+module Avo
```
adrianthedev, line 3: Hey, it seems that a few styling changes ended up in this PR.
Could I ask you to revert them as they were? It makes evaluating this PR much more difficult.

spec/dummy/app/models/sibling.rb
```diff
@@ -0,0 +1,15 @@
...
+#
+#  id         :bigint           not null, primary key
+#  name       :string
+#  type       :string
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#  user_id    :bigint
+#  person_id  :bigint
+#
+class Sibling < Person
```
adrianthedev, line 13: Doesn't `Spouse` do the same thing?

adrianthedev (commented):
I left a few comments below. Could you address those please before I review it? They clutter the diff.

Thank you!

spec/dummy/app/models/sibling.rb
elassadi: Sure it does , but I wanted to make more clearer that you can address more than one resource in the same has_many field

lib/avo/base_resource.rb
```diff
@@ -48,6 +48,7 @@ class BaseResource
     class_attribute :record_selector, default: true
     class_attribute :keep_filters_panel_open, default: false
     class_attribute :extra_params
+    class_attribute :link_to_child_resource, default: false
```
adrianthedev, line 51: Do we still need this attribute on the base resource? I see we're not mutating it anywhere.

app/components/avo/base_component.rb
```diff
@@ -43,4 +43,21 @@ def reflection_parent_resource
...
+  def parent_or_child_resource
+    return @resource unless link_to_child_resource_is_enabled?
+
+    ::Avo::App.get_resource_by_model_name(@resource.model.class).dup
+  end
+
+  def link_to_child_resource_is_enabled?
+    return field_linked_to_child_resource? if @parent_resource
+
+    @resource.link_to_child_resource
```
adrianthedev, line 56: This seems to always return `false`. Can we rewrite it somehow and remove the dependency to `@resource.link_to_child_resource`?

adrianthedev (commented):
There's one more thing I'm not sure I understood. Do we still need the `link_to_child_resource` option on the base resource? If not, can we remove it?

Also, could you add some tests? I'd hate it to break in the future and not know about it.

Otherwise, the PR looks amazing. Thank you!

app/components/avo/base_component.rb
elassadi: it's still needed. This option allows the global linking of parent resource instead of child resource
when the resource is targeted directly without the usage of a field.

lib/avo/base_resource.rb
elassadi: Related to my other comment. Its not mutated. But its hold the resource global configuration.
if enabled ( value: true)

app/components/avo/base_component.rb
adrianthedev: ok. and where do we set it to true? I can't seem to find that.

or maybe I didn't really understood the use-case. can you give me an example of when we'd set it to `true`?

elassadi: I just pushed the changes in the people resource to see how to enable this feature on resource level. 
with the option toggled true.

Navigate to http://127.0.0.1:3000/admin/resources/people and edit or show the records.
It will use the child resource instead of the parent resource automatically. 
Hope it made it clear ?

adrianthedev: Ok, I see that now, and how it works. However, I'm pondering if this configuration switch is really worth it that much.

I'll explain what I mean.

1. I try to keep only the necessary configuration on each entity (`BaseResource`). We added and added things (and we'll add more in the future), that might make it more difficult to manage and it becomes a "God object". So efficiency is taken into account here.
1. Do you really need to toggle that switch on the resource level? Are the savings worth it? I mean, if you don't declare it on the resource you'll declare it on the field/fields. How many fields are there anyway? Maybe 10 in a truly large app? But usually just one or two places in a regular app?
1. Is this API really helpful. One developer might be used to declaring this option on the field and want to make some changes, but another developer in the team declared it on the resource. So developer 1. is trying to troubleshoot why it's not working like they are used to it because they can't seem to find it on the field. I know I hit this snag in the past.

I guess, what I'm trying to say is that I want to avoid having two ways of doing the same thing. It adds more maintenance, and the gains aren't really that high.

I'd love to hear your thoughts about this.

elassadi: Hi Adrian 

Thanks for your comment, I really appreciate the discussion and your patience.

Just to be clear:
**The field option link_to_child_resource will not override the resource option**


However, I disagree with your argumentation and I will try to answer your questions, and explain my point. 

1. I see what you mean, and I totally agree with the approach to minimize the configuration.
2. **Yes**. I don't see any other way to set the resource to link to the child resource instead of the parent resource (which is now the default behavior), when no fields declaration is involved at all.
If you have any idea how to force for example  ->//admin/resources/people to show a list of records, each linked to their respective child resources, we can do it without the resource level configuration, I am total not keen about it. 

3. Your assumption here, that the filed option will override the resource option, is not correct. 

- if developer 1 set the resource option to true, it will be used **only** when we point the browser to show page of the parent resource.
- if the second developer set the field option, we will respect the option set in the field or use the default behavior.

adrianthedev: Oh, ok. Your answer on point 2. really made it click for me. I always saw it like, "you only need to split the resources on the `has_many` field level, not on the "root" `/admin/resources/people`.

Thanks for explaining! That makes sense.

iainbeeston:
Is there any documentation on how to use this? I can't find any on [docs.avohq.io](https://docs.avohq.io). I'd like to try it with delegated types (rather than STI) but I'm not sure if I'm doing it correctly

adrianthedev:
Hmm, no docs. I hate it when this happens. I'll write them right now.

You can use this in two ways:

1. Declare this option on the whole resource (`self.link_to_child_resource = true` on the resource). See in the dummy app the person resource has STI with `SpouseResource` and `ChildResource`, so when you go to the index page and click on the view button on one of the rows you will actually be redirected to a Child or Spouse resource instead of a Person resource.
2. Use it on a `has_many` field (`field :peoples, as: :has_many, hide_search_input: true, link_to_child_resource: false`). On the `PersonResource` see the `has_many` people field. it will have the same effect but in the `has_many` field listing.

PR #1378, opened 2022-10-29
fix: visibility issues for tabs panels and sidebars

### LINKED ISSUE #1352: Visibility of fields under tabs not working as intended
### Describe the bug
1) `visible` attribute on field definition has no effect when inside tabs.
2) `only_on` is not working for fields that are inside panels within tabs.

### Steps to Reproduce

1)

- Add a field under a tab, with `visible` block 
image
- Check show page:
image

2) 

- Add a field under a panel within a tab, with `only_on` attribute
image
- Check show page:
image

Expected: visibility checks are respected
Actual: are not respected

### Notes
I have some guesses as what the causes could be. I am looking at the `visible_items` method in `tab.rb` (let me know if I am looking at the wrong place):
image
1) This method seems to only check for visibility for a view, not the fields' `visible` block
2) When there is no panel, `items` represents **fields**. However, when a panel is inside a tab `items` represents **panels**, so it does not do the view visibility checks. In this case the fields are found _one layer further_ `items.first.items_holder.items`

### System configuration
**Avo version**: 2.15.3

**Rails version**: 6.1.6.1

**Ruby version**: 3.0.2

#### Impact

 - [ ] High impact
 - [] Medium impact
 - [x] Low impact

#### Urgency

 - [ ] High urgency
 - [ ] Medium urgency
 - [x] low urgency

Paul-Bob:
# Description

Fixes #1352

# Checklist:

## Manual review steps

1. Step 1
1. Step 2


spec/dummy/app/avo/resources/spouse_resource.rb
```diff
@@ -9,4 +9,34 @@ class SpouseResource < Avo::BaseResource
 
   field :id, as: :id
   field :name, as: :text
+
+  # Uncomment this to test visibility inside tabs, panels and sidebar
+  # tabs do
+  #   field :hidden_field_inside_tabs, as: :text, visible: -> (resource:) { false }
+
+  #   tab "hidden testing visibility inside tabs" do
+  #     field :hidden_field_inside_tabs_inside_tab, as: :text, visible: -> (resource:) { false }
```
adrianthedev, line 18: This is still visible on my end.

spec/dummy/app/avo/resources/spouse_resource.rb
```diff
@@ -9,4 +9,34 @@ class SpouseResource < Avo::BaseResource
...
+
+  #     panel do
+  #       field :hidden_field_inside_tabs_inside_tab_inside_panel, as: :text, visible: -> (resource:) { false }
+  #     end
+  #   end
+  # end
+
+  # sidebar do 
+  #   field :hidden_field_inside_sidebar, as: :text, visible: -> (resource:) { false }
```
adrianthedev, line 27: This is visible too.

adrianthedev (commented):
GG on the visibility refactor for panel and tab 👌

I think we should have some tests for this feature.

spec/dummy/app/avo/resources/spouse_resource.rb
Paul-Bob: did you restart the local server before trying this one?

PR #1379, opened 2022-10-29
feature: persistent filters

elassadi:
Persistent Filters

# Description

The feature you may be waiting for is here :). 
This PR will make the availability of persistent filters in Avo possible. 
**All Resources will now automatically retain the filters, that you have setup.** 

you no longer need to repeat the same steps each time you return to the resource. With this feature, you will be able to pick up right where you left off!

To get persistent filters in action, simply enable the configuration in your avo initializer.

```ruby
config.cache_resource_filters = true
```
Or if you need more flexibility you are able to enable this feature based on user profil, resources or any other requirement. Just pass a lambda instead of a boolean value. 
Here an example

```ruby 
config.cache_resource_filters = ->(resource:) do
  Avo::App.current_user.cache_resource_filters?
end
```

 

# Checklist:


adrianthedev:
Thank you, @elassadi. Let me test it out in the next couple of days.

adrianthedev:
This PR is great 👌 @elassadi.

There's one thing that I'd do. Can we make `Avo::FiltersHelper` a concern?
The reasoning is that a helper is something that you use in the view layer and this naming might confuse people.

adrianthedev:
BTW, I added the `CommentBodyFilter` to be able to test this feature across multiple resources and `has_many` associations

elassadi:
> 

agree

adrianthedev:
Good job @elassadi! Thank you for your time and contribution!

PR #1380, opened 2022-10-31
Pass resource to default

iainbeeston:
# Description

At the moment if a block is passed to default it is executed in isolation. It would be very useful to be able to access the model or the resource in default so that a default can be set that is based on parameters or other attributes (similar to how we can with visible).

Fixes # (issue)

# Checklist:

## Manual review steps

1. Set a default value for an attribute that takes a parameter and uses the resource to calculate a default value
2. Make sure this works and doesn't crash avo

### DIFF
--- a/lib/avo/fields/base_field.rb
+++ b/lib/avo/fields/base_field.rb
@@ -170,7 +170,7 @@ module Avo
         if @view.in?([:new, :create]) || @action.present?
           if default.present?
             final_value = if default.respond_to?(:call)
-              default.call
+              default.call resource: @resource
             else
               default
             end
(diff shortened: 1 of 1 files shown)


adrianthedev:
Thanks for this. We're trying to reconcile all (or most) blocks to be run inside [execution hosts](https://docs.avohq.io/2.0/evaluation-hosts.html).

Would you be available to apply the same logic to this PR?

Please check out [the PR](https://github.com/avo-hq/avo/commit/2a74ac08c86e12882cc0b0f8c6e22258cdce3769#diff-cb1043e781e1318272e0c86c8ea6ae768cd0a231a218e93d96c03847455e7877R9) where we refactor the `required` attribute, or [this one](https://github.com/avo-hq/avo/commit/b0a533bd29bd2d7bc5a668be9b82a06142e8a44d#diff-5f9106c27c1d4ead3e9471af835c410f44fe82c3fd628c265b39c4566dd00704R9) with `readonly`.

adrianthedev:
Closing in favor of https://github.com/avo-hq/avo/pull/1389.

It will be out in the next release. Thanks @iainbeeston.

PR #1381, opened 2022-10-31
Support virtual_attributes in key_value field

rickychilcott:
# Description
Key value fields only appear to work with ActiveRecord attributes. In order to support virtual attributes (and real attributes) a subtle code change is needed.

# Checklist:

## Manual review steps

1. Step 1
1. Step 2


adrianthedev:
Thank you @rickychilcott

PR #1398, opened 2022-11-04
Fallback to resources class / link to child resource feature

elassadi:
# Description

Link to child resources
 
A small bug. 
if the parent resource is the same as the base class or if the child class is missing we fallback to the parent class. 
We may as well raise an exception ? 

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/app/components/avo/base_component.rb
+++ b/app/components/avo/base_component.rb
@@ -46,8 +46,9 @@ class Avo::BaseComponent < ViewComponent::Base
 
   def parent_or_child_resource
     return @resource unless link_to_child_resource_is_enabled?
+    return @resource if @resource.model.class.base_class == @resource.model.class
 
-    ::Avo::App.get_resource_by_model_name(@resource.model.class).dup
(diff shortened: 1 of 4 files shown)


adrianthedev:
Hey @elassadi. Can you share an example where this bug occurs in the dummy app?

elassadi:
I just prepared https://github.com/elassadi/avo/tree/demo/reference-to-base-class  to show the case

adrianthedev:
OK! Got it! I see the issue and the fix now. Thank you!

Can I ask you to add a spec for this so we can bullet proof it?

elassadi:
I don't see any useful specs to cover a missing Model.
I am not even sure if we should not let the exception being raised
But since it will affect parent resources as well. I adopt to use a rescue (fallback) in our case. 
If I got some time, I will thing about adding meaningful specs, for now I dont see any

adrianthedev:
I think it would be enough adding that `Brother` model, seed the DB with a record, and make sure the page does not crash like before.

I'll look into it and see if I can add one.

PR #1403, opened 2022-11-06
feature: arguments on filters and actions

### LINKED ISSUE #1247: Is it possible to send parameters to filters and actions?
### Context
I'd like to have a way to send parameters to filters and actions, so I can work on generic actions and filters. Sometimes, when dealing with a lot of tables, building a filter/action for every table becomes way too time consuming and hurts `DRY`.

For example, something like this:

```ruby
action ExportToCsv, fields: [:name, :id]
```

```ruby
filter LikeFilter, fields: [:name, :description]
```

### System configuration
**Avo version**: 2.15.3

**Rails version**: 7.0.3.1

**Ruby version**: 3.1.2

**License type**:

 - [ ] Community
 - [x] Pro

Paul-Bob:
# Description

Fixes #1247

Now, actions and filters are able to be initialized with arguments!

For example:

```ruby
  filter NameFilter, arguments: {
    case_insensitive: true
  }
```

Inside `NameFilter`:
```ruby
  def apply(request, query, value)
    if arguments[:case_insensitive]
      query.where("LOWER(name) LIKE ?", "%#{value.downcase}%")
    else
      query.where("name LIKE ?", "%#{value}%")
    end
  end
```

For actions:
```ruby
  action DummyAction, arguments: {
    special_message: true
  }
```

Inside `DummyAction`:
```ruby
 def handle(**args)
    # Do something here
    if arguments[:special_message]
      succeed "I love 🥑"
    else
      succeed "Success response ✌️"
    end

    warn "Warning response ✌️"
    inform "Info response ✌️"
    fail "Error response ✌️"
  end
```

Inside the NameFilter and the DummyAction visible blocks you can access these arguments too. That make filters and actions even more customizable for each resource!

# Checklist:

## Manual review steps

1. Step 1
1. Step 2


adrianthedev:
It just occurred to me. Aren't these overriding the [filter options](https://docs.avohq.io/2.0/filters.html#boolean-filter)?

adrianthedev:
Especially in the `option` method. I see the use case where you would want to override the filter options based on the options you pass in the resource.

Paul-Bob:
> It just occurred to me. Aren't these overriding the [filter options](https://docs.avohq.io/2.0/filters.html#boolean-filter)?

Actually it do override, i didn't noticed that :S 
We should change that keyword, any suggestion?

adrianthedev:
How about we use `args` for filters and `options` for the rest (cards and actions).

Paul-Bob:
> Wanna push some docs too?

Yes!

PR #1408, opened 2022-11-07
feature: make sign-out path name configurable

### LINKED ISSUE #1407: Add ability to completely customize destroy_user_session_path
### Feature
Loving Avo so far! We ran into a small issue with the sign-out menu item though. We don't use Devise, and so our sign-out path does not follow the pattern of `destroy_*_session_path`. This means that we cannot configure the link by just setting `current_user_resource_name`. We would benefit from having the ability to customize the entire `destroy_user_session_path`.

Would you be interested in a PR to add this configuration setting? If so, I'd be happy to work on it.

### Current workarounds
We are currently working around this issue by monkey patching `Avo::SidebarProfileComponent` to completely override the `destroy_user_session_path` method. Something like this:
```ruby
require "avo/sidebar_profile_component"

class Avo::SidebarProfileComponent
  def destroy_user_session_path
    :logout_path
  end
end
```

The only other approach I can think of would be to rename the route to match the currently supported pattern.

nkriege:
# Description
Added a new configuration option called `sign_out_path_name`. This allows callers to completely customize the name of the path that we use in our sign-out menu item.

Note that I have also turned on view component tests so that I could write some unit tests for this. If those were previously off for a reason, though, I can revert this part of the change.

Fixes #1407 

# Checklist:

## Manual review steps
1. Verify that the sign-out menu item works if the sign-out path in your app uses the standard name of `destroy_user_session_path`
1. Verify that the sign-out menu item works if your app has a custom name for its sign-out path which you pass into `config.sign_out_path_name`

### DIFF
--- a/app/components/avo/sidebar_profile_component.rb
+++ b/app/components/avo/sidebar_profile_component.rb
@@ -33,8 +33,11 @@ class Avo::SidebarProfileComponent < ViewComponent::Base
     end
   end
 
+  # The name of the path for destroying the current user session (not the path itself).
   def destroy_user_session_path
-    "destroy_#{Avo.configuration.current_user_resource_name}_session_path".to_sym
+    # If `destroy_user_session_path` is configured, use it. Otherwise construct the
+    # path name based on `current_user_resource_name`.
+    (Avo.configuration.destroy_user_session_path || "destroy_#{Avo.configuration.current_user_resource_name}_session_path").to_sym
   end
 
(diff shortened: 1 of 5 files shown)


adrianthedev:
Thank you for the contribution @nkriege.

The docs are [here](https://github.com/avo-hq/avodocs) for updates.

It's getting late here but I'll review this in the morning and release it in tomorrow's release.

PR #1409, opened 2022-11-07
refactor: replace breadcrumbs_on_rails with custom breadcrumbs implementation

### LINKED ISSUE #1394: Installing Avo breaks projects already using breadcrumbs
After adding Avo to a project where I already used a bredcrumbs gem ([Gretel](https://github.com/kzkn/gretel)), the project breaks, because Avo pulls `breadcrumbs_on_rails`, which defines same methods as Gretel, but with different arity.

This happens because `breadcrumbs_on_rails` [fearlessly patches](https://github.com/weppos/breadcrumbs_on_rails/blob/12962e873cce8864a53e01d3d54dc0a299897f5e/lib/breadcrumbs_on_rails/railtie.rb) `ActionController::Base` and as a result it's not possible to limit its usage only to Avo controllers[^1].

Looking at the code of `breadcrumbs_on_rails`, which is quite short, I think it would be best in Avo internalized it, renamed `breadcrumbs`-like methods to `avo_breadcrumbs` and removed the dependency on this external library. I think I can prepare a pull request if you approve this direction.

[^1]: Gretel kind of does the same, but with views, so they both are not the greatest ecosystem citizens in this regard.

katafrakt:
# Description

`breadcrumbs_on_rails` gem caused incompatibility with Rails projects which used other breadcrumbs gems using same method names. Instead, an internal implementation based on `breadcrumbs_on_rails` was created, using method names with "avo" (`add_avo_breadcrumb`, `avo_breadcrumbs`...).

The implementation was also stripped down to only use features that Avo uses.

### Questions

* This does not add any tests on its own. I'm willing to write some, but there's not any place where it fits. Should I create `spec/concerns`?
* Currently `lib/avo/concerns` only contains model concerns (I think), so not sure if it's okay to mix in a controller concern in there.

Fixes #1394 

# Checklist:

## Manual review steps

1. In example app, go to any subsection (e.g. `/admin/resources/products`)
2. Verify that breadcrumbs display correctly and are clickable

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -121,8 +121,7 @@ gem "appraisal"
 
 gem 'meta-tags'
 
-gem 'breadcrumbs_on_rails'
-
+gem 'manifester'
 
 # Search
 gem 'ransack'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -6,7 +6,6 @@ PATH
       active_link_to
       activerecord (>= 6.0)
       addressable
-      breadcrumbs_on_rails
       docile
       dry-initializer
       httparty
@@ -127,8 +126,6 @@ GEM
     bootsnap (1.13.0)
       msgpack (~> 1.2)
     brakeman (5.3.1)
-    breadcrumbs_on_rails (4.1.0)
-      railties (>= 5.0)
     builder (3.2.4)
     bump (0.10.0)
     bundler-integrity (1.0.9)
@@ -238,6 +235,9 @@ GEM
       nokogiri (>= 1.5.9)
     mail (2.7.1)
       mini_mime (>= 0.1.1)
+    manifester (0.1.8)
+      rails (>= 6.0)
+      zeitwerk
     marcel (1.0.2)
     matrix (0.4.2)
     meta-tags (2.17.0)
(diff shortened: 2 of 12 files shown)


katafrakt:
I should probably also mentioned that I tried this out with a problematic app from #1394 and it no longer conflicts with the other gem.

adrianthedev:
Hey @katafrakt. I'm curious if we **need** to rename the `add_breadcrumb` method to `add_avo_breadcrumb`.

This is a breaking change for all apps as it's present in the default avo.rb initializer.

adrianthedev:
Also, I see this issue in a manual test.

CleanShot 2022-11-08 at 13 55 26

katafrakt:
Thanks for checking @adrianthedev!

There was indeed a bug with appending breadcrumb on top of previous ones - and I think I fixed that.

As for using `add_avo_breadcrumbs`, good point with initializer and breaking changes. I checked a few existing breadcrumbs implementations for Rails and they are not using `add_breadcrumb` method (except, of course, `breadcrumbs_on_rails`, but the implementation should be compatible). After that I reverted the name of the method back to `add_breadcrumb`.

adrianthedev:
That's perfect! Thanks for understanding @katafrakt, and for making the changes back 🙏

PR #1431, opened 2022-11-12
fix: Fix problem with HasHTMLAttributes memoization

### LINKED ISSUE #1430: bug: block notatation for html attributes is only calling block once
### Describe the bug

When using the [block notation](https://docs.avohq.io/2.0/stimulus-integration.html#the-block-notation) for adding html attributes to a field, the block is only evaluated once. It's evaluated on the first record but the output is applied to all records.

### Steps to Reproduce

1. Make the change to `TeamResource` described below.
2. Go to `/admin/resources/teams`

Each team name should be a different color. Instead, they are all the color of the first team. See screenshot below.

### Models and resource files

Change the `name` field on `TeamResource` to this:
 
```ruby
  field :name, as: :text, sortable: true, html: -> do
    index do
      wrapper do
        style do
          if record.color
            "color: #{record.color}"
          end
        end
      end
    end
  end
```

### System configuration
**Avo version**: 2.19.0

**Rails version**: 7.0.4

**Ruby version**: 3.1.2

**License type**:

 - [ ] Community
 - [x] Pro

**Are you using Avo monkey patches or overriding views or view components?**

 - [x] No

image

#### Impact

 - [ ] High impact
 - [ ] Medium impact
 - [x] Low impact

#### Urgency

 - [ ] High urgency
 - [ ] Medium urgency
 - [x] low urgency

nkriege:
# Description

`parsed_html` was being memoized at the field level, but it depends on the record. Since the field is only instantiated once, that resulted in cached results from the first record being applied to all records.

I have removed the memoization and refactored the code slightly so that the parsing will happen once per call to `get_html`. There might be a way to reintroduce memoization in a more robust way, but I'm not sure if the speed increase would be enough to justify the complexity. Let me know what you think.

I updated `TeamResource` to exercise this code and added a test.

Fixes #1430

# Checklist:

## Manual review steps

1. Notice that `TeamResource` has been modified to generate a style for each team name on the index page.
1. Go to /admin/resources/teams

Before this change all rows used the color of the first team. Now they all have the appropriate color.

image

### DIFF
--- a/lib/avo/concerns/has_html_attributes.rb
+++ b/lib/avo/concerns/has_html_attributes.rb
@@ -16,12 +16,14 @@ module Avo
           default_attribute_value name
         end
 
-        attributes = if html_builder.is_a? Hash
-          get_html_from_hash name, element: element, view: view
-        elsif html_builder.is_a? Avo::HTML::Builder
-          get_html_from_block name, element: element, view: view
-        elsif html_builder.nil?
-          # Handle empty html_builder by returning an empty state
+        parsed = parse_html
+
+        attributes = if parsed.is_a? Hash
+          get_html_from_hash name, element: element, hash: parsed, view: view
+        elsif parsed.is_a? Avo::HTML::Builder
+          get_html_from_block name, element: element, html_builder: parsed, view: view
+        elsif parsed.nil?
+          # Handle empty parsed by returning an empty state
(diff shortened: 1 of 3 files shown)


adrianthedev:
Amazing PR! Thank you for digging in an fixing this! We appreciate the help!

PR #1433, opened 2022-11-13
fix: i18n missing translations

### LINKED ISSUE #1216: I18n missing translations
### Describe the issue and screenshots

- [x] Can't translate pagination details.

image

- [x] Can't translate "Filter by"

image

- [x] Despite changing the translations for search placeholder it don't apply, don't know if it's because I have some bad configuration

image

- [x] Using pluralized `one` instead `zero` on `no_item_found` and on `no_related_item_found` translation
 ```ruby
error_log:
        zero: "log de erro zero"
        one: "log de erro one"
        other: "logs de erro other"
```
image

### System configuration
**Avo version**:
2.14.2

#### Impact

 - [ ] High impact
 - [ ] Medium impact
 - [x] Low impact

#### Urgency

 - [ ] High urgency
 - [ ] Medium urgency
 - [x] low urgency

### PR DESCRIPTION
**Pull request by Paul-Bob**
# Description

Fixes #1216

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### CODE REVIEW

File: app/controllers/avo/base_controller.rb [hunk 1]
```diff
@@ -14,6 +14,8 @@ class BaseController < ApplicationController
     before_action :fill_model, only: [:create, :update]
     # Don't run base authorizations for associations
     before_action :authorize_base_action, if: -> { controller_name != "associations" }
+    # Set pagy locale from params or from avo configuration, if both nil locale = "en"
+    before_action { @pagy_locale = params[:force_locale] || Avo.configuration.locale || "en" }
```
Paul-Bob, line 18: only_index

lib/generators/avo/templates/locales/pagy/nn.yml
```diff
@@ -0,0 +1,17 @@
+# :one_other pluralization (see https://github.com/ddnexus/pagy/blob/master/lib/locales/utils/p11n.rb)
+# TODO: Need translation
```
Paul-Bob, line 2: make ticket

lib/avo/base_resource.rb
```diff
@@ -50,6 +50,7 @@ class BaseResource
     class_attribute :keep_filters_panel_open, default: false
     class_attribute :extra_params
     class_attribute :link_to_child_resource, default: false
+    class_attribute :empty_state_name
```
Paul-Bob, line 53: Why empty_state_name? For each language we can have it plural or singular, for example: 
EN: No related comments found.
RO: Nu s-au găsit comentarii relationate. SAU Nici un comentariu relationat gasit.
PT: Nenhum comentário relacionado encontrado.

So it make sense to make it custom. This way some users can use Localization, One, Zero, Other or even a custom string....

adrianthedev: It seems a bit too much to add a base resource option to translate something. Can't we use the translation helpers? Why doesn't `zero` help? I see that you added `self.empty_state_name = "comments"` somewhere below. Isn't the value of `zero` "comments" in the translation?

lib/generators/avo/templates/locales/pagy/ro.yml
```diff
@@ -0,0 +1,17 @@
...
+
+ro:
+  pagy:
+    item_name:
+      one: "un record"
+      other: "recorduri"
+    nav:
+      prev: "&lsaquo;&nbsp;Înapoi"
+      next: "Înainte&nbsp;&rsaquo;"
+      gap: "&hellip;"
```
adrianthedev, line 11: I don't see these working in the UI

CleanShot 2022-11-26 at 15 37 42

spec/dummy/app/avo/resources/photo_comment_resource.rb
```diff
@@ -5,6 +5,7 @@ class PhotoCommentResource < Avo::BaseResource
   self.includes = [:user]
   self.model_class = ::Comment
   self.authorization_policy = PhotoCommentPolicy
+  self.empty_state_name = "comments"
```
adrianthedev, line 8: this definitely does not belong here.

adrianthedev (changes requested):

lib/avo/base_resource.rb
Paul-Bob: It is too much, and unnecessary, i don't know why i think it was a good idea while implementing it...

lib/generators/avo/templates/locales/pagy/ro.yml
Paul-Bob: Did you restart the server after changing to this branch?

app/components/avo/index/resource_grid_component.html.erb
```diff
@@ -9,6 +9,6 @@
   </div>
 <% else %>
   <div class="bg-white rounded shadow-panel">
-    <%= helpers.empty_state resource_name: @resource.name.downcase.pluralize, related_name: params[:related_name], view_type: :grid %>
+    <%= helpers.empty_state resource_name: @resource.name(count: 0).downcase, related_name: params[:related_name], view_type: :grid %>
```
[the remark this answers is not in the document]
adrianthedev: Actually, let's create a new issue to track the `count: 0` problem, and just fix what's under the scope of this issue (the missing locale tags).

adrianthedev (commented):
I'd say we fix the missing locales for this PR and tackle the "portugese multiple items uses the zero variant" in a different PR.

How about it?

app/components/avo/index/resource_grid_component.html.erb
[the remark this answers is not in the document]
Paul-Bob: I think this is the native translation package fix, we only can use the pluralize method on English.

> Using pluralized `one` instead `zero` on `no_item_found` and on `no_related_item_found` translation

 ```ruby
error_log:
        zero: "log de erro zero"
        one: "log de erro one"
        other: "logs de erro other"
```
image

As we can see from the original issue, we are using the name of the resource and pluralize it in order to get the empty state. Each language has it's own pluralization methods that's why i think using the zero count should work for every language in this case, we want to tell to user "0 related resource_name found". If `zero` count for that resource is not defined `one` will be used as default so it will never brakes.

https://guides.rubyonrails.org/i18n.html#pluralization

BTW i think this happens in every language with different words, not only on portughese
Example: In `avo.ro.yml` we have 
    `no_related_item_found: Nu s-au găsit %{item} asociate`
For `resource_name = "fruct"` we get the message `Nu s-au găsit fructe asociate`
For `resource_name = "jucator"` we get the message `Nu s-au găsit jucători asociate` and here we are stuck, can't translate it, with the cout implementation, we can change to:
 `no_related_item_found: Niciun %{item} asociat găsit`

```en:
  avo:
    resource_translations:
      fruct:
        zero: fruct
        one: fruct
        other: fructe
      comment:
        zero: jucător
        one: jucător
        other: jucători
```
Now we get always the correct message.
 Or, if the developer realy want can use the `i18n context` to give different translations for masculine or feminine words. My point is, we should not force this message to use always plural, sometimes it don't fit

Anyway, if you think this implementation is risky let me know and we can remove this part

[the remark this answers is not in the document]
adrianthedev: I understand now better what the issue is. Thanks for explaining.

I think the best solution would be to avoid using the phrase "no users found" and use something generic like "no items found" or "no related items found". This way we don't need to translate and figure out the pluralization in the same translation string.

What do you think?

Can we do that in this PR? So no `count: 0` argument?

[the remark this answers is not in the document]
Paul-Bob: I agree, that way we reduce the unnecessary complexity here! Implementing it on next commit

lib/generators/avo/templates/locales/avo.en.yml
```diff
@@ -62,9 +63,9 @@ en:
     new: new
     next_page: Next page
     no_cards_present: No cards present
-    no_item_found: No %{item} found
+    no_item_found: Record not found
```
adrianthedev, line 66:
```suggestion
    no_item_found: No record found
```

adrianthedev (approved):
I left two comments about English translations.

PR #1446, opened 2022-11-24
fix: resource getting wrong model

### LINKED ISSUE #1444: Missing resources inside action visible block
### missing resources inside action visible block

```ruby
   class SetToPrimaryAction < ::ApplicationBaseAction
    self.name = "SetToPrimary"

    self.visible = lambda do
      begin
        return false unless view == :show || view == :index

        # we expect the parent_resource and resource to be different 
      end
    end
```
the resource we receive in the visible block is always the parent resource. Regardless if the action is involved in
a regular show view or in an index view for the  has_many association. 
We are expecting to get the resource of the has_many association . 
can u please confirm the problem/bug  ?

### Steps to Reproduce
1. checkout the demo code https://github.com/elassadi/avo-demo
2. create a post and a comment 
3. hit http://127.0.0.1:3000/avo/resources/posts/1
4. check the debug log 

parent_resource and resource should be targeting the association resource  and the parent resource respectively
### System configuration
**Avo version**: latest

**Rails version**:

**Ruby version**:

**License type**:

 - [x] Community
 - [ ] Pro

**Are you using Avo monkey patches or overriding views or view components?**

 - [x] No

#### Impact

 - [] High impact
 - [x] Medium impact
 - [ ] Low impact

#### Urgency

 - [ ] High urgency
 - [x] Medium urgency
 - [ ] low urgency

Paul-Bob:
# Description

Fixes #1444

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/app/controllers/avo/application_controller.rb
+++ b/app/controllers/avo/application_controller.rb
@@ -176,7 +176,7 @@ module Avo
     end
 
     def hydrate_related_resource
-      model = @related_resource.model_class.descendants.include?(@model.class) ? @model : nil
+      model = @related_resource.model_class == @model.class ? @model : nil
       @related_resource.hydrate(view: action_name.to_sym, user: _current_user, model: model)
     end
 
--- a/app/controllers/avo/application_controller.rb
+++ b/app/controllers/avo/application_controller.rb
@@ -176,7 +176,8 @@ module Avo
     end
 
     def hydrate_related_resource
-      @related_resource.hydrate(view: action_name.to_sym, user: _current_user, model: @model)
+      model = @related_resource.model_class.descendants.include?(@model.class) ? @model : nil
+      @related_resource.hydrate(view: action_name.to_sym, user: _current_user, model: model)
     end
 
     def authorize_base_action
--- a/app/controllers/avo/application_controller.rb
+++ b/app/controllers/avo/application_controller.rb
@@ -172,12 +172,11 @@ module Avo
     end
 
     def hydrate_resource
-      @resource.hydrate(view: action_name.to_sym, user: _current_user)
+      @resource.hydrate(view: action_name.to_sym, user: _current_user, model: @model)
     end
 
     def hydrate_related_resource
-      model = @related_resource.model_class == @model.class ? @model : nil
-      @related_resource.hydrate(view: action_name.to_sym, user: _current_user, model: model)
+      @related_resource.hydrate(view: action_name.to_sym, user: _current_user, model: @related_model)
     end
 
     def authorize_base_action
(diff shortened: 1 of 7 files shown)


lib/avo/base_resource.rb
```diff
@@ -151,7 +151,7 @@ def hydrate(model: nil, view: nil, user: nil, params: nil)
       @user = user if user.present?
       @params = params if params.present?
 
-      if model.present?
+      if model.present? && model.class == model_class
```
adrianthedev, line 154: I'm not sure this is the proper place to fix this. I mean, `BaseResource.model_class` falls back to computing the model class based on it's own name.
It seems a bit arbitrary.

How about finding the place where we wrongly hydrate that resource and applying the fix there?

Paul-Bob: Hmm, the issue starts here:

image

I'm not sure how we can verify it without model_class... Yes, mostly it's computed but sometimes it's not, and i think this covers both situations. It's there any situation where we want a resource to have a model different from his model_class, ancestors or dependents?

adrianthedev: What I mean is, fix the issue where we run `resource.hydrate(model: WRONG_MODEL)` instead of checking inside the resource. Because it's clear to me that somewhere we're making that mistake.

Paul-Bob: Got it, i agree, issue found: 

image

app/controllers/avo/application_controller.rb
```diff
@@ -176,7 +176,8 @@ def hydrate_resource
     end
 
     def hydrate_related_resource
-      @related_resource.hydrate(view: action_name.to_sym, user: _current_user, model: @model)
+      model = @related_resource.model_class == @model.class ? @model : nil
```
adrianthedev: I think this check will fail if the referenced association is of the same model like `Comment` -> `has_many` -> `Comments`. The wrong model will be associated.

We should check **why** is a model hydrated there, and which one.

adrianthedev (commented):
I pushed some changes.

I think the error was that we were hydrating the `parent_resource` with the `model` instead of the `parent_model`. We also missed to hydrate the `resource` with `model`, which I added here.

WDYT about this approach?

adrianthedev:
The gist is that you won't have a `resource.model` (a user in our example) available in the action `visible` block, in an index view of users because there are many, not one. Instead, in a `has_one` (Team has one admin), you'll have it.

adrianthedev:
I released `2.21.1.pre.issue1444` to test out.

Paul-Bob:
> I pushed some changes.
> 
> I think the error was that we were hydrating the `parent_resource` with the `model` instead of the `parent_model`. We also missed to hydrate the `resource` with `model`, which I added here.
> 
> WDYT about this approach?

Much better, straight to the point!

PR #1451, opened 2022-11-25
Warn developers if they forget to set attachment_key on trix fields

iainbeeston:
# Description

To use attachments on a trix field, you need to set an attachment_key in the field definition (which gives the name of the association which attachments can be stored in).

Although there is code in the trix stimuls controller to warn the user if they haven't set this, it doesn't work, because it expects the `data-attachment-key` attribute to be set but empty and does not work if the attribute isn't set at all (which is how the html actually works if attachment_key isn't set).

Also, although the server throws an error if you try to attach something without setting the attachment name (it throws an error because it cannot find the right association to use), the error is unhelpful:

    TypeError ([...] is not a symbol nor a string)

(Where the array contains the name of every association on the model.)

I've updated this so that:

1. The javascript error is shown if the HTML attribute is empty OR undefined
2. It raises a more descriptive error on the server if it cannot find the correct association

Fixes # (issue)

# Checklist:

## Manual review steps

1. Add a trix field without setting `attachment_key:`, try to attach a file and confirm that an error is raised in the browser console
2. Add a trix field with an `attachment_key:` but make it for an association that does not exist, then when you try to attach a file it should show a helpful error in the rails logs
3. Add a trix field with a valid `attachment_key:` and it should work as expected.

### DIFF
--- a/app/controllers/avo/attachments_controller.rb
+++ b/app/controllers/avo/attachments_controller.rb
@@ -10,6 +10,10 @@ module Avo
       blob = ActiveStorage::Blob.create_and_upload! io: params[:file], filename: params[:filename]
       association_name = BaseResource.valid_attachment_name(@model, params[:attachment_key])
 
+      if association_name.blank?
+        raise ActionController::BadRequest.new("Could not find the attachment association for #{params[:attachment_key]} (check the `attachment_key` for this Trix field)")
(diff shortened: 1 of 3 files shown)


adrianthedev:
Thank you!

PR #1452, opened 2022-11-25
Don't set the csrf token if it's not available

iainbeeston:
# Description

It's possible to disable csrf tokens by setting
`config.action_controller.allow_forgery_protection = false` in your environment config file. This is the default in the test environment.

Unfortunately, the trix field controller assumes that the csrf token will always be available, which makes it crash if it's not. Because of that it's not possible to use avo trix fields in tests.

I've changed the code so that if the csrf token isn't set then we don't use it.

Fixes # (issue)

# Checklist:

## Manual review steps

1. Disable csrf tokens in your development environment by setting `config.action_controller.allow_forgery_protection = false` in `config/environments/development.rb`
2. Try to edit and save a resource with a trix field (this should work on this branch but hit errors on main)

### DIFF
--- a/app/javascript/js/controllers/fields/trix_field_controller.js
+++ b/app/javascript/js/controllers/fields/trix_field_controller.js
@@ -93,7 +93,10 @@ export default class extends Controller {
 
     xhr.open('POST', this.uploadUrl, true)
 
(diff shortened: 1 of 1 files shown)


adrianthedev:
Amazing! Thank you!

PR #1454, opened 2022-11-25
fix: select all when query can't be converted to_sql

### LINKED ISSUE #1255: Rest API Resource select all feature breaks
### Describe the bug

We have implemented two rest API resources following this recipe: https://docs.avohq.io/2.0/recipes/rest-api-integration.html and after updating to Avo 2.14 it started throwing an exception (watch screenshot). The problem is, in our rest API based resources we don't have a DB query because our index rows are fetched from another application via API and they are not DB persistent models.

### Screenshots

image

#### Impact

 - [x] High impact
 - [ ] Medium impact
 - [ ] Low impact

#### Urgency

 - [x] High urgency
 - [ ] Medium urgency
 - [ ] low urgency

Paul-Bob:
# Description

Fixes #1255

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/app/components/avo/index/resource_table_component.rb
+++ b/app/components/avo/index/resource_table_component.rb
@@ -15,23 +15,11 @@ class Avo::Index::ResourceTableComponent < ViewComponent::Base
   end
 
   def encrypted_query
-    return if query.nil?
+    return :select_all_disabled if query.nil? || !query.respond_to?(:all) || !query.all.respond_to?(:to_sql)
 
     Avo::Services::EncryptionService.encrypt(
-      message: message_to_encrypt,
+      message: query.all.to_sql,
       purpose: :select_all
     )
-  end
-
-  def message_to_encrypt
-    if query.respond_to? :all
-      if query.all.respond_to? :to_sql
-        query.all.to_sql
-      else
-        query.all
-      end
-    else
-      query
-    end
   end
 end
--- a/app/components/avo/index/resource_table_component.rb
+++ b/app/components/avo/index/resource_table_component.rb
@@ -2,7 +2,7 @@
 
 class Avo::Index::ResourceTableComponent < ViewComponent::Base
   include Avo::ApplicationHelper
-  attr_reader :pagy
+  attr_reader :pagy, :query
 
   def initialize(resources: nil, resource: nil, reflection: nil, parent_model: nil, parent_resource: nil, pagy: nil, query: nil)
     @resources = resources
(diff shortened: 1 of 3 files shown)


app/components/avo/index/resource_table_component.rb
```diff
@@ -15,11 +15,23 @@ def initialize(resources: nil, resource: nil, reflection: nil, parent_model: nil
...
+    return if query.nil?
 
     Avo::Services::EncryptionService.encrypt(
-      message: @query.all.to_sql,
+      message: message_to_encrypt,
       purpose: :select_all
     )
   end
+
+  def message_to_encrypt
```
adrianthedev, line 26: I think we're being too clever here, and will bite us in the behind in the future. Why fallback to `.all`? Isn't that going to output all the items in a custom adapter?

I was thinking of having a `serialized` method on the adapter. The `ActiveRecord` one would use `to_sql` and others will implement their own.

How about we just check if it responds to `to_sql`. It it doesn't, then output nothing, and let's disable the feature in the UI.
How does that sound?

adrianthedev (changes requested):
Left a comment for changes

app/components/avo/index/resource_table_component.rb
Paul-Bob: We can do it. First approach was to struggle and send it anyway, that way users with custom HTTP resources could use the select all feature. But i agree it can be quite risky, if the object to encrypt is to large it can break.

app/javascript/js/controllers/item_select_all_controller.js
```diff
@@ -63,4 +61,9 @@ export default class extends Controller {
       this.selectAllOverlayTarget.classList.add('hidden')
     }
   }
+
+  // True if there are more pages available and if query encryption run successfully
+  selectAllEnable() {
```
adrianthedev, line 66: I think this should be `selectAllEnabled` (state) not `selectAllEnable` (action).
```suggestion
  selectAllEnabled() {
```

app/javascript/js/controllers/item_select_all_controller.js
```diff
@@ -35,8 +34,7 @@ export default class extends Controller {
     this.itemCheckboxTargets.forEach((checkbox) => allSelected = allSelected && checkbox.checked)
     this.checkboxTarget.checked = allSelected
 
-    // Only run "all matching" if there are more pages available
-    if (this.pageCountValue > 1) {
+    if (this.selectAllEnable()) {
```
adrianthedev, line 37:
```suggestion
    if (this.selectAllEnabled()) {
```

adrianthedev (approved):
👌

PR #1471, opened 2022-12-01
feature: add disabled option

adrianthedev:
# Description

Fixes https://github.com/avo-hq/avo/discussions/1458

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/app/components/avo/fields/has_one_field/show_component.html.erb
+++ b/app/components/avo/fields/has_one_field/show_component.html.erb
@@ -5,7 +5,7 @@
 <% else %>
   <%= render Avo::PanelComponent.new(name: @field.name) do |c| %>
     <% c.tools do %>
-      <% if !@field.is_readonly? && can_attach? %>
+      <% if !@field.is_readonly? && !@field.is_disabled? && can_attach? %>
         <%= a_link attach_path,
           icon: 'heroicons/outline/link',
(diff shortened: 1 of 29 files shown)


app/components/avo/fields/has_one_field/show_component.html.erb
```diff
@@ -5,15 +5,15 @@
...
+      <% if !@field.is_readonly? && !@field.is_disabled? && can_attach? %>
         <%= a_link attach_path,
           icon: 'heroicons/outline/link',
           color: :primary,
           'data-turbo-frame': 'attach_modal' do %>
           <%= t('avo.attach_item', item: @field.name.downcase) %>
         <% end %>
       <% end %>
-      <% if !@field.is_readonly? && can_see_the_create_button? %>
+      <% if !@field.is_readonly? && !@field.is_disabled? && can_see_the_create_button? %>
```
Paul-Bob, line 16: Suggestion: Add method `not_read_only` to be called on line 8, 16

adrianthedev: Not a bad suggestion, but I think about what is the opposite of disabled? enabled. But `enabled?` could mean so many things.

I would leave it like it is for now just so it will be inline with the rest of the properties.

PR #1480, opened 2022-12-04
Fix/activestorage dependency details

MrJoy:
# Description

Minor fixups for noise when ActiveStorage isn't present, and for workstation setup for contributors.

# Checklist:

## Manual review steps

Executed fragment of code in IRB.

Unfortunately, I'm encountering major problems getting the test suite running locally.

For starters, seemingly every test fails because of a license key check where the license_key value is nil.  If I change the stub, then I get other problems:

* Missing URL helpers in a bunch of tests (`undefined method `sign_out_path_name' for #<Module:0x000000010eff3598>`)
* Failures in the code that's meant to check the licensing functionality and is rendered incorrect by having changed the stub
* Capybara failures because the user wasn't logged in and so the login page was rendered (e.g. `Unable to find css "turbo-frame[id='dashy_users_metric'][data-card-index='3']"` -- screenshot shows a completely unstyled login page), etc.

Basically, just all kinds of weirdness.

### DIFF
--- a/bin/init
+++ b/bin/init
@@ -27,6 +27,7 @@ app_root do
   header 'Installing Yarn packages'
   run! 'yarn'
   run! '(cd spec/dummy; yarn)'
+  run! '(cp spec/dummy/.env.test.sample spec/dummy/.env.test)'
 
   if use_docker == 'y'
     header 'Creating the Docker volume'
--- a/bin/init
+++ b/bin/init
@@ -26,6 +26,7 @@ app_root do
 
   header 'Installing Yarn packages'
   run! 'yarn'
+  run! '(cd spec/dummy; yarn)'
 
   if use_docker == 'y'
     header 'Creating the Docker volume'
--- a/lib/avo/app.rb
+++ b/lib/avo/app.rb
@@ -59,10 +59,12 @@ module Avo
 
         # Set the current host for ActiveStorage
         begin
-          if Rails::VERSION::MAJOR === 6
-            ActiveStorage::Current.host = request.base_url
-          elsif Rails::VERSION::MAJOR === 7
-            ActiveStorage::Current.url_options = request.base_url
+          if defined?(ActiveStorage::Current)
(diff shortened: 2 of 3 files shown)


adrianthedev:
Thanks for writing and for the improvements @MrJoy.

Regarding the test runs... I pushed some changes to this PR.
Can you pleasepull these latest changes and copy over the `.env.test` file (`cp spec/dummy/.env.test.sample spec/dummy/.env.test`) to see if you can run the tests?

MrJoy:
Much improved!  I initially had some failures, but that was due to my hack for the license request stub in place.  Removing that, and rerunning the failed tests, they all passed.  I'm doing a full run again just to be sure, but so far it looks like things work.

MrJoy:
Ah.  Not quite.  A couple errors turned up again on the full run:

image

Also, the `Unable to find css "turbo-frame[id='dashy_users_metric'][data-card-index='0']"` turned up again.  Guessing there's some intermittency going on.

MrJoy:
Ok, no, they're failing consistently.  I must've not re-run all the failed tests.

```
rspec ./spec/system/avo/dashboards_spec.rb:54 # Dashboards dashboard with cards metric card
rspec ./spec/system/avo/dashboards_spec.rb:127 # Dashboards dashboard with cards chartkick area card
```

The exact specs that are failing.

adrianthedev:
Yes, it's a race condition that happens when the cards load up in the dashboard. There are about 14 requests being made in the same time (for each card) and that's where it the race condition manifests itself. Not sure why it only happens on local (the CI passes everytime). Probably the hardware executes at a different speed.

We worked on and fixed this issue in Avo 3.

I think I'd merge this PR as is so we don't pollute it with a bunch of other unrelated changes and come back with another PR to fix that in Avo 2.

WDYT?

MrJoy:
I'm completely fine with merging the PR if you are.

PR #1494, opened 2022-12-21
Remove select from count in search

iainbeeston:
# Description

The latest versions of avo introduced a regression for my app, where search would raise the following error:

    ActiveRecord::StatementInvalid (PG::UndefinedFunction: ERROR:  function count(people, text) does not exist
    LINE 1: SELECT COUNT(DISTINCT people.*, LOWER(UNACCENT(people...
    
    HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
    ):

    avo (2.21.0) app/controllers/avo/search_controller.rb:54:in `search_resource'
    avo (2.21.0) app/controllers/avo/search_controller.rb:34:in `block in search_resources'
    avo (2.21.0) app/controllers/avo/search_controller.rb:28:in `map'
    avo (2.21.0) app/controllers/avo/search_controller.rb:28:in `search_resources'
    avo (2.21.0) app/controllers/avo/search_controller.rb:21:in `show'

This happens because I'm doing two things:

1. In my resource I have defined `resolve_query_scope` to call `model_class.alphabetical`, where the `alphabetical` scope sorts alphabetically regardless of accents or case (`scope :alphabetical, -> { order(Arel.sql("LOWER(UNACCENT(people.name))")) }`). I want this so that the models are always shown alphabetically.
2. To make this work with ransack, in previous versions of avo I had to add `"LOWER(UNACCENT(people.name))"` to the `search_query` in my resource, like this: `scope.ransack(name_cont: params[:q]).select("people.*, LOWER(UNACCENT(people.name))")` (otherwise ransack would throw an error that I'm sorting by a column that does not exist in the select statement).

This has worked in many versions of avo, but since 0ea3acd4c8c1fb941a9c986c0aedf61b6bc5775f this now throws the error above, because the search controller is trying to run `SELECT COUNT(DISTINCT people.*, LOWER(UNACCENT(people.name)))` which doesn't make sense (AFAIK you can't count two different things like that).

To resolve the issue I've unscoped select when performing the count, so that this works as it used to.

Fixes # (issue)

# Checklist:

## Manual review steps

1. Search for any resource and confirm that the count (the number of matching models) is correct

### DIFF
--- a/app/controllers/avo/search_controller.rb
+++ b/app/controllers/avo/search_controller.rb
@@ -51,7 +51,7 @@ module Avo
       ).handle
 
       # Get the count
-      results_count = query.count
+      results_count = query.unscope(:select).count
 
       # Get the results
       query = query.limit(8)


adrianthedev:
This proves how a small change can cause ripples. Thank you!

iainbeeston:
Unfortunately this doesn't actually work 😭 Sorry I should have tested it more thoroughly. With this I get a different error:

```
*** ActiveRecord::StatementInvalid Exception: PG::UndefinedFunction: ERROR:  function count(people, text) does not exist
LINE 1: SELECT COUNT(DISTINCT people.*, LOWER(UNACCENT(people...
               ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

The whole query is:

```
SELECT COUNT(DISTINCT people.*, LOWER(UNACCENT(people.name))) FROM "people" WHERE LOWER(UNACCENT("name")) ILIKE '%iain%'
```

What I think should really happen is this:

```
SELECT COUNT(DISTINCT people.id) FROM "people" WHERE LOWER(UNACCENT("name")) ILIKE '%iain%'
```

Which is what happens for other resources. I'll raise a new PR to make it work that way. Sorry for the confusion!

PR #1496, opened 2022-12-24
Resource generator incorrect date_time field fix

aldent95:
# Description

This fixes a bug with the Resource generator where it uses the wrong field type for datetime database fields. It should map to the Avo date_time field, whereas it currently maps incorrectly to datetime.

Also added password_digest to the ignored db fields as this is another common password field people use in Rails apps.

# Checklist:

- [x] I have made corresponding changes to the [documentation](https://github.com/avo-hq/docs) (N/A)

## Manual review steps

1. Create a minimal rails app with a model that has a datetime field, I.E. a User model with a confirmed_at datetime field
1. Run the Avo resource generator for that model (I.E. ```bin/rails generator avo:resource user```) and observe the datetime field getting set to ``` as: :date_time ``` whereas running the same steps from the main branch will result in ```as: :datetime```

### DIFF
--- a/spec/dummy/db/migrate/20221224005546_create_events.rb
+++ b/spec/dummy/db/migrate/20221224005546_create_events.rb
@@ -1,0 +1,10 @@
+class CreateEvents < ActiveRecord::Migration[6.1]
+  def change
+    create_table :events do |t|
+      t.string :name
+      t.datetime :event_time
+
+      t.timestamps
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


spec/dummy/db/migrate/20221224005546_create_events.rb
```diff
@@ -0,0 +1,10 @@
+class CreateEvents < ActiveRecord::Migration[6.1]
```
adrianthedev, line 1:
```suggestion
class CreateEvents < ActiveRecord::Migration[6.0]
```

adrianthedev: The Dummy app is a Rails 6.1 app, but we run tests (and migrations) on a Rails 6.0 app too. That's why we need to declare this migration as 6.0.

PR #1514, opened 2023-01-04
Make search work when using strange selects in ransack

iainbeeston:
# Description

Another attempt at fixing fa513705f2b79ca8f498d05e755f55759f42c824. It turns out that simply unscoping select does not fix the issue as rails does something weird with the count query, as explained in #1494 

Fixes # (issue)

 #1494 

# Checklist:

## Manual review steps

1. Search for any resource and confirm that the count (the number of matching models) is correct
2. 

### DIFF
--- a/app/controllers/avo/search_controller.rb
+++ b/app/controllers/avo/search_controller.rb
@@ -51,7 +51,7 @@ module Avo
       ).handle
 
       # Get the count
-      results_count = query.unscope(:select).count
+      results_count = query.reselect(:id).count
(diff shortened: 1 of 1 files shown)


iainbeeston:
I've double checked this and it:

1. Works for my use case (using `select` in the ransack search query)
2. Produces the same SQL query as the code in avo 2.21 (before #1494)

adrianthedev:
All good! Let's have another go at this.

Issue #816, opened 2022-04-10
Includes are not utilized with actions (other than `index`)


rickychilcott:
### Describe the bug
My expectation is that within a Avo::Rescoure setting `self.includes` would be used in all locations where a resource is being queried. However, it's only utilized for the `index` action.

This isn't a problem in most cases, but if you utilize [strict loading](https://guides.rubyonrails.org/configuring.html#config-active-record-action-on-strict-loading-violation), especially when choosing to raise an exception, then Avo is unuseable.

I see that there is `resolve_find_scope` and `resolve_query_scope` which could be set, per resource, but I'm thinking that this should be the default

I'm happy to provide some tests to build this functionality, but I'm not sure the ideal interaction between the two methods above with what I'm asking for.

### System configuration
**Avo version**: 2.1.1

**Rails version**: 6.1.x and 7.x

**Ruby version**: 3.0.3

#### Impact

 - [ ] High impact
 - [x] Medium impact
 - [ ] Low impact

#### Urgency

 - [ ] High urgency
 - [x] Medium urgency
 - [ ] Low urgency

adrianthedev:
Hey @rickychilcott,

I don't know a lot of details about strict loading. 
Can you provide a bit more context around this? 
Where is this an issue? Can you share some code?

Thanks!

rickychilcott:
Here is how strict loading works:

```ruby
class User
  self.strict_loading_by_default = true
  
  has_many :memberships
  has_many :organizations, through: :memberships
end

User.find(1).memberships # will raise ActiveRecord::StrictLoadingViolationError
User.includes(:memberships).find(1).memberships  # won't raise

User.includes(memberships).find(1).organizations  # will raise ActiveRecord::StrictLoadingViolationError
User.includes(memberships: :organizations).find(1).organizations  # won't raise
```

When viewing a resource index, everything works fine when we have the following config:
```ruby
class UserResource < Avo::BaseResource
  self.includes = [{memberships: :organizations}]
end
```

But when you hit the `edit` or `show` actions it will raise `ActiveRecord::StrictLoadingViolationError` because the `includes` are not being used in the finders.

adrianthedev:
Ok. Got it. I see the issue now. 

Let me see what I can do to fix this.

adrianthedev:
@rickychilcott can you remove `strict_loading_by_default` from your app? 

We're a little backed up with a few other chores and need to make headway with the other features.

rickychilcott:
[...]

PR #1541, opened 2023-01-22
fix: always eager load all associations

djfpaagman:
# Description
Since Rails 6.1 you can set `strict_loading` to raise errors when you access associations that are not explicitedly eager loaded.  [It's explained a bit further in this blog post](https://www.bigbinary.com/blog/rails-6-1-adds-strict_loading-to-warn-lazy-loading-associations).

For a new app I'm working on I want to set this globally via the Rails config. I'm running into issues with Avo though, where for the non-index actions the assocations are not eager loaded yet.

Since there is already a `self.includes` setting for resources, it was quite easy to also load them for the other actions.

Also found an earlier reported issue (but closed) for this: #816 

I wasn't sure how to test this, there are a few options:
1. Configure it on one model that sets `self.strict_loading_by_default = true` and have regular tests for all actions. 
2. Configure it on one model that sets `self.strict_loading_by_default = true` and have explicit tests for the eager loading part.
3. Configure it globally (this is what I did in this PR) and not change any tests.

I'm also not sure if there is a downside to always loading the assocations, I don't think so though. The only situation I can think of is having fields specified in `self.includes` that are then not used in the form (no fields that use it). Right now that would not load the resource, with this change it would. What are your thought on that?

# Checklist:

## Manual review steps
1. Set `config.active_record.strict_loading_by_default = true` or `self.strict_loading_by_default = true` on a specific model.
5. Load a Avo resource in the Avo panel that includes a field for an association
6. See that is raises an error without these changes

### DIFF
--- a/app/controllers/avo/application_controller.rb
+++ b/app/controllers/avo/application_controller.rb
@@ -136,7 +136,7 @@ module Avo
     end
 
     def model_find_scope
-      eager_load_files(@resource, model_scope)
+      eager_load_associations(@resource, model_scope)
     end
 
     def model_scope
@@ -149,7 +149,7 @@ module Avo
       @related_model = if @field.is_a? Avo::Fields::HasOneField
         @model.send association_name
       else
-        eager_load_files(@related_resource, @model.send(association_name)).find params[:related_id]
+        eager_load_associations(@related_resource, @model.send(association_name)).find params[:related_id]
       end
     end
 
@@ -235,7 +235,7 @@ module Avo
       App.get_resource_by_model_name reflected_model
     end
 
-    def eager_load_files(resource, query)
+    def eager_load_associations(resource, query)
       if resource.attached_file_fields.present?
         resource.attached_file_fields.map do |field|
           attachment = case field.class.to_s
@@ -249,6 +249,10 @@ module Avo
 
           return query.includes "#{field.id}_#{attachment}": :blob
         end
+      end
+
+      if resource.includes.present?
+        query = query.includes(*resource.includes)
       end
 
       query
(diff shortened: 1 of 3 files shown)


adrianthedev:
Thanks for getting this started @djfpaagman.
I pushed some changes to conditionally enable strict loading for Rails 6.1 (we test with 6.0) too.
Also, there was an early return in the `eager_load_associations` method.

The tests are running now.

adrianthedev:
> 1. What's the reason for not testing Ruby 3.1 and 3.2 and Rails 7.0 yet?

The reason is that I couldn't get Appraisal (which generates the gemfiles) to work with some gems. https://github.com/ddnexus/pagy in particular at that time. Switching from one version to another required some API changes. And because they are on different versions of ruby, I couldn't support that. I think...
I'd love to add Rails 7.X and higher ruby versions to the matrix, though. Do you have experience in that field?

> 2. The Rails version check will fail for Rails 7, since that is on minor version 0. Do you prefer something like

That's correct, but I think Rails 7 supports `config.active_record.strict_loading_by_default = true` right? so we should be good? So, enable that feature on Rails 6.1 and higher (including 7.0, and 7.1).
Am I wrong on that?

Yeah, some tests still fail with strict loading. So they sill need fixing.

CleanShot 2023-01-23 at 13 31 28

dhnaranjo:
Short term workaround:

```
# app/overrides/avo/application_controller_override.rb

module Avo
  class ApplicationController
    before_action -> { ActiveRecord.action_on_strict_loading_violation = :log }
  end
end
```

This requires that you're set up for [reopening classes Zeitwerk-style](https://guides.rubyonrails.org/engines.html#overriding-models-and-controllers).

PR #1553, opened 2023-01-26
feature: add optional collapsible option

### LINKED ISSUE #1551: Alias `collapsable` as `collapsible`
### Feature

```ruby
section "Resources", icon: "resources", collapsable: true do
  resource :course
end
```
Add an alias for `collapsable` as `collapsible` on menu editor.

Reason: 

> "Collapsable" is an adjective for something that can collapse, and "collapsible" is an adjective for something capable of collapsing. Both words have the same definitions. Although the most common writing is "collapsible."

Source: https://thecontentauthority.com/blog/collapsable-vs-collapsible

Paul-Bob:
# Description

Fixes #1551

# Checklist:

### DIFF
--- a/app/components/avo/sidebar/base_item_component.rb
+++ b/app/components/avo/sidebar/base_item_component.rb
@@ -22,7 +22,7 @@ class Avo::Sidebar::BaseItemComponent < ViewComponent::Base
   end
 
   def collapsable
-    item.collapsable
+    item.collapsible || item.collapsable
   end
 
   def collapsed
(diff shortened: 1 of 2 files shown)


Paul-Bob:
@adrianthedev is there a way to use an alias for dry-initializer options? I did not found any that's why i came up with this workaround.

adrianthedev:
How about we leave it like this for Avo 2 and rename the option to `collapsible` in Avo 3 (with upgrade guide). This way we don't confuse people with two names for the same options here and we fix it in the next version.
WDYT?

Paul-Bob:
Not sure if renaming it add any value, both words ("collapsible" and "collapsable") apply on this context. They end up meaning the same thing. Or do you think "collapsible" apply better on the menu context? Anyway, fell free to close this one, we can discuss and implement it if needed on Avo 3

adrianthedev:
Let's leave it like it is now. We'll see next time when this pops up

PR #1575, opened 2023-02-11
Enable nonce for inline styles.

simi:
- similar to https://github.com/avo-hq/avo/pull/201 but for inline style
- there is no helper support in Rails directly, following https://dev.to/mdoyle13/rails-7-csp-how-to-add-a-nonce-to-an-inline-tag-if-you-must-59fm


adrianthedev:
Thank you for your contribution @simi!

PR #1589, opened 2023-02-19
By default have text fields convert ASCII-8BIT to UTF-8

lorint:
# Description
When columns with binary data come in as a text field (such as Geospatial data in MySQL) then it is an ASCII string that does not render properly in Avo.  By using a `format_using` on the field the encoding can be safely changed to UTF-8, and out of convenience this can be established as a default behaviour for rendering text fields.

# Checklist:

## Manual review steps
1. Obtain (or create) a database in MySQL that has text fields encoded as ASCII-8BIT.  The original **[sakila](https://downloads.mysql.com/docs/sakila-db.zip)** test database from MySQL AB does use this kind of data type in the **Address** table for geospatial data.
2. Create a Resource file for the affected model, can use this for **sakila**:
```
bin/rails g avo:resource address
```
3. When rendering the page, note that this error appears:
```
incompatible character encodings: UTF-8 and ASCII-8BIT
```
4. Implement this `format_using` on the `location` field:
```
->(value) { value.encode!("UTF-8", invalid: :replace, undef: :replace, replace: "?") }
```
5. Note that the page now renders, and invalid characters in the `location` column have been replaced with **?**:
Screenshot 2023-02-19 at 11 46 54

### DIFF
--- a/app/views/layouts/avo/application.html.erb
+++ b/app/views/layouts/avo/application.html.erb
@@ -35,7 +35,7 @@
             <div class="content p-4 lg:p-6 flex-1 flex flex-col justify-between items-stretch <%= @container_classes %>">
               <%= render partial: "avo/partials/custom_tools_alert" %>
               <div class="flex flex-1 flex-col justify-between items-stretch space-y-8">
-                <%= yield %>
+                <%= yield.force_encoding('UTF-8') %>
                 <%= render partial: "avo/partials/footer" %>
               </div>
             </div>
--- a/lib/generators/avo/resource_generator.rb
+++ b/lib/generators/avo/resource_generator.rb
@@ -214,7 +214,7 @@ module Generators
       end
 
       def field(name, type)
-        names_mapping[name.to_sym] || fields_mapping[type.to_sym] || {field: "text"}
+        names_mapping[name.to_sym] || fields_mapping[type&.to_sym] || {field: "text"}
       end
 
       def associations_mapping
--- a/spec/dummy/db/seeds.rb
+++ b/spec/dummy/db/seeds.rb
@@ -15,7 +15,6 @@ Review.delete_all
 Fish.delete_all
 Course.delete_all
 Course::Link.delete_all
-Fish.delete_all
 TeamMembership.delete_all
 Team.delete_all
 Comment.delete_all


adrianthedev:
Hey @lorint. Can this conversion be done with `format_using` or something similar?
I would love to support this use case, but forcing everything to be UTF-8 and doing arbitrary search & replace does not sound like a good idea.

Maybe that particular column can be converted using a getter on that model where you can apply all kinds of manipulations and then correctly displayed in Avo?

lorint:
> forcing everything to be UTF-8 ... does not sound like a good idea

I know it sounds a bit goofy... but hear me out :)

Have put together a simple project with 6 tables, and added several popular Rails admin panels.  This is to compare and demonstrate everything a little better:

https://user-images.githubusercontent.com/5301131/221380345-1db3e709-03ed-43f0-aaed-8c60e3e240e0.mp4

adrianthedev:
Ok. I see what you mean. It's an opportunity 🙂 I'll take it!

Let me investigate to see how I can make a better version for that computed check in the `TextField`.

One thing though. I've been having issues trying to import the sakila database on my machine. Something related to the delimiters. After 30 minutes of googling and trying out different things, I stopped.
Do you know how I can create an ASCII or non-utf-8 column in a postgres database so I can try it out?

lorint:
> ... how I can create an ASCII or non-utf-8 column in a postgres database so I can try it out?

Have sent to your email a setup script that can be used in this way:

```
createdb northwind
psql -d northwind < northwind.sql
```

Will put this into a public repo later this evening and add the link to that here so that more folks can also try this out.

Should end up with the `categories` table having a `picture` column which is `bytea` -- and then this comes through as ASCII-8BIT using the `pg` gem.  Fun times!

adrianthedev:
Hey @lorint.

I'm thinking of merging this PR without the `text_field` override.
That can be applied by each user.

lorint:
> without the `text_field` override

Sure, that works -- will take out that commit now.

adrianthedev:
I was going to suggest that I could do it, but thank you!

lorint:
Cool!  Thanks for the merge.

Here's a patch if anyone needs the generic text field stuff:
```
class ::Avo::Fields::TextField
  alias _original_initialize initialize
  def initialize(id, **args, &block)
    if instance_of?(::Avo::Fields::TextField) || instance_of?(::Avo::Fields::TextareaField)
      args[:format_using] ||= ->(value) do
        if value.is_a?(String) && value.encoding != Encoding::UTF_8
          value = value.encode!("UTF-8", invalid: :replace, undef: :replace, replace: "?")
        end
        value
      end
    end
    _original_initialize(id, **args, &block)
  end
end
```

adrianthedev:
I didn't know about this `alias` technique. Thanks!

PR #1625, opened 2023-03-12
feature: field level file authorization

### LINKED ISSUE #1624: Field level file authorization
### Feature

Ability to police each file upload/download/delete

Let's say we have 2 attachments on a model

```ruby
class Model < ApplicationRecord
  has_many_attached :photos
  has_many_attached :documents
end
```

It would be great if we could define a police where every user can download the photos, only admins can download documents and only registered users with student role can upload document.

Paul-Bob:
# Description

Now we can police each file using field level authorization methods.

```ruby
# app/avo/resources/post_resource.rb
field :cover_photo, as: :file, is_image: true
```

```ruby
# app/policies/post_policy.rb
def upload_cover_photo?
  user.student?
end

def download_cover_photo?
  true
end

def delete_cover_photo?
  user.admin?
end
```

Fixes #1624

# Checklist:

Docs PR https://github.com/avo-hq/avodocs/pull/25

### DIFF
--- a/lib/avo/fields/concerns/file_authorization.rb
+++ b/lib/avo/fields/concerns/file_authorization.rb
@@ -8,15 +8,23 @@ module Avo
         delegate :authorize_action, to: :authorization
         delegate :id, :model, to: :@field
 
-        # Dynamically generate can_upload_file?, can_delete_file?, and can_download_file? methods
-        [:upload, :delete, :download].each do |action|
-          define_method "can_#{action}_file?" do
-            # Check if the current user can perform the corresponding actions
-            # upload_attachments?, delete_attachments?, and download_attachments?
-            if authorize_action("#{action}_attachments?".to_sym, raise_exception: false)
-              # Check if the current user can perform the action on the current file
-              authorize_action("#{action}_#{id}?", record: model, raise_exception: false)
-            end
+        def can_upload_file?
+          authorize_file_action(:upload)
+        end
+
+        def can_delete_file?
+          authorize_file_action(:delete)
+        end
+
+        def can_download_file?
+          authorize_file_action(:download)
+        end
+
+        private
+
+        def authorize_file_action(action)
+          if authorize_action("#{action}_attachments?", raise_exception: false)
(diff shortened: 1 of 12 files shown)


lib/avo/fields/concerns/file_authorization.rb
```diff
@@ -0,0 +1,25 @@
...
+    module Concerns
+      module FileAuthorization
+        extend ActiveSupport::Concern
+
+        delegate :authorization, to: :@resource
+        delegate :authorize_action, to: :authorization
+        delegate :id, :model, to: :@field
+
+        # Dynamically generate can_upload_file?, can_delete_file?, and can_download_file? methods
+        [:upload, :delete, :download].each do |action|
```
adrianthedev, line 12: I appreciate attempting to DRY the code a bit, but I think that having searchable code will help our contributors and us in the long run. For example, you can't find these methods if you do a global search for `can_download_file?` in the project.

I'm find with a little duplication if we get all that power instead. What do you think?

adrianthedev (approved):
The PR looks perfect!
I left a tiny suggestion to make the code easier to understand in the long run.

Thank you!

lib/avo/fields/concerns/file_authorization.rb
Paul-Bob: Sure, and I agree, this isn't searchable, i just love how much we can DRY the code with ruby and I let myself be carried away by the moment =')) I'll make it searchable and more readable on the next commit =)

Paul-Bob: Hmm I just thanked about this, whenever we use `define_method` we should add an comment with `# def method_name(s)` this way will be found on a global search. In this case:
```ruby
# def can_download_file?
# def can_upload_file?
# def can_delete_file?
```

Anyway i already committed and removed the dynamic `define_method` here

adrianthedev: Yeah, that could work. I don't see any downsides to that.

Issue #1513, opened 2023-01-04
Location field


rctneil:
Hi,

Could a field be added for a location.    Maybe it could display as a marker on a map and the ability to type the lat and lng in directly.  Then store in a variety of ways.  I typically store in separate lat and lng fields.

adrianthedev:
Yes, that would be nice to have. I would merge a PR if there was one open.

yondifon:
Saw this from Beginner Boutines. Can I send in a PR?

adrianthedev:
Hey @yondifon. Thanks for wanting to contribute.
The goal of the Beginner Boutines app is to help beginners get into Ruby development.

Do you consider yourself a junior developer?

yondifon:
In the Ruby space. Yes. 

but I have some experience in other communities.

adrianthedev:
Sounds good @yondifon! 
Yes. Go for it!

adrianthedev:
And thank you!

adrianthedev:
I'm writing an approach now

adrianthedev:
## Approach:

I'll use the Trix field as an example because it should be similarly built.
You'll need to produce a few files:

 - the field file ([example](https://github.com/avo-hq/avo/blob/main/lib/avo/fields/trix_field.rb)). That registers the field with Avo. It also holds some options that are specific to that field (`always_show`, `attachments_disabled`, etc.)
 - the view components ([show](https://github.com/avo-hq/avo/blob/main/app/components/avo/fields/trix_field/show_component.html.erb), [show 2](https://github.com/avo-hq/avo/blob/main/app/components/avo/fields/trix_field/show_component.rb), similar for edit)
 - a js controller ([example](https://github.com/avo-hq/avo/blob/main/app/javascript/js/controllers/fields/trix_field_controller.js)). In order to make the JS to work you need to add the js file and register the controller [here](https://github.com/avo-hq/avo/blob/main/app/javascript/js/controllers.js)

As a database column [I see](https://stackoverflow.com/questions/8150721/which-data-type-for-latitude-and-longitude) that some use the `point` data type, but most people have two columns `latitude` and `longitude`.
We could join them into one field using a getter and setter on the model.
```ruby
class City
  def coordinates
    [latitude, longitude]
  end
  
  def coordinates=(value)
    self.latitude = value.first
    self.longitude = value.last
  end
end
```

This is my approach, but feel do your thing.

Thanks!

yondifon:
[...]

PR #1631, opened 2023-03-16
Location field review

helpotters:
# Description
Location tag field implementation. Submitting upon @adrianthedev's request in [issue #1513](https://github.com/avo-hq/avo/issues/1513). 

Need help with: 

- [x] How to create two data inputs for one field (or an alternative strategy)
- [x] Review edit_component.html.erb

image


helpotters:
Thank you for helping me with the field inputs! I'll work on the map display in the show component. 

I'll also look into an autocomplete feature. That may require some sort of coupling between `coordinates` and an address, but I'll leave that alone for now and focus on the map display first. 

Google Maps seems to be the better option for ease-of-use, but Mapbox's customization may be more suited for avo's audience. For now, I'll implement Google Maps for simplicity (and autocomplete later).

adrianthedev:
Yes, I think that's a good approach. Let's focus on showing the map with the proper coordinates first.

We can tackle the address and auto-complete in a later iteration. 
Our goal here is to ship something useful and usable in it's "initial version" and later add features based on user feedback.

Ok. let's try Google Maps first. I'm good with that decision.
Not sure how Google Maps operates these days and if it still requires an API key or not, but just add it as an environment variable at first and we'll see how we make it permanent later.

Thanks for working on this.

helpotters:
## Gmaps vs Mapbox 

After comparing Google Maps and MapBox, I believe MapBox is much more user/developer-friendly. The time it took to get up an running with Google Maps was much longer than Mapbox. Mapbox also has great documentation to extend the map display even further. 

## What I added

I added the [mapkick gem](https://chartkick.com/mapkick) as it makes map displays a breeze, but also leaves room to implement it via controllers if need be. It's been updated as recently as last week, but I'm not sure what the protocol is on new gems.

But this gem makes adding a map display an one-liner in the field with straightforward options. 
```ruby 
<%= js_map [{latitude: @field.value[0], longitude: @field.value[1]}] %>
```
 
### MAPBOX_API_KEY

_**Important note:**_  The API Key _does not_ work as-is. 

Setting `ENV["MAPBOX_ACCESS_TOKEN"] was done manually in `initializers/mapbox.rb` by simply assigning it to a string. I was uncertain how environmental variables should be handled, so I left it like this for now. Mainly due to uncertainty about how avo handles credentials and the master key. 


## TODO and Double-checks

- [ ] Generate your own mapbox api key 
- [ ] Proper implementation of `ENV["MAPBOX_ACCESS_TOKEN"]`
- [ ] - I imported `mapkick` into `javascript/js/application.js`. Is this the right place for avo assets? 
- [ ] `city#coordinates` now returns an array (for now). `value[0]` doesn't scream latitude, but it will do for now. Suggestions welcome!


image _Here's some neat test coordinates_

helpotters:
# How it looks now! 
image

helpotters:
I can't figure out why I can't get `rails credentials:edit` to run. I noticed that there's an existing `credentials.yml.enc` file in the top level and dummy directories as well (probs unrelated). It's probably an issue with rbenv on my end though. 

Ideally, the `mailbox_api_key` would be configured via the credentials tool.

adrianthedev:
Hey, @helpotters. I'm rather swamped today, but I'll try to have a look and review/unblock this is necessary.

Thanks!

helpotters:
# bump!

helpotters:
5 specs are now passing for `location_tag_field_spec.rb`.

Items of note: 
- I'll try to refactor the specs for repetition. 
- Consider VCR for API-requests recording. I couldn't figure out how to get it to record the `js_map` rendering
- I had to use `MAPBOX_ACCESS_TOKEN="token_here" bundle exec rspec spec/system/avo/location_tag_field_spec.rb` to run `renders a map` spec. Without defining the token variable, the spec fails (which is correct because the map isn't rendered).

helpotters:
I'm really happy with how this is turning out. Thank you.

lib/avo/fields/location_field.rb
```diff
@@ -0,0 +1,70 @@
...
+        elsif get == :long
+          "#{id}[#{stored_as.last}]"
+        end
+      end
+
+      def as_lat_long_placeholder(get: nil)
+        if get == :lat
+          placeholder = "Enter #{stored_as.first}"
+        elsif get == :long
+          placeholder = "Enter #{stored_as.last}"
```
fastjames, line 32: You can skip assigning these to a variable, since they will be the return value of the `if` statement / the method itself.

```suggestion
          "Enter #{stored_as.last}"
```

spec/system/avo/location_tag_field_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  describe 'with regular value' do
+    # Eiffel Tower coordinates used
+    let!(:city) { create :city, latitude: 48.8584, longitude: 2.2945 }
+
+    context "show" do
+      it "renders a map" do
+        visit "/admin/resources/cities/#{city.id}"
+
+        expect(page).to have_css("canvas.mapboxgl-canvas")
```
fastjames, line 58: I examined this test to see why it was failing, and I think there may be several things going on here:

1. Headless chrome (which is where these tests are normally running in CI) appears to have issues with using WebGL. I don't have any great links with simple solutions, but you can read a little about [issues with getting webgl to run in headless](https://stackoverflow.com/questions/70948512/how-to-enable-webgl-in-headless-chrome-96-within-selenium-docker-project-to-r). I was able to replicate the failure locally on MacOS Ventura.
2. I tried running it locally in non-headless mode (HEADFULL=true bundle exec rspec spec/system/avo/location_tag_field_spec.rb:58) and while it still failed, the screenshot was different! Instead of the usual Failed to initialize WebGL message, the error shown was An API access token is required to use Mapbox GL. See https://www.mapbox.com/api-documentation/#access-tokens-and-token-scopes. That's very encouraging, because it looks like mapbox did attempt to start in that test run.

Avo currently includes webmock, so that seems to be the convention for mocking out external service dependencies.

helpotters: I'll look into mocking out the Mapbox API for the spec. Thank you for the review and the testing!

adrianthedev:
I ended up stubbing the `js_map` method in the tests. I think it's ok to not test the map live.

Thank you @helpotters for championing this feature and @fastjames for your input. I love it when the community comes together like this.

PR #1634, opened 2023-03-18
feature: add search_result_url and search_policy_class

glaucocustodio:
Add two new class attributes to `Avo::BaseResource` to further customize the search behaviour:

- search_policy_class: responsible for setting the policy used to authorize the search on the applied resource
- search_result_url: responsible for setting the url to be used in the search results

## Context
An Avo app contains the following classes: 

- ReportResource (used only by admin users to manage reports)
- Avo::Customers::ReportsController (used by customer users to view reports)
- ReportPolicy (allowing only admin users to manage ReportResource)
- Customers::ReportPolicy (allowing customer users to view reports from the controller aforementioned)

When a customer searches for a report on the global search, Avo 2.28 will use `ReportPolicy` to authorize the search and it will deny since a customer can't manage ReportResource.

In order to fix that, I am introducing `search_policy_class`. This class attribute will allow people to set which policy class must be used to authorize seaches:

```ruby
class ReportResource < Avo::BaseResource
  self.title = :title
  self.includes = []
  self.search_policy_class = Customers::ReportPolicy
end
```

Once the authorization is fixed with the `search_policy_class` above, another problem is found: Avo 2.28 will point to `/avo/resources/reports/:report_id` as url in the search results. Remember: customers can't manage reports, so, search results should point to `/avo/customers/reports/:report_id` instead, that's why `search_result_url` has been introduced:

```ruby
class ReportResource < Avo::BaseResource
  self.title = :title
  self.includes = []
  self.search_policy_class = Customers::ReportPolicy
  self.search_result_url = ->(report, main_app, params) do
    if params[:global] == "true"
      main_app.customer_report_path(report)
    end
  end
end
```

In the example above, if the search is global (performed possibly by a customer), search results will point to `customer_report_path`, otherwise, if it's performed on the ReportResource index page (only by an admin) it will point to  `/avo/resources/reports/:report_id` (sames as in Avo 2.28).

# Checklist:

- [ ] I have made corresponding changes to the [documentation](https://github.com/avo-hq/docs) -> will open a PR for that

## Manual review steps

1. Create a resource class, eg: ReportResource
2. Create a "tool" controller to read these resources
3. Create two policy classes: ReportPolicy to authorize ReportResource and Customers::ReportPolicy to authorize the tool controller
4. Perform a global search with an user that can only read the tool controller (not the ReportResource)

### DIFF
--- a/app/controllers/avo/search_controller.rb
+++ b/app/controllers/avo/search_controller.rb
@@ -27,8 +27,7 @@ module Avo
       resources
         .map do |resource|
           # Apply authorization
-          action = Avo.configuration.authorization_methods.stringify_keys['search'] || :search
-          next unless @authorization.set_record(resource.model_class).authorize_action(action, raise_exception: false)
+          next unless @authorization.set_record(resource.model_class).authorize_action(:search, raise_exception: false)
 
           # Filter out the models without a search_query
           next if resource.search_query.nil?
--- a/app/controllers/avo/search_controller.rb
+++ b/app/controllers/avo/search_controller.rb
@@ -27,7 +27,9 @@ module Avo
       resources
         .map do |resource|
           # Apply authorization
-          next unless @authorization.set_record(resource.model_class).authorize_action(:index, raise_exception: false)
+          action = Avo.configuration.authorization_methods.stringify_keys['search'] || :search
+          next unless @authorization.set_record(resource.model_class).authorize_action(action, raise_exception: false)
+
           # Filter out the models without a search_query
           next if resource.search_query.nil?
 
@@ -120,10 +122,16 @@ module Avo
       models.map do |model|
         resource = avo_resource.dup.hydrate(model: model).hydrate_fields(model: model)
 
+        record_path = if resource.search_result_path.present?
+          Avo::Hosts::ResourceRecordHost.new(block: resource.search_result_path, resource: resource, record: model).handle
+        else
+          resource.record_path
+        end
+
         result = {
           _id: model.id,
           _label: resource.label,
-          _url: resource.record_path,
+          _url: record_path
         }
 
         if App.license.has_with_trial(:enhanced_search_results)
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -129,6 +129,7 @@ RSpec.configure do |config|
   # examples within a transaction, remove the following line or assign false
   # instead of true.
   config.use_transactional_fixtures = true
+  config.filter_run_when_matching :focus
 
   config.before(:each, type: :system) { driven_by test_driver }
 
--- a/CONTRIBUTING.MD
+++ b/CONTRIBUTING.MD
@@ -68,6 +68,17 @@ gem 'avo', path: '../avo'
 
 Avo's assets will not show up by default, resulting in 404 errors on `/avo-assets/avo.base.js` and `/avo-assets/avo.base.css`. To avoid this, you need to compile the asset bundles, and symlink them into `public/avo-assets`.
 
+First, make sure you have `yarn` installed and then install Avo's dependencies:
+```bash
+yarn install
+```
+
+Run the first build to generate the files `app/assets/builds/avo.base.js` and `app/assets/builds/avo.base.css`:
+
+```bash
+yarn build
+```
+
 Create symlinks for compiled assets into the `public` directory. You'll only need to do this once.
 
 ```bash
--- a/app/components/avo/views/resource_index_component.html.erb
+++ b/app/components/avo/views/resource_index_component.html.erb
@@ -31,7 +31,7 @@
     <% end %>
     <% c.body do %>
       <div class="flex flex-col xs:flex-row xs:justify-between space-y-2 xs:space-y-0 py-4 <%= 'hidden' if @resource.search_query.nil? && @filters.empty? && available_view_types.count <= 1 %>">
-        <% unless hide_search_input %>
+        <% if show_search_input %>
           <div class="flex items-center px-4 w-64">
             <%= render partial: 'avo/partials/resource_search', locals: {resource: @resource.route_key, via_reflection: via_reflection} %>
           </div>
--- a/app/components/avo/views/resource_index_component.rb
+++ b/app/components/avo/views/resource_index_component.rb
@@ -125,6 +125,9 @@ class Avo::Views::ResourceIndexComponent < Avo::ResourceComponent
   end
 
   def hide_search_input
+    if @resource.authorization.has_action_method?("search")
+      return true unless @resource.authorization.authorize_action("search", raise_exception: false)
+    end
     return true unless @resource.search_query.present?
 
     field&.hide_search_input || false
--- a/app/components/avo/views/resource_index_component.rb
+++ b/app/components/avo/views/resource_index_component.rb
@@ -125,6 +125,7 @@ class Avo::Views::ResourceIndexComponent < Avo::ResourceComponent
   end
 
   def hide_search_input
+    # Hide the search if the authorization prevents it
     if @resource.authorization.has_action_method?("search")
       return true unless @resource.authorization.authorize_action("search", raise_exception: false)
     end
--- a/app/components/avo/views/resource_index_component.rb
+++ b/app/components/avo/views/resource_index_component.rb
@@ -124,14 +124,19 @@ class Avo::Views::ResourceIndexComponent < Avo::ResourceComponent
     @resource.resource_description
   end
 
-  def hide_search_input
-    # Hide the search if the authorization prevents it
-    if @resource.authorization.has_action_method?("search")
-      return true unless @resource.authorization.authorize_action("search", raise_exception: false)
-    end
-    return true unless @resource.search_query.present?
+  def show_search_input
+    return false unless authorized_to_search?
+    return false unless @resource.search_query.present?
+    return false if field&.hide_search_input
 
-    field&.hide_search_input || false
+    true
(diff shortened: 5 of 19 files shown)


app/controllers/avo/search_controller.rb
```diff
@@ -120,12 +121,18 @@ def apply_search_metadata(models, avo_resource)
       models.map do |model|
         resource = avo_resource.dup.hydrate(model: model).hydrate_fields(model: model)
 
+        if resource.search_result_url
+          url = resource.search_result_url.call(model, main_app, params)
```
glaucocustodio, line 125: I had to pass `main_app` to be able to use url helpers in the lambda.
I had to pass `params` to be able to check if the search is global or not

spec/features/avo/search_with_policy_class_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      end
+
+      it "can find the city" do
+        visit url
+        open_global_search_box
+        expect_search_panel_open
+
+        write_in_search "São Paulo"
+        expect(page).to have_content "São Paulo"
+        # expect(page).to have_link(href: "/any/custom/#{city.id}")
```
glaucocustodio, line 26: I couldn't get that commented assert passing, I've seen this url in the search response but for somehow I can't see it in the page html during test

app/controllers/avo/search_controller.rb
adrianthedev: If we use the Evaluation Host we will be able to delegate `main_app` to the `view_context`.
See [this](https://github.com/avo-hq/avo/blob/main/lib/avo/fields/base_field.rb#L19-L22).

spec/features/avo/search_with_policy_class_spec.rb
adrianthedev: You might need to regex it from the page.
```suggestion
        # expect(page).to have_link(href: /\/any/custom/#{city.id}/)
```

adrianthedev (commented):
Thanks for working on this @glaucocustodio! This is a good feature indeed!

Regarding `search_policy_class`; Instead of having another policy to work with that only has one method `index?`, how about we use the policy of that resource (which you can customize using `authorization_policy`) and add another method `search?` (which once can customize it using [different-policy-methods](https://docs.avohq.io/2.0/authorization.html#using-different-policy-methods)).
This would DRY up the code IMHO.

The `search_result_url` is 👨‍🍳🤌 (chef's kiss). One thing I'd do is use an [Evaluation host](https://docs.avohq.io/2.0/evaluation-hosts.html). That will allow us not to pass in variables to that block and for better future-proofing.
Let me know if I can help with that.

glaucocustodio:
> Thanks for working on this @glaucocustodio! This is a good feature indeed!
> 
> Regarding `search_policy_class`; Instead of having another policy to work with that only has one method `index?`, how about we use the policy of that resource (which you can customize using `authorization_policy`) and add another method `search?` (which once can customize it using [different-policy-methods](https://docs.avohq.io/2.0/authorization.html#using-different-policy-methods)). This would DRY up the code IMHO.
> 
> The `search_result_url` is 👨‍🍳🤌 (chef's kiss). One thing I'd do is use an [Evaluation host](https://docs.avohq.io/2.0/evaluation-hosts.html). That will allow us not to pass in variables to that block and for better future-proofing. Let me know if I can help with that.

Sorry, I didn't give you the full context, I have the following policies:
```ruby
class AdminPolicy < ApplicationPolicy
  def index?
    allowed?
  end

  def show?
    allowed?
  end

  def create?
    allowed?
  end

  def update?
    allowed?
  end

  def destroy?
    allowed?
  end

  def upload_attachments?
    allowed?
  end

  def download_attachments?
    allowed?
  end

  def delete_attachments?
    allowed?
  end

  class Scope < Scope
    def resolve
      scope.all
    end
  end

  private

  def allowed?
    user.admin?
  end
end

# used by the Resource
class ReportPolicy < AdminPolicy
end

# used by the tool controller
module Customers
  class ReportPolicy < ApplicationPolicy
    def index?
      true
    end

    def show?
      true
    end
  end
end
```
`Customers::ReportPolicy` is used to authorize two tool controllers actions: index and show (index is also used by search).

I would rather managing customer policies in `Customers::ReportPolicy` (including search which uses ìndex?`) and keep `ReportPolicy` for managing admin policies related to the resource CRUD, that's why I think would be nice having `search_policy_class`.

app/controllers/avo/search_controller.rb
glaucocustodio: I've changed to use the search host, but I think we still need to pass `model` since the evaluation host does not seem to have the resource model instance. What do you think?

spec/features/avo/search_with_policy_class_spec.rb
glaucocustodio: The regex doesn't work also.

During the test, if I print `url` I can see `/any/custom/1`, but not in `page.body`, that's why the assert fail 🤔

adrianthedev:
I understand your usage. I just think it would fall more in-line with what we currently have which is all things related to the authorization of a resource (CRUD, associations, attachments) are controlled from one policy.
For example, for `CityResource` you have `CityPolicy`.
So adding search there makes sense IMHO.
You could still implement that `allowed?` pattern you used showed me with 

```ruby
def search?
  allowed?
end
```

Otherwise, everyone will have to add a new policy for search with just one method `index?`.

So it would be something like

```ruby
class AdminPolicy < ApplicationPolicy
  def index?
    allowed?
  end

  def show?
    allowed?
  end

  def create?
    allowed?
  end

  ...


  def search?
    allowed?
  end
end
```

Does that make sense?


Regarding the evaluation host, apologies, I didn't provide an example for that. I'll push some code to show what I mean.

adrianthedev:
Check out my latest commit with the `ResourceRecordHost` implemented.
Now we can have a block without parameters for the `search_result_path`.

```ruby
  self.search_result_path = -> do
    # you have access to resource, record, view_context and more from ResourceRecordHost
    avo.resources_city_path record, custom: "yup"
  end
```

glaucocustodio:
> I understand your usage. I just think it would fall more in-line with what we currently have which is all things related to the authorization of a resource (CRUD, associations, attachments) are controlled from one policy. For example, for `CityResource` you have `CityPolicy`. So adding search there makes sense IMHO. You could still implement that `allowed?` pattern you used showed me with
> 
> ```ruby
> def search?
>   allowed?
> end
> ```
> 
> Otherwise, everyone will have to add a new policy for search with just one method `index?`.
> 
> So it would be something like
> 
> ```ruby
> class AdminPolicy < ApplicationPolicy
>   def index?
>     allowed?
>   end
> 
>   def show?
>     allowed?
>   end
> 
>   def create?
>     allowed?
>   end
> 
>   ...
> 
> 
>   def search?
>     allowed?
>   end
> end
> ```
> 
> Does that make sense?
> 
> Regarding the evaluation host, apologies, I didn't provide an example for that. I'll push some code to show what I mean.

How would Avo look for `search?` instead of `index?` when authorising a search? if I override index like below would mess with other policies no?!
```
ruby
config.authorization_methods = {
    index: 'search?',
}
```

adrianthedev:
It would be just `search?` by default.

```ruby
config.authorization_methods = {
    search: 'search?',
}
```

So instead of thinking of search as something like the `index` method/action, it has it's own `search?` method where you can control only the search.

glaucocustodio:
> ```ruby
> search: 'search?',
> ```

But for that we need to replace `next unless @authorization.set_record(resource.model_class).authorize_action(:index, raise_exception: false)` by `next unless @authorization.set_record(resource.model_class).authorize_action(:search, raise_exception: false)` in `app/controllers/avo/search_controller.rb`, right?!

adrianthedev:
Yes, correct. 
And probably do something like this before so we don't send `:search` but the one configured in by the user.

```ruby
# fetch the configured name or use the default :search
action = Avo.configuration.authorization_methods.stringify_keys['search'] || :search

next unless @authorization.set_record(resource.model_class).authorize_action(action, raise_exception: false)
```

spec/features/avo/search_with_policy_class_spec.rb
glaucocustodio: should I remove this commented line? Do you have any other suggestion on how to assert that?

glaucocustodio:
We don't need this bit:
```ruby
config.authorization_methods = {
    search: 'search?',
}
```
It will work out of the box now, thanks for the suggestion 🙃

adrianthedev:
@glaucocustodio I'll have a look and merge it tomorrow. Today was swamped for me.

spec/rails_helper.rb
```diff
@@ -129,6 +129,7 @@ def driver_options(headless: false)
   # examples within a transaction, remove the following line or assign false
   # instead of true.
   config.use_transactional_fixtures = true
+  config.filter_run_when_matching :focus
```
glaucocustodio, line 132: this is useful for running a single test when we want

glaucocustodio:
> Today was swamped for me.

No worries, hopefully the tests will pass next time

adrianthedev:
This looks good!
I pushed some code to hide the search box from the individual index pages too when un-authorized.

CleanShot 2023-03-21 at 13 27 46@2x

adrianthedev:
Thank you for working on this @glaucocustodio!

You championed this feature from 0 to 1! Good job!

glaucocustodio:
Thank you for guiding me to the best solution, I am proud to contribute back to Avo 💙

PR #1667, opened 2023-04-04
improvement: another approach on attachments field level authorization

### LINKED ISSUE #1662: The download button is still visible if the specific file policy method returns false
### Describe the bug

When using authorization with file fields the download button on each individual file is not hidden if the individual policy method `download_images?` returns false. The button gets hidden only when the general method `download_attachments?` returns false.

### Steps to Reproduce

```ruby
def download_attahcments?
  true
end

def download_images?
  false
end
```

The `images` `files` field still displays the delete button when it shouldn't.

Paul-Bob:
# Description

Fixes #1662

# UPDATE

We have discovered that assigning different weights to two methods that police the same thing can be a complex and challenging task. As a result, we have decided to make `upload_attachments?`, `download_attachments?`, and `delete_attachments?` obsolete.

From now on, only field level authorization will be considered. If it is defined and returns true, the action will be granted; otherwise, it will not.

## Previous approach

~~How it works:~~

~~`upload_attachments?`, `download_attachments?` and `delete_attachments?` are first level authorization for attachments. 
`upload_file_name?`, `download_file_name?` and `delete_file_name?`  are second level authorization for each field.~~

~~Lets pick uploads example:~~

~~If `upload_attachments?` is NOT defined only `upload_file_name?` will count for authorize an action on a field...
If `upload_attachments?` is defined,  `upload_file_name?` will count  only if `upload_attachments?` return true.~~

~~Why? This way you can do something like:~~

~~All admins and managers can upload attachments but only admins can upload that important document...~~

```ruby
def upload_attachments?
  user.admin? || user.manager?
end

def upload_important_document?
   user.admin?
end
```

~~OR~~

~~Nobody can upload any attachments but admins can upload that important document...~~

```ruby
def upload_important_document?
   user.admin?
end
```

~~OR~~

~~Anyone can upload attachments but nobody can upload that important document.~~

```ruby
def upload_attachments?
  true
end

def upload_important_document?
   false
end
```

~~NOTICE:~~

~~The follow example don't make much sense to use it...
If upload_attachments are false, field level authorization will not be checked for uploads.
If you want to turn all uploads off and allow only specific uploads do NOT  define `uploads_attachments?`~~
```ruby
def upload_attachments?
  false
end

def upload_important_document?
   true
end
```
~~

# Checklist:

### DIFF
--- a/app/controllers/avo/attachments_controller.rb
+++ b/app/controllers/avo/attachments_controller.rb
@@ -40,9 +40,7 @@ module Avo
     private
 
     def authorized_to(action)
-      if @resource.authorization.authorize_action("#{action}_attachments?".to_sym, raise_exception: false)
-        @resource.authorization.authorize_action("#{action}_#{params[:attachment_name]}?", record: @model, raise_exception: false)
-      end
+      @resource.authorization.authorize_action("#{action}_#{params[:attachment_name]}?", record: @model, raise_exception: false)
     end
   end
 end
--- a/lib/avo/fields/concerns/file_authorization.rb
+++ b/lib/avo/fields/concerns/file_authorization.rb
@@ -22,32 +22,7 @@ module Avo
 
         private
 
-        # First we check if user have authorization to act on attachments
-        # If attachemnts level authorization is defined and user do NOT have authorization to act on attachments
-        # We return false
-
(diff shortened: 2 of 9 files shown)


adrianthedev (approved):
I left a few comments on the docs repo.

PR #1679, opened 2023-04-13
Resource index map component

fastjames:
# Description
Now that avo has a location field type, add a new view_type to the resource index page: `map`. This view displays a map containing markers for all of the resources.

Note that this is still a work in progress, so the approach may change considerably.

# Checklist:

## Manual review steps

1. Step 1
1. Step 2


adrianthedev:
Hey @fastjames. Thanks for the contribution!

This feature definitely goes in the same direction we were thinking of going too. I'd love to have you champion this feature going forward.

In Avo 3 we have prototyped a plugin system where someone could just plug a view like this from an external gem.

So, let's keep this PR open, "finish it" and let's see if this will land in Avo 2 or Avo 3.

app/helpers/avo/resources_helper.rb
```diff
@@ -66,5 +66,37 @@ def item_select_all_input
...
+      return [] unless resources.first.record.respond_to? :coordinates
+
+      resources.map do |resource|
+        resource_location_marker(resource)
+      end.compact
+    end
+
+    def resource_map_options(resource, map_type: :index)
+      options_method = "resource_#{map_type}_map_options".to_sym
+      if resource.respond_to?(options_method)
```
fastjames, line 80: For the map marker configuration and the map options configuration, right now I'm using method defined on the Resource class. Would it be more conventional to define those another way, either as class methods (like `search_query`) or via a new `location_marker` DSL call?

lib/avo/base_resource.rb
```diff
@@ -295,10 +295,17 @@ def available_view_types
       view_types = [:table]
 
       view_types << :grid if get_grid_fields.present?
+      view_types << :map if mappable?
 
       view_types
     end
 
+    def mappable?
```
fastjames, line 303: There is a slight mismatch between this method and the `resource_location_markers` helper method above. This one returns true if the resource has _any_ location fields, while the helper method only returns markers if the resource has a `coordinates` method.

I have not thought much about allowing the developer to specify a location field name other than `coordinates` for the map. It seems like the GridCollector provides a pattern for doing that sort of thing.

app/assets/svgs/map-empty-state.svg
fastjames, line 1: I noticed that the other two already-existing `*-empty-state.svg` files produce these invalid image errors as well, in MacOS Finder and here.

fastjames:
I am not sure about the cause of the codeball failure. It looks like the last action it took (or attempted to take) was to create a `codeball:needs-careful-review` tag to apply to this PR. Maybe the codeball plugin does not have permission to create new tags for this repo?

app/components/avo/index/resource_map_component.html.erb
```diff
@@ -0,0 +1,10 @@
+<% if @resources.present? %>
+  <div
+    data-selected-resources-name="<%= @resource.model_key %>" data-selected-resources="[]">
+    <%= js_map(helpers.resource_location_markers(@resources), **helpers.resource_map_options(@resource, map_type: :index))  %>
```
fastjames, line 4: Another thing I'd like to support here is adding additional markers to the map. The initial use would be for my "near this location" filter, as I'd like to display a marker indicating the search location itself. Others might want to include additional markers for other things on the map.

adrianthedev:
This is cool! Thank you for working on this feature @fastjames 🚀

I'll leave a few thoughts here. These aren't tasks that only you should take, but a general approach to this feature.

- [ ] let's make this configurable. The `self.map = {` configuration seems to be the closest to what we need. it will allow us to configure the view with multiple options.
- [ ] the view should have the option to render the table or not.
- [ ] the view should have the option to render map and table on the top, right, or left (airbnb style)
- [ ] the configuration should take a `record_location` option which should have the `record` available and the user should return a `[latitude, longitude]` tuple that the map will use.
- [ ] the configuration object can take a `additional_locations` option where you can pass in a block with more markers
- [ ] **later** on this configuration object we can add more options that we can forward to mapkick (controls, zoom, etc.)

These are my general thoughts on this. What do you think?

CleanShot 2023-04-18 at 11 49 35@2x

lib/avo/base_resource.rb
fastjames: I resolved this by making the method look for the field by both type and name, sot hat it matches the helper method. If it's not correct, it is at least consistent.

fastjames:
> This is cool! Thank you for working on this feature @fastjames 🚀
> 
> I'll leave a few thoughts here. These aren't tasks that only you should take, but a general approach to this feature.
> 
>     * [ ]  let's make this configurable. The `self.map = {` configuration seems to be the closest to what we need. it will allow us to configure the view with multiple options.
> 
>     * [ ]  the view should have the option to render the table or not.
> 
>     * [ ]  the view should have the option to render map and table on the top, right, or left (airbnb style)
> 
>     * [ ]  the configuration should take a `record_location` option which should have the `record` available and the user should return a `[latitude, longitude]` tuple that the map will use.
> 
>     * [ ]  the configuration object can take a `additional_locations` option where you can pass in a block with more markers
> 
>     * [ ]  **later** on this configuration object we can add more options that we can forward to mapkick (controls, zoom, etc.)
> 
> 
> These are my general thoughts on this. What do you think?
> 
> CleanShot 2023-04-18 at 11 49 35@2x

I think these are great pointers for improvements. I'll dig into them and see what works well.

fastjames:
@adrianthedev thanks again for your very helpful feedback on this. I believe I have implemented most of the changes you recommended -- the configuration is now handled via a class attr, the map/table layout is configurable, and there's also a mobile breakpoint to hide the table (users can still toggle to it via the view toggle buttons). The only item I have not addressed is the ability to provide additional markers to the map. That is a little more complex for my desired case since I'd like the marker to come from the filter (so it may require a little bit of stimulus magic). Other than that, I think I have this feature in a pretty good place now. As always, your feedback is heartily appreciated.

app/helpers/avo/resources_helper.rb
fastjames: These configuration points were moved to elements of a hash class attr.

fastjames:
I have opened a PR to add content to the doc site for this feature:

https://github.com/avo-hq/avodocs/pull/36

fastjames:
@adrianthedev I have added tests and (in the other repo) docs for the feature. Is there anything you think needs to be included in this iteration of the feature?

adrianthedev (commented):
Amazing job @fastjames! I really appreciate the level of detail here.

A few things come to mind.

- we should automatically filter out the entries which don't have valid coordinates. I used something like this 
```ruby
resources
  .map do |resource|
    marker_proc.call(record: resource.record)
  end
  .compact
  .filter do |coordinates|
    coordinates[:latitude].present? && coordinates[:longitude].present?
  end
```
- if the table is hidden, then the map should span the whole width automatically using CSS. We should provide this safeguard in case that the table has a dynamic rule of show/hide.
- the CSS rules around the `.map-view-container` are a bit complex, and I wonder if we could achieve them using just the regular Tailwind classes. I can help with that next week after I push out the Avo 3 beta.

I also left a few comments around the helper methods, but otherwise, this PR is 90% there. Thank you for that.

Apologies for the delay. I had to finish a few things around [Friendly.rb](https://friendlyrb.com) and Avo 3.
Are you in a hurry with this feature, or could we have it in next week?

app/components/avo/fields/location_field/show_component.html.erb
```diff
@@ -1,6 +1,6 @@
 <%= field_wrapper **field_wrapper_args do %>
   <% if field.value_present? %>
-    <%= js_map [{latitude: field.value[0], longitude: field.value[1]}] %>
+    <%= js_map [{latitude: field.value[0], longitude: field.value[1]}], **helpers.resource_mapkick_options(@resource.map) %>
```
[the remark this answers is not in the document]
fastjames: I added it because I thought it made sense for consistency, but now that you point it out that argument seems pretty week. I will remove that change. :+1:

app/helpers/avo/resources_helper.rb
```diff
@@ -66,5 +69,50 @@ def item_select_all_input
...
+      resources.map do |resource|
+        marker_proc.call(record: resource.record)
+      end.compact
+    end
+
+    def resource_mapkick_options(map_options)
+      return {} unless map_options.present?
+
+      map_options[:mapkick_options] || {}
+    end
```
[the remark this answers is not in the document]
fastjames: Great advice -- moving these methods inside the component class made them much cleaner!

fastjames:
> Amazing job @fastjames! I really appreciate the level of detail here.
> 
> A few things come to mind.
> 
>     * we should automatically filter out the entries which don't have valid coordinates. I used something like this
> 
> 
> ```ruby
> resources
>   .map do |resource|
>     marker_proc.call(record: resource.record)
>   end
>   .compact
>   .filter do |coordinates|
>     coordinates[:latitude].present? && coordinates[:longitude].present?
>   end
> ```
I added this filter when I moved that method into the component class, so it should be present now.
> 
>     * if the table is hidden, then the map should span the whole width automatically using CSS. We should provide this safeguard in case that the table has a dynamic rule of show/hide.
> 
>     * the CSS rules around the `.map-view-container` are a bit complex, and I wonder if we could achieve them using just the regular Tailwind classes. I can help with that next week after I push out the Avo 3 beta.
I am very new to using Tailwind, but I took a swing at this and I am pretty pleased with the results. I am sure it will require a little more tweaking, but I was able to remove all of the changes from `avo.base.css`. TIL!
> 
> I also left a few comments around the helper methods, but otherwise, this PR is 90% there. Thank you for that.
As suggested, moving the helper methonds into the component class was a big win. :100:

> Apologies for the delay. I had to finish a few things around [Friendly.rb](https://friendlyrb.com) and Avo 3. Are you in a hurry with this feature, or could we have it in next week?
I would always love to have it sooner (see my github username), but given your schedule next week seems reasonable.

fastjames:
Hmm, that test failure looks like a false negative -- I don't think any of my changes touched code that would have affected it. Hopefully a re-run will cure things.

adrianthedev:
Thanks for all the work @fastjames! 

I pushed a couple of tweaks around code styling and renamed the option from `map` to `map_view`.

PR #1683, opened 2023-04-15
feature: files view type

### LINKED ISSUE #1661: Files field with listing view
### Feature

When you have a lot of files that are not a gallery of images, it might be rather difficult to see the filenames as they are truncated and there is too much wasted space on the gallery tiles.

Let's implement a list view that can be enabled by the `view_type: :list` option on the field.

Please try to use the list view design from this [ticket](https://github.com/avo-hq/avo/issues/1131).

### Current workarounds

Create a custom field with a custom view.

CleanShot 2023-04-04 at 20 27 16@2x

Paul-Bob:
# Description

Fixes #1661
## `field :files, as: :files`
image

## `field :files, as: :files, view_type: :list`
image

[files.webm](https://user-images.githubusercontent.com/69730720/232238439-3ed1210f-c602-4691-aeb5-2b2c98eedb64.webm)

## `field :files, as: :files, view_type: :list, hide_view_type_changer: true`
image

# Checklist:

- [x] I have made corresponding changes to the [documentation](https://github.com/avo-hq/avodocs/pull/34)

## Manual review steps

1. Step 1
1. Step 2


adrianthedev (approved):
I applied some cosmetic changes.

Otherwise... fantastic job! Thank you!

Can we add some tests for it?

FLX-0x00:
VERY nice - we are waiting for this as our implementation works not so fine :/

adrianthedev:
[2.30.1.pre1.pr1683](https://rubygems.org/gems/avo/versions/2.30.1.pre1.pr1683) released with this PR and all the other changes in the latest draft release [here](https://github.com/avo-hq/avo/releases).

FLX-0x00:
Will give it a shot in our staging env

FLX-0x00:
image

`  field :downloads, as: :files, view_type: :list, hide_view_type_changer: false, only_on: :show` 

does not look right. also the alignment not fits to other fields. tried with another browser without any caches. Any hints?

app/components/avo/fields/files_field/edit_component.html.erb
```diff
@@ -1,5 +1,5 @@
-<%= field_wrapper **field_wrapper_args, full_width: true do %>
-  <%= render Avo::Fields::Common::FilesListViewerComponent.new(field: @field, resource: @resource) if @field.value.present? %>
+<%= field_wrapper **field_wrapper_args, stacked: true do %>
```
Paul-Bob, line 1: @adrianthedev  what about make files field stacked attribute true by default but always get the stacked from the field? Someone may want to have this false and would not be possible without monkey patches

Paul-Bob: Same apply on show component

adrianthedev: Yes. My thoughts exactly. Have the stackable option control it for the wrapper too.

adrianthedev:
@FLX-0x00 that behavior is intended. I made it full width so you can have more space to view your content.

We'll probably have the `stackable` option control that.

FLX-0x00:
okay, observed something on firefox

image


and on chrome
image


`field :downloads, as: :files, view_type: :list,  hide_view_type_changer: true`

Firefox will not hande the top paddign and on chrome the scrollbar is always visible.

lib/avo/fields/files_field.rb
```diff
@@ -37,6 +40,14 @@ def fill_field(model, key, value, params)
 
         model
       end
+
+      def viewer_component
+        "Avo::Fields::Common::Files::ViewType::#{view_type.to_s.capitalize}Component".constantize
+      end
+
+      def view_type
```
adrianthedev, line 48: I think I'd like to change this implementation. We should not rely on the params in a file field to get the view type. If we need it somewhere we should inject it to the method as a dependency from a context where we do have the `params` (aka the view layer).

lib/avo/fields/files_field.rb
```diff
@@ -37,6 +40,14 @@ def fill_field(model, key, value, params)
 
         model
       end
+
+      def viewer_component
```
adrianthedev, line 44: What is the `viewer_compoent` and why does this live on the file field? Isn't this a "view layer" dependency?

We're passing params inside a file class to fetch the name of view component (outside-inside-outside). It seems a bit complex.

Let's move this somewhere closer to the view (maybe on a view component)?

Does this make sense?

adrianthedev (changes requested):

adrianthedev:
Hey @FLX-0x00. Just pushed https://rubygems.org/gems/avo/versions/2.30.1.pre2.pr1683.

> Firefox will not hande the top padding

Hmmm... that's weird. I checked on Firefox (mac) and @Paul-Bob checked on Firefox (linux) and we can't seem to replicate it. Can you try to froce-refresh and make sure the new assets are getting loaded and nothing gets wrongly cached?

> chrome the scrollbar is always visible

I pushed a fix for that now. The scrollbar should auto-hide like on a mac. Can you confirm it's fixed?

> also the alignment not fits to other fields

With his new version you can mark the field as `stacked: true|false` to make it in-line with the other fields or full width in order to use the whole space.

FLX-0x00:
> [2.30.1.pre2.pr1683](https://rubygems.org/gems/avo/versions/2.30.1.pre2.pr1683).

Perfect. now looks fine in both firefox and chrome browsers! Awesome work.

adrianthedev:
Ok. I think I have an idea of what causes it.

adrianthedev:
https://rubygems.org/gems/avo/versions/2.30.1.pre3.pr1683 released with that fix.
Heads-up we renamed `hide_view_type_changer` to `hide_view_type_switcher`.

FLX-0x00:
> https://rubygems.org/gems/avo/versions/2.30.1.pre3.pr1683 released with that fix. Heads-up we renamed `hide_view_type_changer` to `hide_view_type_switcher`.

Works as expected. Nice job on this


image

lib/avo/fields/files_field.rb
Paul-Bob: I'm not sure why I let that on the field, in some point of the iteration I was using that on several independent components, I noticed that now it's only used in one component, you're right, this do not belongs here, will fix on the next commit

lib/avo/fields/files_field.rb
Paul-Bob: Agree, changed on last PR

adrianthedev:
There's a bug with the success toast not appearing after attachment delete.

PR #1698, opened 2023-04-20
Chore: support view component v3 by updating slot setter methods to new `with_*` API

stevegeek:
# Description

A new API for setting slots on ViewComponent instances was added in v2.54.0. The old method was deprecated at that point. 

In v3 of ViewComponent the old slot setter methods have been removed.

This PR changes the slot setter calls to the new `with_*` methods and also:

- removes some component test files for components which have seemingly been removed.
- sets the gem dependency of ViewComponent to at least 2.54.0
- Updates hotwire-livereload for Ruby 3.2 support
- fixes a seemingly unrelated test failure with the 'Loading' text on views where the text seems to always be singular https://github.com/avo-hq/avo/blob/52ffaf7d47fd34a425b66a68873ed9040a04e665/lib/avo/fields/base_field.rb#L125

# Checklist:

## Manual review steps

1. Check that resource pages render  correctly


stevegeek:
One question I had hear is the gemspec says Avo works with Ruby >= 2.6, but Ruby 3 is the lowest version in the text matrix and view_component 2.72 dropped support for 2.6, maybe the gemspec Ruby version needs updating?

adrianthedev:
Thanks for the nudge and the work @stevegeek!

Issue #1706, opened 2023-04-24
`ActiveStorage::Current.url_options` should be a hash of url options


stevegeek:
### Describe the bug

Avo sets `ActiveStorage::Current.url_options` to a string (https://github.com/avo-hq/avo/blob/main/lib/avo/app.rb#L66), however in Rails 7 this should be a hash, Eg see https://github.com/rails/rails/blob/main/activestorage/app/controllers/concerns/active_storage/set_current.rb

This manifests as a problem when using the `DiskService` which tries to splat it https://github.com/rails/rails/blob/main/activestorage/lib/active_storage/service/disk_service.rb#L141

I will open a PR for a fix

### System configuration
**Avo version**: v2.30.2

**Rails version**: 7+

**Ruby version**: any?

**License type**:

 - [ ] Community
 - [x] Pro

**Are you using Avo monkey patches or overriding views or view components?**

 - [ ] Yes. If so, please post code samples.
 - [x] No

#### Impact

 - [ ] High impact
 - [ ] Medium impact
 - [x] Low impact

#### Urgency

 - [ ] High urgency
 - [ ] Medium urgency
 - [x] low urgency

PR #1707, opened 2023-04-24
Fix: ActiveStorage current url_options should be set as a hash of url options

stevegeek:
# Description

`ActiveStorage::Current.url_options` is set as a hash for Rails 7.

Fixes https://github.com/avo-hq/avo/issues/1706

# Checklist:


adrianthedev:
I wonder how it worked before now. I know I used Avo in a Rails 7 project. I'm not the only one either.

PR #1714, opened 2023-04-26
Add autocomplete field option

fastjames:
# Description
In order to allow admins to add `autocomplete` attributes to form fields (e.g. to support address autofill from mapbox), create a new `autocomplete` field option. Default should be no `autocomplete` attribute at all on the form input element.

Fixes # (issue)

# Checklist:

## Manual review steps

1. Step 1
1. Step 2


adrianthedev (approved):
The PR looks good!
I only have one question. Do you think we'd need the autocomplete attribute on any other fields?

PR #1715, opened 2023-04-27
Add Rails 7.0.x to the test matrix

fastjames:
# Description
In order to ensure compatibility with the latest rails release, add `7.0` to the test matrix in github CI.

Fixes # (issue)

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,6 +31,7 @@ jobs:
         rails:
           - '6.0'
           - '6.1'
+          - '7.0'
         exclude:
           - ruby: '3.2.2'
             rails: '6.0'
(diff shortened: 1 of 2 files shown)


fastjames:
Hmm, those failures on 3.2.2 / 7.0 look like they might be legitimate.

fastjames:
After reviewing the other seven test failures, I found that applying the fix I found for the first failure fixed the other 7 as well.

adrianthedev:
Yes. I agree that the fix is in the right place. Fix it at the root.

I think one of the reasons I haven't added 7 to the matrix yet was that some gem (might have been pagy. can't remember exactly) needed a special version and `appraisal` (the tool that generates the gemfiles) couldn't generate it properly. 
I'm fuzzy of the details, but I want to leave this comment here for posterity 😅

Thanks @fastjames!

PR #1720, opened 2023-05-02
[Fix] When initializing an Action any class variables should be set to nil if appropriate

stevegeek:
# Description

Hey all. I think I have spotted an issue here and this is my suggested fix, but Im not 100% sure why the original `present?` checks are there, so please let me know if Im missing something fundamental!

When initializing an Action any class variables should be set to nil if no value provided. For example when on an `Index` view, `model` == `nil` intentionally. 

The existing implementation behaves in such a way that breaks `.message` for example. 

Lets say you visit a show page, an Action is initialized with `model` set to the current record you are viewing. Thus `self.class.model` is set to it.

if you then navigate to the `index` page, the Action is initialized with `model == nil`, but the `present?` check means `self.class.model` still points to previous record instance. Therefore when `message` is called `record` references the last visited show page model and you will possibly generate a message for that record rather that the Index page.

# Checklist:

- [ ] I have added tests that prove my fix is effective or that my feature works  (Im not sure where to add test for this ...)

### DIFF
--- a/lib/avo/base_action.rb
+++ b/lib/avo/base_action.rb
@@ -58,10 +58,10 @@ module Avo
     end
 
     def initialize(model: nil, resource: nil, user: nil, view: nil, arguments: {})
-      self.class.model = model if model.present?
-      self.class.resource = resource if resource.present?
-      self.class.user = user if user.present?
-      self.class.view = view if view.present?
+      self.class.model = model
+      self.class.resource = resource
+      self.class.user = user
+      self.class.view = view
       @arguments = arguments
 
       self.class.message ||= I18n.t("avo.are_you_sure_you_want_to_run_this_option")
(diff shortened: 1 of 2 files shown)


adrianthedev:
Hey @stevegeek. I had a look at this PR today and wanted to merge it bu I'm getting a few test fails. @Paul-Bob is going to investigate later on tonight.

Paul-Bob:
Hey @stevegeek and @adrianthedev.

First of all thank you @stevegeek for identifying and resolving this issue. While this PR successfully addresses the problem, I am uncertain about the approach taken. We can merge this fix for the next release, but I suggest revisiting the code and replacing class attributes with instance attributes on the `BaseAction` class.

Regarding the specs, some of them are failing due to the updated behavior. Specifically, the `ReleaseFish` `self.message` block fails because `record` becomes `nil` on the `index` view after executing the action in the `show` view. It is possible that this did not occur earlier because the class attribute retained an outdated `record` from a different spec.

I have made a commit to this PR to resolve the situation, and all the specs now pass successfully.

adrianthedev:
Ok. that hypothesis makes sense.

Issue #1719, opened 2023-05-02
When installing Avo detect if the app has a User model and create a resource for it.


adrianthedev:
### Feature

From [here](https://discord.com/channels/836637622432170028/1101956759729483827/1102836906645016586).

Most apps have a `User` model. When you install Avo in an existing app it should detect that and generate a `User` resource automatically.

### Current workarounds

Generate the resource manually.

### Approach

Update the [`install` generator](https://github.com/avo-hq/avo/blob/819cc1711979d68f7c982bffe90b1a625fd1cc4b/lib/generators/avo/install_generator.rb#L1) to also run `bin/rails generate avo:resource user`.

FLX-0x00:
Note that such resources do not always use the name User. For rodauth, it is Account by default.

adrianthedev:
That's a good point. Thanks!
Maybe we could search for rodauth and `Account` as well and incorporate that too.

PR #1737, opened 2023-05-10
feature: when installing Avo detect if User model exists and create a resource for it

marianDdev:
# Description
Fixes https://github.com/avo-hq/avo/issues/1719

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/lib/generators/avo/install_generator.rb
+++ b/lib/generators/avo/install_generator.rb
@@ -14,6 +14,17 @@ module Generators
 
         template "initializer/avo.tt", "config/initializers/avo.rb"
         directory File.join(__dir__, "templates", "locales"), "config/locales"
+        create_resources
+      end
+
+      def create_resources
+        if defined?(User).present?
+          Rails::Generators.invoke("avo:resource", ["user", "-q"], {destination_root: Rails.root })
+        end
+
+        if defined?(Account).present?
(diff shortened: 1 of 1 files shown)


lib/generators/avo/install_generator.rb
```diff
@@ -15,6 +15,10 @@ def create_initializer_file
         template "initializer/avo.tt", "config/initializers/avo.rb"
         directory File.join(__dir__, "templates", "locales"), "config/locales"
       end
+
+      def create_user_resource
+        unless "User".safe_constantize.nil? run "./bin/rails generate avo:resource user"
```
adrianthedev, line 20: This is not bad, but we have a way to invoke generators programmatically.

So instead of `run ....` we could use `Rails::Generators.invoke("avo:resource", ["user", "-q"], {destination_root: Rails.root})`.

adrianthedev: Also, instead of `"User".safe_constantize.nil?` which might raise some errors in some scenarios we could use `defined?("User").present?`.

adrianthedev (commented):
We probably need to trigger the `create_user_resource` method after the `directory` statement.

We should look for the `User` and `Account` models.

adrianthedev:
Thanks for your contributions @marianDdev! To many more!

Issue #1659, opened 2023-04-03
Use the translations in Avo as a fallback if the user does not eject them.


adrianthedev:
### Feature

It's not that nice to have to eject the translations from the start. Let's try and use the ones in Avo and if the user ejects them, make sure we use the ejected ones.

PR #1765, opened 2023-05-23
feature: use avo translations as fallback

marianDdev:
# Description
Use the translations in Avo as a fallback if the user does not eject them

Fixes https://github.com/avo-hq/avo/issues/1659

# Checklist:

## Manual review steps

1. Step 1
1. Step 2


lib/avo/engine.rb
```diff
@@ -72,6 +72,14 @@ class Engine < ::Rails::Engine
       require_relative "../generators/model_generator"
     end
 
+    LOCALES_PATH = File.join(__dir__, "config", "locales")
+
+    config.i18n.load_path += if File.directory?(LOCALES_PATH)
+                               Dir[Rails.root.join("config", "locales", "*.{rb,yml}")]
+                             else
+                               Dir[Avo::Engine.root.join("lib", "generators", "avo", "templates", "locales", "*.{rb,yml}")]
+                             end
```
adrianthedev, line 81: There are a few things here:

1. the `LOCALES_PATH` does not check fo the parent app locales dir. so we're essentially checking the wrong path.
2. I don't know if this is a good strategy, loading one locale or another. I think we should add the to rails and say something like, "if you have it in the parent app, use that; if not, fallback to ours".

I wrapper everything up in an initializer to have it run in the right point in the boot-up flow.

```suggestion
    initializer "avo.locales" do |app|
      I18n.load_path += Dir[Avo::Engine.root.join("lib", "generators", "avo", "templates", "locales", "*.{rb,yml}")]
      I18n.load_path += Dir[Rails.root.join("config", "locales", "*.{rb,yml}")]
    end
```

adrianthedev (commented):
Thank you for digging through and finding the method we need to access. I added a tiny suggestion that seems to be working as a fallback.

lib/avo/engine.rb
```diff
@@ -72,6 +72,11 @@ class Engine < ::Rails::Engine
       require_relative "../generators/model_generator"
     end
 
+    initializer "avo.locales" do |app|
+      I18n.load_path += Dir[Avo::Engine.root.join("lib", "generators", "avo", "templates", "locales", "*.{rb,yml}")]
+      I18n.load_path += Dir[Rails.root.join("config", "locales", "*.{rb,yml}")]
```
simi, line 77: Any reason why this is added? Engine should not care about `Rails.root` files at all. This breaks rubygems.org [tests](https://github.com/rubygems/rubygems.org/actions/runs/5136907592/jobs/9244269425?pr=3827).

adrianthedev: Hey @simi. Thanks for writing.

When installing Avo a bunch of locales were being copied into the parent app.

 - we try to keep the number of files we add as little as possible
 - those locales are not auto-updatable. They only update when the developer runs `rails generate avo:locales` (which rarely happens).

I can't quite understand what broke from those tests, but I'd love to help fix it. 
Do you have any ideas why the tests fail? Do we need to do some kind of a `defined?(Rails)` check?

rickychilcott: These lines seem also to be breaking `clearance` gem's I18n for me as well -- I'm not able to override clearanaces language with the latest avo versions.

simi: I'm sorry, but I do have not a much time to explore these days. But can you explain why `I18n.load_path += Dir[Rails.root.join("config", "locales", "*.{rb,yml}")]` is needed? Why do you need to add apps locales to `load_path`? I understand the line above, load Avo locales, but I don't follow this one.

adrianthedev: @simi we added that so you can add your own overrides in you're `config/locales` directory.
Otherwise, if we don't add it you have to add it yourself somewhere in `application.rb`.

Can you please try `'avo', '2.34.4.pre.1'` and see if that removes the CI crash.
I removed the `I18n.load_path += Dir[Rails.root.join("config", "locales", "*.{rb,yml}")]` statement.

adrianthedev: @simi released `2.34.4` now with that fix.

PR #1797, opened 2023-06-21
feature: format_using and update_using field options

adrianthedev:
# Description

Enables developers to add custom formatters before the data is displayed in the UI and before it gets saved to the database.

We retrofitted `format_using` to use the `ExecutionContext` and dropped the `value` argument.

The `update_using` block is run just before the value is saved on the model.

[Docs](https://docs.avohq.io/2.0/field-options.html#fields-formatter)

```ruby
field :metadata, as: :code,
  format_using: -> {
    if view == :edit
      JSON.generate(value)
    else
      value
    end
  },
  update_using: -> do
    ActiveSupport::JSON.decode(value)
  end
```

# Checklist:

### DIFF
--- a/app/components/avo/fields/area_field/edit_component.html.erb
+++ b/app/components/avo/fields/area_field/edit_component.html.erb
@@ -1,5 +1,6 @@
 <%= field_wrapper **field_wrapper_args do %>
   <%= @form.text_field @field.id,
+    value: @field.value,
     class: classes("w-full"),
     value: field.value.to_s,
     placeholder: @field.placeholder,
--- a/app/components/avo/fields/boolean_field/edit_component.html.erb
+++ b/app/components/avo/fields/boolean_field/edit_component.html.erb
@@ -1,6 +1,7 @@
 <%= field_wrapper **field_wrapper_args, dash_if_blank: false do %>
   <div class="h-8 flex items-center">
     <%= @form.check_box @field.id,
+      value: @field.value,
       checked: @field.value,
       class: "text-lg h-4 w-4 checked:bg-primary-400 focus:checked:!bg-primary-400 #{@field.get_html(:classes, view: view, element: :input)}",
       data: @field.get_html(:data, view: view, element: :input),
--- a/app/components/avo/fields/code_field/edit_component.html.erb
+++ b/app/components/avo/fields/code_field/edit_component.html.erb
@@ -1,6 +1,7 @@
 <%= field_wrapper **field_wrapper_args, full_width: true do %>
   <div data-controller="code-field">
     <%= @form.text_area @field.id,
+      value: @field.value,
       class: classes("w-full"),
       data: {
         'code-field-target': 'element',
--- a/app/components/avo/fields/country_field/edit_component.html.erb
+++ b/app/components/avo/fields/country_field/edit_component.html.erb
@@ -1,5 +1,6 @@
 <%= field_wrapper **field_wrapper_args do %>
   <%= @form.select @field.id, @field.select_options, {
+      value: @field.value,
       selected: @field.value,
       include_blank: @field.include_blank
     },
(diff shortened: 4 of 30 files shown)


koppen:
@adrianthedev Thanks for all your hard work, and please take the following as constructive criticism:

Dropping the value argument is a breaking change and it did indeed break our application in production. We've got a bunch of formatters in our application using the documented API, ie

    field :is_writer, as: :text, format_using: -> (value) { value.present? ? '👍' : '👎' }

which are apparently supposed to now be

    field :is_writer, as: :text, format_using: -> { value.present? ? '👍' : '👎' }

but we had no way of knowing that we needed to make this change. Sure, we could have verified the changes in more depth, but when the release notes state that a new feature is added (which we don't use) and links to documentation that states our way of doing things is correct, we didn't raise any red flags.

We like Avo (so much that we've paid for it and contributed previously) and I beg of you; please be more wary about breaking changes in the future, they make me look bad ;) I would much have preferred if the change wasn't breaking. Or if it had to be breaking, that it was clearly indicated in the release notes. Also, the [documentation](https://docs.avohq.io/2.0/field-options.html#fields-formatter) should probably have been updated, especially seeing how it was linked from the [release notes](https://avohq.io/releases/2.36.0).

Cheers! I'm looking forward to digging into `update_using`, though :)

adrianthedev:
Hey @koppen.

Thanks for the feedback. Yes, this is constructive 🙌
Let me

koppen:
The change is fine, and I am perfectly okay with breaking changes, I just wish it had been more obvious to me.

> We usually cover all changes in our release notes. Even this one is covered.

The release notes is what we usually run through on upgrades, mostly because that's what Dependabot surfaces for us. And indeed, the change was mentioned [there](https://github.com/avo-hq/avo/releases/tag/v2.36.0) as a new feature.

> We also provide an up-to-date [upgrade guide](https://docs.avohq.io/2.0/upgrade.html)

I did not know about this 😳 That said, it seems to stop at v2.11 and doesn't mention v.2.36?

CleanShot 2023-08-09 at 09 22 17

> Regarding the documentation, I'm not sure how that could have been out of sync. The [release](https://github.com/avo-hq/avo/releases/tag/v2.36.0) went out on July 4th and the docs were updated [a few days prior](https://github.com/avo-hq/avodocs/commit/4196a294539acee3de3b1e5462e78b7b771cf4d2).

Again, something doesn't seem right here. I do not see the updated documentation at that URL:

CleanShot 2023-08-09 at 09 23 45

So a couple of concrete suggestions that would help me in the future:

1. Any chance the upgrade guide could be linked from the release notes? I reckon it'd fit right into the "More information and release video [here](https://avohq.io/releases/2.35.1)" line, something like "Upgrade guide at https://docs.avohq.io/2.0/upgrade.html and more information and release video here"
2. Please explicitly mark breaking changes in the release notes, we're not likely to dive into an upgrade guide unless we know we need to.

adrianthedev:
Wow. That is still hapenning. You're a long-time user of Avo.

You are seeing the old docs. Please try to do a ctrl/command+shift+R on that page and clear the cookies and service worker for docs.avohq.io.
The old docs website used a service worker which was indefinitely cached on the browser. There's no "turning it off" from the server. We're open to suggestions on that.

We changed those docs about a year ago.
Sorry about that.

adrianthedev:
Your suggestions do make sense. I'll go ahead and follow them in the next releases ❤️

koppen:
Whoa! Everything looks new and fresh! It's like the wool has been pulled from my eyes 😆 

Yeah, service workers can be a problematic relationship, and indeed, there's no turning it off server-side - short of changing the URLs entirely, I guess.

adrianthedev:
Posted about it on our [Discord server](https://discord.com/channels/740892036978442260/931204374171119676/1138793077138522225) too.
We'll mention it in our next newsletter too.

PR #1801, opened 2023-06-26
🐛 Update field to nil when detaching optional field

RocKhalil:
# Description
When detaching an association that can have a nil value, we shouldn't delete the record

# Checklist:

## Manual review steps
1. Create a has many association with optional field
2. Detach it

example:
1. Player -> team_id ( optional )
2. Team -> players


adrianthedev:
Hey @RocKhalil. Thanks for submitting this. Apologies for the late reply.

So, I tested this with `optional: true` and `false`, and they both yield the same results.

Shouldn't the `optional: true` just detach it and `false` destroy it?
I mean, I see scenarios of needing both of them. Sometimes that pivot model is not needed so `optional: false` would destroy it and other times when you need it and `true` just nullifies the column.

Is that what you were thinking about too? Maybe I misunderstood.

FYI: I tested this by detaching the `memberships` from a team page.

RocKhalil:
@adrianthedev you're right, that's a bug. I guess `(relationship.options || {})[:optional]` should be changed to `(relationship.options || {})[:optional] == true` ( which is weird )

adrianthedev:
I'm still getting the same results. nomatter what I do, the id gets nullified and the record is not destroyed

RocKhalil:
hey @adrianthedev sorry I missed the notification.

if I use
```
belongs_to :team, optional: true
```
then it sets `team_id` to `nil`

otherwise,
```
belongs_to :team, optional: false
```
then it deletes the record.

adrianthedev:
Maybe I'm doing something wrong?

Paul-Bob:
Hello @RocKhalil thanks for bringing this up! 

I'm a little reticent about this approach, according to [active record's official documentation the `optional` option](https://guides.rubyonrails.org/association_basics.html#optional) enables or disables validations for "the presence of the associated object". 

When we're talking about record's destruction we should look for the `dependent` option, `dependent` will dictate the fate of the remaining object in the relationship when one of them is destroyed.

In order to better comprehend the issue we'll need a little bit more context. We need this four code snippets (or a reproduction repository) to reproduce the issue and be able to found the best solution:
- Models relationship configuration **from both models**
- Migration file regarding the relationships **from both tables**

I wasn't able to replicate the issue and we should assure that DB / Active Record configuration is on point before jumping into Avo code.

RocKhalil:
Hey @Paul-Bob , this isn't related to ActiveRecord..
The things is that avo's code `deletes` the record in all cases, optional or not.
What I'm trying to do is that `when it's optional, and we're doing a detach, let's nullify it instead of deleting it`.

so instead of `.delete` / `.destroy` we do `.update!(team_id: nil)`

Paul-Bob:
Hmm got it, anyway i can't replicate it, I'm detaching from a has_many table and none of the records are getting destroyed that's why was asking for all the configuration in order to replicate the bug

RocKhalil:
@Paul-Bob if you're using my PR, then this should be the case, it shouldn't delete it; not sure why it's deleting on @adrianthedev 's end.

Paul-Bob:
For the record, we continued some of the discussion [here](https://discord.com/channels/740892036978442260/1126460129852784651).

> The things is that avo's code `deletes` the record in all cases, optional or not.

I'm not able to replicate this behavior at all please make a little reproduction repository so we can experience the bug.

Our `AssociationsController` have this line `@model.send(association_name).delete @attachment_model` that deletes the relationship **not** the record, it's doing something like: `boathouse.workout_videos.delete video`. Both, `video` and `boathouse` still exist, none of them are deleted, however that `video` doesn't belong to that `boathouse` anymore... 

The only thing I can point wrong on this code is that we're using `delete` instead `destroy`. We should use `destroy` to assure that callbacks are called when relationship between records is broken.

Paul-Bob:
@RocKhalil i'll close this PR in favor of this one https://github.com/avo-hq/avo/pull/1856

We opted to NOT render the detach button if optional argument is not present on the `belongs_to` side of the relationship.

RocKhalil:
sorry for not working on it, still busy with a project I'm working on right now.

Paul-Bob:
No worries! Thanks for noticing this issue and for trying to fix it!

Issue #1796, opened 2023-06-21
Issue with has_many Relationship Not Loading Correctly in Show View


abdellahi-brahim:
### Steps to Reproduce
- Create 2 models, with a has_many or has_and_belongs_to_many relationship
- In the index represent the relationship with tags and in the show with the corresponding association field

### Expected behavior & Actual behavior
- Render the field

### Models and resource files
- The models told above (Dog and Person and their resources)

### System configuration
**Avo version**:
2.35

**Rails version**:
7 and 6 too

**Ruby version**:
3.1.0

**License type**:

 - [X ] Community
 - [X] Pro

**Are you using Avo monkey patches, overriding views or view components?**

 - [ ] Yes. If so, please post code samples.
 - [X] No

### Screenshots or screen recordings

![image](https://github.com/avo-hq/avo/assets/79083991/7f573065-0128-4f48-a417-f7a88e24f2fe)
![image](https://github.com/avo-hq/avo/assets/79083991/de048ae5-e1a1-4cc5-8bb0-92b3fc382119)

#### Impact

 - [ ] High impact (It makes my app un-usable.)
 - [X] Medium impact (I'm annoyed, but I'll live.)
 - [ ] Low impact (It's really a tiny thing that I could live with.)

#### Urgency

 - [ ] High urgency (I can't continue development without it.)
 - [X] Medium urgency (I found a workaround, but I'd love to have it fixed.)
 - [ ] Low urgency (It can wait. I just wanted you to know about it.)

adrianthedev:
Hey @abdellahi-brahim I think I found a bug there. Thanks for the reproduction repo. That really helped out.
I'm working on it and will have a fix next week.

Paul-Bob:
Hi @abdellahi-brahim, what's up!?

Thank you for the detailed description and repo!

As @adrianthedev mentioned, it can be considered a bug. Avo do not allow field id's duplication.

Having:
```ruby
with_options only_on: :index do
  field :id, as: :id
  field :name, as: :text
  field :dogs, as: :tags do |model, resource, view|
    model.dogs.pluck(:name)
  end
end

with_options only_on: :show do
  field :id, as: :id
  field :name, as: :text
  field :dogs, as: :has_many
end
```

it's equal to having:
```ruby
field :id, as: :id, only_on: :index
field :name, as: :text, only_on: :index
field :dogs, as: :tags, only_on: :index do |model, resource, view|
  model.dogs.pluck(:name)
end

[...]

PR #1803, opened 2023-06-26
fix: incorrect field matcher

adrianthedev:
# Description

Fixes https://github.com/avo-hq/avo/issues/1796

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/app/controllers/avo/application_controller.rb
+++ b/app/controllers/avo/application_controller.rb
@@ -223,7 +223,7 @@ module Avo
 
     def related_resource
       # Find the field from the parent resource
-      field = @resource.get_field params[:related_name]
+      field = @resource.get_field params[:related_name], for_view: :show
 
       return field.use_resource if field&.use_resource.present?
 
--- a/app/controllers/avo/associations_controller.rb
+++ b/app/controllers/avo/associations_controller.rb
@@ -109,7 +109,11 @@ module Avo
     end
 
     def set_attachment_resource
-      @attachment_resource = @field.use_resource || (App.get_resource_by_model_name @attachment_class)
+      @attachment_resource = if @field.present? && @field.respond_to?(:use_resource) && @field.use_resource.present?
+        @field.use_resource
+      else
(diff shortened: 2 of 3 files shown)


app/controllers/avo/associations_controller.rb
```diff
@@ -109,7 +109,11 @@ def set_attachment_class
     end
 
     def set_attachment_resource
-      @attachment_resource = @field.use_resource || (App.get_resource_by_model_name @attachment_class)
+      @attachment_resource = if @field.present? && @field.respond_to?(:use_resource) && @field.use_resource.present?
+        @field.use_resource
+      else
+        App.get_resource_by_model_name @attachment_class
+      end
```
Paul-Bob, line 116: I think all associations fields respond to `use_resource` but it can return `nil`

This validations of `@field.present? && @field.respond_to?(:use_resource) && @field.use_resource.present?` are silencing some errors that maybe we don't want to silence. 

Is there any moment that we want `associations_controller` called without a `@field`?!

adrianthedev: > Is there any moment that we want associations_controller called without a @field?!

Good question! I guess not.

This is a good place to raise an error with a proper message. "We couldn't find the `X` association." or "We couldn't find a resource for `X` association. Are you sure you created one?"

So the users understand what happened.

app/controllers/avo/application_controller.rb
```diff
@@ -223,7 +223,7 @@ def resource
 
     def related_resource
       # Find the field from the parent resource
-      field = @resource.get_field params[:related_name]
+      field = @resource.get_field params[:related_name], for_view: :show
```
Paul-Bob, line 226: We should not hard-code the view, it can also be `:edit`

adrianthedev: Makes sense. I'll follow your lead.

adrianthedev:
Closing for now and fixing in Avo 3

PR #1818, opened 2023-06-30
feature: Allow custom destroy methods.

davekruse:
# Description
This is a proposed feature to add custom destroy methods to resources. This makes Avo compatible with soft delete gems like discard. Please reject if not appropriate. 

```
# widget_resource.rb
...
  self.destroy_record_method = -> (record:) {
    record.discard
  }
...
```

If not defined, the default destroy! is used. 

# Checklist:

### DIFF
--- a/app/controllers/avo/base_controller.rb
+++ b/app/controllers/avo/base_controller.rb
@@ -215,21 +215,21 @@ module Avo
 
     def save_model
       perform_action_and_record_errors do
-        save_record_action
+        save_model_action
       end
     end
 
-    def save_record_action
+    def save_model_action
       @model.save!
     end
 
     def destroy_model
       perform_action_and_record_errors do
-        destroy_record_action
+        destroy_model_action
       end
     end
 
-    def destroy_record_action
+    def destroy_model_action
       @model.destroy!
     end
 
--- a/app/controllers/avo/base_controller.rb
+++ b/app/controllers/avo/base_controller.rb
@@ -215,8 +215,12 @@ module Avo
 
     def save_model
       perform_action_and_record_errors do
-        @model.save!
+        save_record_action
       end
+    end
+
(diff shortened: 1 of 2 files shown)


adrianthedev:
Could this be achieved by overriding the `destroy_model` method in a specific controller?

davekruse:
@adrianthedev yes, perfect. 

Now, customizing the delete method is as simple as:

```
class UsersController
  ...
  def destroy_record_action
    @model.discard
  end
end
```

Is that what you had in mind?

davekruse:
What's the preferred naming convention in this case where we are straddling naming conventions for the model and the resource? Within `destroy_model` and `save_model`, should it be:

- `destroy_record_action` and `save_record_action`
- `destroy_model_action` and `save_model_action`

adrianthedev:
So, in the begining we used `model` everywhere, but that's not the best name for that. A better name would be `record` (pundit uses that and other gems too).
In Avo 3 we migrated to `record`, but in Avo 2 we still have a few places where we use `model`.

Let's use `model` here too as others might have overridden that method in their controller.

davekruse:
Since the existing `destroy_model` and `save_model` are remaining the same, they are backwards compatible - current overrides won't be broken by this change.

For the new methods, we shouldn't have conflict with existing implementations. But for forward compatibility looking ahead to v3, would it make sense to keep 'resource' in the names so that overrides of the method could carry forward without change?

Or do you prefer a consistent use of 'model' throughout v2? 

Lemme know and I'll make it so.

adrianthedev:
Yes. I would prefer to use model in v2.

`save_model` -> `save_model_action`

adrianthedev:
Thank you for the contribution. I'll cut a release in a few hours as I have another PR almost ready.

PR #1824, opened 2023-07-04
improvement: apply policy scope on has many search query

### LINKED ISSUE #1808: searchable option partially bypasses the scope in has_many association
### Describe the bug
Considering a model `Account` and a model `Role`. An account has many roles and a role has and belongs to many accounts. 

Simple resource of the accounts
```ruby
class AccountResource < Avo::BaseResource
...
field :roles, as: :has_many, searchable: true
```

And simple scope / policy of roles that uses the `account_editor` role to scope only roles that starts with "user_". So an account with the `account_editor` role attached is not able to see his own role.

```ruby
  class Scope < Scope
    def resolve
      if user.has_any_role? :admin
        scope.all
      elsif user.has_role? :editor
        scope.where("name LIKE ?", "user_%")
      else
        scope.none
      end
    end
  end
```
This works as expected

image

But clicking on the search bar will reveal the `account_editor` role that is attached to the user even if this is not in scope.

image

I would like to ask if this is normal behavior. Maybe I am missing the concept and not seeing the difference between the search and the "index" view of the has_many association.

Paul-Bob:
# Description

Fixes #1808

### Scope example
```ruby
class UserPolicy < ApplicationPolicy

  # . . .

  # . . .

  class Scope < Scope
    def resolve
      scope.where("first_name like ?", "%a%")
    end
  end
end
```

### Before
When searching on a has many field, the policy scope was not being applied, returning all association records.

[before.webm](https://github.com/avo-hq/avo/assets/69730720/f6b6045c-793e-4804-912b-76f13c9257b1)

### After
Now the policy scope is being applied on the search result.

[after.webm](https://github.com/avo-hq/avo/assets/69730720/709c916b-b066-42fd-afb7-b18f8c687c84)

# Checklist:

### DIFF
--- a/app/controllers/avo/search_controller.rb
+++ b/app/controllers/avo/search_controller.rb
@@ -112,12 +112,12 @@ module Avo
     # This scope is applied if the search is being performed on a has_many association
     def apply_has_many_scope
       association_name = BaseResource.valid_association_name(parent, params[:via_association_id])
+
+      # Get association records
       scope = parent.send(association_name)
 
-      begin
-        scope = policy_scope scope
-      rescue Pundit::NotDefinedError
-      end
+      # Apply policy scope if authorization is present
+      scope = resource.authorization&.apply_policy scope
 
       Avo::Hosts::SearchScopeHost.new(block: @resource.search_query, params: params, scope: scope).handle
     end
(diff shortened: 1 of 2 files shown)


adrianthedev:
Can we also add some tests?

PR #1826, opened 2023-07-04
fix: tab visibility

### LINKED ISSUE #1812: resources inside tabs are not checked against the index? policy
### Context
Consider a has_may relationship, for example, Accounts has many roles and you want to display this in the Account resource.

```ruby
field :roles, as: :has_many
```
Before a simple panel will be visible, avo checks the `index?` policy for roles if you are using pundit or something else. If `index?` is `false`, the panel will not be rendered.

In tab view the pill is rendered and results in a `content missing` message or show loading screen if it is the first item. It would be nice if a resource inside a tab would also be checked against the `index?` policy and the corresponding tab would not be visible.

```ruby
  tabs style: :pills do
    field :roles, as: :has_many
  end
```
```ruby
  def index?
    false
  end
end
```
image

Paul-Bob:
# Description

Fixes #1812

Tab header was wrongly showing buttons to not visible tabs

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/app/components/avo/tab_switcher_component.rb
+++ b/app/components/avo/tab_switcher_component.rb
@@ -81,7 +81,7 @@ class Avo::TabSwitcherComponent < Avo::BaseComponent
     return true if !item.is_field?
     return true if !item.has_own_panel?
 
-    return false unless item.visible_on?(view)
+    return false if !item.visible_on?(view)
 
     # If item is hydrated with the correct resource and is not authorized, it's not visible
     return false if item.resource.present? && !item.authorized?
--- a/app/components/avo/tab_switcher_component.rb
+++ b/app/components/avo/tab_switcher_component.rb
@@ -88,6 +88,10 @@ class Avo::TabSwitcherComponent < Avo::BaseComponent
         visible = item.authorized?
       end
 
+      if item.items.count == 1 && first_item.is_field? && first_item.has_own_panel? && first_item.respond_to?(:authorized?) && first_item.resource.present?
+        visible = first_item.authorized?
+      end
+
       visible
     end
   end
--- a/app/components/avo/tab_switcher_component.rb
+++ b/app/components/avo/tab_switcher_component.rb
@@ -66,12 +66,9 @@ class Avo::TabSwitcherComponent < Avo::BaseComponent
     tabs.select do |tab|
       next false if tab.items.blank?
       next false if tab.is_field? && !tab.authorized?
-
-      if tab.items.count == 1
-        next false if !single_item_visible?(tab.items.first)
-      end
-
-      next false if !tab.visible? || !tab.visible_on?(view)
+      next false if tab.has_a_single_item? && !single_item_visible?(tab.items.first)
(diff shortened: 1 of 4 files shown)


lib/avo/fields/field_extensions/visible_in_different_views.rb
```diff
@@ -18,7 +18,7 @@ def initialize(id = nil, **args, &block)
         def visible_on?(view)
           raise "No view specified on visibility check." if view.blank?
 
-          send :"show_on_#{view.to_s}"
+          send :"show_on_#{view}"
```
Paul-Bob, line 21: unrelated fix

app/components/avo/tab_switcher_component.rb
```diff
@@ -63,36 +63,31 @@ def selected?(tab)
...
-    tabs.select do |item|
-      visible = true
+    tabs.select do |tab|
+      next false if tab.items.blank?
+      next false if tab.is_field? && !tab.authorized?
 
-      if item.items.blank?
-        visible = false
+      if tab.items.count == 1
+        next false if !single_item_visible?(tab.items.first)
```
adrianthedev, line 71: can't these be combined into one line?

`tab.items.count == 1 && !single_item_visible?(tab.items.first)`

app/components/avo/tab_switcher_component.rb
```diff
@@ -63,36 +63,31 @@ def selected?(tab)
...
       end
 
-      first_item = item.items.first
-      if item.items.count == 1 && first_item.is_field? && first_item.has_own_panel? && !first_item.visible_on?(view)
-        # Return nil if tab contians a has_many type of fields and it's hidden in current view
-        visible = false
-      end
+      next false if !tab.visible? || !tab.visible_on?(view)
```
adrianthedev, line 74: let's break these on two lines as they are two separate logical statements.

app/components/avo/tab_switcher_component.rb
adrianthedev: maybe we coudl add a `has_a_single_item?` method on the tab and make it more readable here?

app/components/avo/tab_switcher_component.rb
```diff
@@ -63,36 +63,31 @@ def selected?(tab)
...
-        visible = item.visible?
-      end
+  private
 
-      if item.respond_to?(:authorized?)
-        visible = item.authorized?
-      end
+  def single_item_visible?(item)
+    # Item is visible if is not a field or don't have its own panel
+    return true if !item.is_field? || !item.has_own_panel?
```
adrianthedev, line 84: same here break on two lines?

adrianthedev (approved):
It looks good.
Should we add a test for that?
If the manual tests pass, feel free to merge it.

Paul-Bob:
> It looks good. Should we add a test for that? If the manual tests pass, feel free to merge it.

Yes, we should have a test because this method is kinda complex and sensible, I'll implement a simple one.

PR #1832, opened 2023-07-05
fix: allow nil values on grid cover

### LINKED ISSUE #1816: Enable the file field to take `nil` values and not crash
### Feature

Please check out https://github.com/avo-hq/avo/issues/1813#issuecomment-1613172532

Paul-Bob:
# Description

Fixes #1816

When cover type is `file` check if value is present before calling `attached?` on it. Before this PR having a `nil` value on grid cover would break the app, now it renders the `Avo::Index::GridCoverEmptyStateComponent`

# Checklist:

## Manual review steps

1. Step 1
1. Step 2


adrianthedev:
can we ensure that this is happening on `file` and `files` field too please?

Paul-Bob:
> can we ensure that this is happening on `file` and `files` field too please?

Both `file` and `files` was accepting `nil` on all pages. The issue was only on `grid do` block the `cover` was breaking if nil value present. We don't accept `cover :id, as: :files` although can be an interesting feature to have an carousel of images on the grid cover configuration.

adrianthedev (approved):

PR #1833, opened 2023-07-05
fix: action visibility on show controls

### LINKED ISSUE #1769: The action's `visible` block is not respected in the `action` show control
### Describe the bug

If an action has `self.visible = -> {false}`, the action is still displayed on the page if the `show_controls` uses it `action TheAction`

Paul-Bob:
# Description

Fixes #1769

Action visibility block runs even on show controls actions now

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/lib/avo/resources/controls/action.rb
+++ b/lib/avo/resources/controls/action.rb
@@ -4,17 +4,17 @@ module Avo
       class Action < BaseControl
         attr_reader :klass
 
-        def initialize(klass, model: nil, resource: nil, view: nil, **args)
+        def initialize(klass, record: nil, resource: nil, view: nil, **args)
           super(**args)
 
           @klass = klass
           @resource = resource
-          @model = model
+          @record = record
           @view = view
         end
 
         def action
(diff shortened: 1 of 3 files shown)


lib/avo/resources/controls/action.rb
```diff
@@ -4,17 +4,17 @@ module Controls
...
           @klass = klass
           @resource = resource
-          @model = model
+          @record = record
           @view = view
         end
 
         def action
-          @instance ||= @klass.new(model: @model, resource: @resource, view: @view)
+          @instance ||= @klass.new(model: @record, resource: @resource, view: @view)
```
Paul-Bob, line 17: For some reason this class is initialized by `Avo::Resources::Controls::Action.new(klass, record: record, resource: resource, view: view, **args)` but the `initialize` method was not expecting the same argument name, using `model` instead `record`. (In avo3 this is already uniformed)

adrianthedev:
Can we add some tests please?

adrianthedev (approved):

PR #1835, opened 2023-07-07
Added zoom to location fields

iainbeeston:
# Description

I've added an extra attribute to the location field so you can override the zoom level in the resource. I've also made mapkick show the map controls so users can zoom in and out if they want to.

Fixes # (issue)

Right now location fields always load at the default zoom level of 15 and there's no way to customize it.

# Checklist:

## Manual review steps

1. Find a resource with a location field in the code
2. View a model that uses that resource and see the zoom level of the location (should be very zoomed-in)
1. Edit the code so that the field has `zoom: 1`  on the location field
3. Reload the page
4. Check that the location field is now zoomed out really far

### DIFF
--- a/lib/avo/fields/location_field.rb
+++ b/lib/avo/fields/location_field.rb
@@ -3,13 +3,14 @@
 module Avo
   module Fields
     class LocationField < BaseField
-      attr_reader :stored_as
+      attr_reader :stored_as, :zoom
 
       def initialize(id, **args, &block)
         hide_on :index
         super(id, **args, &block)
 
         @stored_as = args[:stored_as].present? ? args[:stored_as] : nil # You can pass it an array of db columns [:latitude, :longitude]
+        @zoom = args[:zoom].present? ? args[:zoom].to_i : 15
       end
 
       def value_as_array?
--- a/app/components/avo/fields/location_field/show_component.html.erb
+++ b/app/components/avo/fields/location_field/show_component.html.erb
@@ -1,6 +1,6 @@
 <%= field_wrapper **field_wrapper_args do %>
   <% if field.value_present? %>
-    <%= js_map [{latitude: field.value[0], longitude: field.value[1]}], id: "location-map" %>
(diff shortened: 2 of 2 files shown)


Paul-Bob (approved):

iainbeeston:
I've added documentation for this in https://github.com/avo-hq/avodocs/pull/66

lib/avo/fields/location_field.rb
```diff
@@ -3,13 +3,14 @@
...
     class LocationField < BaseField
-      attr_reader :stored_as
+      attr_reader :stored_as, :zoom
 
       def initialize(id, **args, &block)
         hide_on :index
         super(id, **args, &block)
 
         @stored_as = args[:stored_as].present? ? args[:stored_as] : nil # You can pass it an array of db columns [:latitude, :longitude]
+        @zoom = args[:zoom].present? ? args[:zoom].to_i : 15
```
Paul-Bob, line 13: I just saw on the docs that `@zoom` accept values between [0, 22] I think that would be interesting to raise an informative error about the accepted range if for some reason the value goes out of this range. @iainbeeston and @adrianthedev what do you think?

Paul-Bob:
Thank you for the amazing work @iainbeeston!

lib/avo/fields/location_field.rb
adrianthedev: My first thought was "let's just take the argument from the user and if it's greater than 22, then we force it as 22, but that would generate some technical support as "why isn't Avo working and not letting me add a number greater than 22".
The second thought I had was, to raise an error when that number is greater than 22. But what if the API changes at some point? We'd have to maintain that parity between our validation and the external API.
Now, I'm thinking of just leaving it without validation and directing the user towards the original library ([mapkick](https://chartkick.com/mapkick)) because of those two reasons.

So my vote goes to leaving it like that and instructing the user that we only forward the options to the original library.

adrianthedev:
Climbing the ladder @iainbeeston 
https://avo-hq.github.io/contributors/

PR #1836, opened 2023-07-07
fix: visible items

### LINKED ISSUE #1812: resources inside tabs are not checked against the index? policy

Paul-Bob:
# Description

Fixes #1812

When a tab is a `has_many_field` for example and is **not** visible by policy now is **not** rendered.

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/lib/avo/concerns/visible_items.rb
+++ b/lib/avo/concerns/visible_items.rb
@@ -28,6 +28,8 @@ module Avo
 
       def visible(item)
         if item.is_field?
+          return false if item.respond_to?(:authorized?) && item.resource.present? && !item.authorized?
+
           item.visible? && item.visible_on?(view)
         else
           item.visible?


adrianthedev:
LGTM

PR #1856, opened 2023-07-19
fix: don't render detach button when can't detach record

Paul-Bob:
# Description

Fixes # (issue)

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/app/components/avo/index/resource_controls_component.rb
+++ b/app/components/avo/index/resource_controls_component.rb
@@ -10,7 +10,7 @@ class Avo::Index::ResourceControlsComponent < Avo::ResourceComponent
   end
 
   def can_detach?
-    is_has_many_association ? super : false
+    is_has_many_association? ? super : false
   end
 
   def can_edit?
@@ -76,7 +76,7 @@ class Avo::Index::ResourceControlsComponent < Avo::ResourceComponent
     ::Avo::App.get_resource_by_model_name @parent_model.class
   end
 
-  def is_has_many_association
+  def is_has_many_association?
     @reflection.is_a?(::ActiveRecord::Reflection::HasManyReflection) || @reflection.is_a?(::ActiveRecord::Reflection::ThroughReflection)
   end
 
--- a/app/components/avo/index/resource_controls_component.rb
+++ b/app/components/avo/index/resource_controls_component.rb
@@ -10,10 +10,7 @@ class Avo::Index::ResourceControlsComponent < Avo::ResourceComponent
   end
 
   def can_detach?
-    @reflection.present? &&
-      @resource.model.present? &&
-      is_has_many_association &&
-      authorize_association_for(:detach)
+    is_has_many_association ? super : false
   end
 
   def can_edit?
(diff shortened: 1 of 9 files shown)


app/controllers/avo/associations_controller.rb
```diff
@@ -81,7 +81,7 @@ def destroy
       association_name = BaseResource.valid_association_name(@model, params[:related_name])
 
       if reflection_class == "HasManyReflection"
-        @model.send(association_name).delete @attachment_model
+        @model.send(association_name).destroy @attachment_model
```
Paul-Bob, line 84: Make sure that validations / callbacks are called

app/components/avo/index/resource_controls_component.rb
```diff
@@ -10,10 +10,7 @@ def initialize(resource: nil, reflection: nil, parent_model: nil, parent_resourc
   end
 
   def can_detach?
-    @reflection.present? &&
-      @resource.model.present? &&
-      is_has_many_association &&
-      authorize_association_for(:detach)
+    is_has_many_association ? super : false
```
adrianthedev, line 13: should we add a question mark at the end?
```suggestion
    is_has_many_association? ? super : false
```

app/components/avo/resource_component.rb
```diff
@@ -20,7 +20,14 @@ def can_delete?
   end
 
   def can_detach?
-    authorize_association_for(:detach)
+    return false if @reflection.blank? || @resource.model.blank? || !authorize_association_for(:detach)
+
+    # If the inverse_of is a belongs_to, we need to check if it's optional in order to know if we can detach it.
+    if @reflection.inverse_of.is_a?(ActiveRecord::Reflection::BelongsToReflection)
```
adrianthedev, line 26: What if it does not have an `inverse_of` configured on the association? Will this fail?
Can we raise a helpful error before, so the user knows about declaring the inverse_of?

Paul-Bob: From my manual tests even if `inverse_of` is not configured rails can find it, not sure if there are cases where this can fail

app/components/avo/index/resource_controls_component.rb
Paul-Bob: I'm divided. With the `?` on the end it is more clear and readable on a `if` condition: `if is_has_many_association?`... However if we use that method on a ternary condition it's not that readable because we end up with duplicated question mark. I can agree with both notations but in this case where we use ternary I prefer without the extra question mark

app/components/avo/resource_component.rb
Paul-Bob: > The automatic guessing of the inverse association uses a heuristic based on the name of the class, so it may not work for all associations, especially the ones with non-standard names.

From https://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_many

adrianthedev:
We should update @jcn from Discord that we made this change and will be available in the next release.

PR #1869, opened 2023-07-27
Made it possible to show a text field as a link

iainbeeston:
# Description

Since the protocol option was added to text fields it's been possible to make text fields appear as a link on show and index pages, but there's no equivalent for http or https links. You can't use the protocol option for these because when rendering text with a protocol avo prepends the protocol to the link, but this doesn't make sense for urls, because they already include the protocol.

I've added a `url` option to text fields which displays them as a link without needing a protocol.

Fixes # (issue)

# Checklist:

## Manual review steps

1. Add a text field without the `url` option set
2. Enter a link (eg. https://github.com) into the field
3. Save
4. View the resource
5. The text field should not be clickable
6. Update the resource definition to include `url: true` as an option
7. View the same resource again
8. The text field should now be a clickable link

### DIFF
--- a/app/components/avo/fields/text_field/index_component.html.erb
+++ b/app/components/avo/fields/text_field/index_component.html.erb
@@ -1,6 +1,8 @@
 <%= index_field_wrapper **field_wrapper_args do %>
   <% if @field.as_html %>
     <%== @field.value %>
+  <% elsif @field.url.present? %>
+    <%= link_to @field.value, @field.value, target: "_blank" %>
   <% elsif @field.protocol.present? %>
     <%= link_to @field.value, "#{@field.protocol}:#{@field.value}" %>
   <% else %>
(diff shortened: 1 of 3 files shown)


Paul-Bob:
Thank you for submitting this PR @iainbeeston!

We already have documented 2 different ways of creating an link: https://docs.avohq.io/2.0/recipes/custom-link-field.html

When talking about links we can want to pass some extra arguments. I'm not sure if this `url` single boolean option is the right path to follow.

@adrianthedev what do you think?

adrianthedev:
Hey @iainbeeston. I see how it should be easier to format a field as a link.

Let us launch Avo 3 next week and get back to this with a bit more free time.

adrianthedev:
Hey @iainbeeston.

Thanks for sending this in. Sorry for the long wait; we are just now starting to settle in after the Avo 3 stable release.

We won't merge this for the reasons @Paul-Bob outlined above (there are ways of doing this currently) and for the fact that others might want to add more things to links (data and custom attributes) and that won't be possible with this approach, but will be with the other approaches.

The functionality of this PR can be implemented as a custom field or as an extension in your own app if you need it to be just like that.

Thank you for the contribution.

PR #1872, opened 2023-07-29
Fix target_resource for direct polymorphic relationships

saxxi:
# Description
If a `MyModel` references eg. `ModelWithInheritance` directly (eg. `myModel.my_model_with_inheritance_id`) instead of classic polymorphic relation (`…_id & …_type` fields) the code breaks with an exception.

The root cause seems to be the the provided change.

Example:

- `class Agency < Organization`
- `class MyModel < ApplicationRecord; belongs_to :agency`
- If MyModel has a `agency_id` the code returns nil

---

Screenshot 2023-07-29 at 15 42 34
Screenshot 2023-07-29 at 15 42 47
Screenshot 2023-07-29 at 15 43 40

### DIFF
--- a/lib/avo/fields/belongs_to_field.rb
+++ b/lib/avo/fields/belongs_to_field.rb
@@ -245,8 +245,9 @@ module Avo
 
         reflection_key = polymorphic_as || id
 
-        if @model._reflections[reflection_key.to_s].klass.present?
-          App.get_resource_by_model_name @model._reflections[reflection_key.to_s].klass.to_s
+        klass = @model._reflections[reflection_key.to_s].klass
+        if klass.present?
+          App.get_resource_by_model_name(klass.to_s) || App.get_resource_by_name(klass.name)
         elsif @model._reflections[reflection_key.to_s].options[:class_name].present?
           App.get_resource_by_model_name @model._reflections[reflection_key.to_s].options[:class_name]
         else


Paul-Bob:
Hi @saxxi thanks for submitting an possible resolution for this issue.

I'm curious why `App.get_resource_by_model_name` can't find the resource for the `Agency` model.

Can you share more about the configuration, like the resource used for the `Agency` model?

saxxi:
Hi @Paul-Bob thanks for looking into this.

Sure, here it is how I use it:

```ruby
# == Schema Information
#
# Table name: organizations
#
#  id               :uuid             not null, primary key
#
# Indexes
#
#  index_organizations_on_slug  (slug) UNIQUE
#
class Agency < Organization
end
```

A reference to this system

```ruby
# == Schema Information
#
# Table name: users
#
# ...
#
# Indexes
#
#  index_users_on_agency_id                       (agency_id)
#
# Foreign Keys
#
#  fk_rails_...  (agency_id => organizations.id)
#
class User < ApplicationRecord
end
```

Why did I do it this way? 😅 

Well when I started I had the agency model which served ,
by the time I needed to generalise this into "organization" I had too much to refactor.

Here imo some (small) added benefit:
- is it's possible to reference directly `agency_id`
- it's still db enforced

Paul-Bob:
And what about the Avo resources? Do you have an `AgencyResource` and `OrganizationResource`?

adrianthedev:
Hey @saxxi.
From this information, nothing seems wrong, and the associations should work fine.
Can you please set up a demo repository that we can clone and test for ourselves?

The faster way to do that is by running these commands:

```
rails new avo_demo_repo
cd avo_demo_repo
bin/rails app:template LOCATION='https://avohq.io/app-template'
```

Or you can start from [this repo](https://github.com/avo-hq/avo-reproduction).

saxxi:
hi @adrianthedev here it is, confirmed also in isolation, thanks!

https://github.com/saxxi/avo_demo_fix_target_resource_poly

adrianthedev:
Cool! Thanks! @Paul-Bob can you have a look please?

Paul-Bob:
Hi @saxxi I cloned your repo and trying to simulate the issue but i think you missed some configurations, there are models without tables and without inheritance.

Ex: `class Agency < ApplicationRecord` and schema don't have the `agencies` table

saxxi:
hi @Paul-Bob thanks again a lot for looking into this.

⚠️ There is NO agencies table.

---

As from here, `agency_id` is referencing organizations table:

Screenshot 2023-08-03 at 02 14 39
Screenshot 2023-08-03 at 02 19 50
Screenshot 2023-08-03 at 02 22 44

---

I've updated the repo so it has [annotate](https://github.com/ctran/annotate_models) gem.

Thanks again

Paul-Bob:
I've been exploring your use case and I noticed that the `AgencyResource` `model_class` is `Organization`. That happens because STI can confuse Avo if resource is not configured properly. 

Check this part of our documentation for more details https://docs.avohq.io/2.0/associations.html#single-table-inheritance-sti

This PR do not fix the problem just avoid the error to raise, in the end the  `AgencyResource` `model_class` remain `Organization`, which is wrong.

In order to fix the issue you should explicitly configure the `model_class` for your `AgencyResource`

```ruby
class AgencyResource < Avo::BaseResource
  self.title = :id
  self.includes = []
  self.model_class = Agency # <-------------------------------------------------------------
  # self.search_query = -> do
  #   scope.ransack(id_eq: params[:q], m: "or").result(distinct: false)
  # end

  field :id, as: :id
  # Fields generated from the model
  field :name, as: :text
  # add fields here
end
```
I will close this PR, but if you have any further questions or if the provided solution doesn't work for your case, feel free to comment or reopen it.

saxxi:
Thanks @Paul-Bob it's working now

PR #1875, opened 2023-07-31
feature: custom parent controller

### LINKED ISSUE #1822: Generate an `Avo::BaseResourcesController` from which all controllers inherit
We'd like to generate by default a base resources controller so users can override methods globally for all resources.

- It should not be named `ResourcesController` because there was a case where one user had the `Resource` model in their app.
- the resource generator should check for this base resource and if it doesn't exist it should create it

Related ticket https://github.com/avo-hq/avo/issues/1758.

Paul-Bob:
# Description

Fixes #1822 

We have implemented the option to customize the parent controller from which newly generated controllers will inherit. This customization can be achieved through two different approaches.

## Initializer setting
Configure the `parent_controller` option on the `avo.rb` initializer

```ruby
Avo.configure do |config|
  # ...
  config.parent_controller = "Avo::MyCustomController" # "Avo::ResourcesController" is default value
  # ...
end
```

## `--parent-controller` option
When generating a resource pass the `--parent-controller` option:
`rails g avo:resource city --parent-controller Avo::MyCustomController`

**The `--parent-controller` option takes precedence when fetched. If left blank, the configuration `parent_controller` will be accessed as a fallback.**

# Checklist:

  - https://github.com/avo-hq/avodocs/pull/73

### DIFF
--- a/lib/avo/configuration.rb
+++ b/lib/avo/configuration.rb
@@ -44,6 +44,7 @@ module Avo
     attr_accessor :sign_out_path_name
     attr_accessor :resources
     attr_accessor :prefix_path
+    attr_accessor :parent_controller
 
     def initialize
       @root_path = "/avo"
@@ -96,6 +97,7 @@ module Avo
       @field_wrapper_layout = :inline
       @resources = nil
       @prefix_path = nil
+      @parent_controller = "Avo::ResourcesController"
     end
 
     def current_user_method(&block)
--- a/lib/generators/avo/concerns/parent_controller.rb
+++ b/lib/generators/avo/concerns/parent_controller.rb
@@ -1,0 +1,20 @@
+module Generators
+  module Avo
+    module Concerns
+      module ParentController
+        extend ActiveSupport::Concern
+
+        included do
+          class_option "parent-controller",
+            desc: "The name of the parent controller.",
+            type: :string,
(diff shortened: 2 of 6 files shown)


Paul-Bob:
@adrianthedev I'll do a PR with docs after this PR is approved.

lib/avo/configuration.rb
```diff
@@ -44,6 +44,7 @@ class Configuration
     attr_accessor :sign_out_path_name
     attr_accessor :resources
     attr_accessor :prefix_path
+    attr_accessor :parent_controller
```
adrianthedev, line 47: Maybe this should be `parent_resource_controller`? Who knows what kind of other controllers we might introduce in the future.

adrianthedev (approved):
Looks good! GG!

lib/avo/configuration.rb
Paul-Bob: We have an controller generator not an resource_controller generator, `rails g avo:controller city --parent-controller SomeRandomController`.

I agree with your suggestion but should we change the above generator to: `rails g avo:resource_controller city --parent-resource-controller SomeRandomController` ?

adrianthedev: `resource_parent_controller`

PR #1876, opened 2023-07-31
Create a `belongs_to` record from the associated record screen

### LINKED ISSUE #1556: Create a `belongs_to` record from the associated record screen
### Feature

In a `Post` `has_many` `Comment`s example, when on a Comment new page, next to the posts dropdown you'd have a plus icon which when clicked it will open a modal with the post form ready to create a new post.

From [here](https://discord.com/channels/740892036978442260/740893011994738751/1059859496832344225)

## Aproach

The way I would go about it is like this:

1. update the belongs_to edit field to add that, plus link
1. it links to the create screen with a special `via_` attribute
1. that lets the controller know to use a different view from the `base/new.html.erb`. it will use a new view that is wrapped inside a modal. it will also be wrapped in a turbo-frame element similar to how actions work
2. that new modal new view would actually load up the `Avo::Views::ResourceEditComponent` in a large modal window.
3. The modal window will have the submit button that will save the form's record.
4. after the update action goes through, we need a way to reload the `belongs_to` dropdown field. we can't do a full page reload because the user might have pre-filled other fields and they won't appreciate loosing those values

## Other considerations

1. a nice to have would be to have that new record selected on the field reload
2. this should work nicely with the searchable belongs_to field
3. the modal should ensure that it displays validation errors if they may incur
4. ideally it will not use any custom JS and work on native Turbo methods
5. it should not worsen the experience for the create action and belongs_to field

### Screenshots

A rough sketch of how envision it to look

CleanShot 2023-01-28 at 12 39 30

sdcoffey:
# Description
Allows creation of a `belongs_to` record from the new or edit pages of a resource. 

Accomplished this by:
1. Adding a link to `Create new x` below the belongs_to field. This link points to the associated resource's `new` view with a special url param `:via_belongs_to_resource_class`.
2. Modifying `BaseController#new` to render a turbo stream adding a modal to the page containing the associated resource's `new` view, when the above param is present.
3. Modifying `BaseController#create_success_action` to, when the above param is present, render two turbo streams, one removing the modal, and the other containing a custom event with the details of the new record.
4. Adding a new stimulus controller to handle the custom event, updating the field with the params from the newly created record. Aside: although this introduces some bespoke js, it turned out to be way less intrusive than some other, non client-side approaches I tried.

Fixes #1556 

# Checklist:

## Manual review steps

1. Create a `Course` from `edit`/`new` of `Course::Link` -  searchable belongs_to association
5. Create a `User` from `edit`/`new` of `Fish` - belongs_to with normal HTML select
6. Create a `Post` or `Project` from `new`/`edit` of `Comment` - polymorphic belongs_to
7. Create a `Fish` from `new`/`edit` of `Review` - searchable, polymorphic belongs_to

# Open Questions
1. Should there be a field-level option to enable/disable this behavior?
2. How should we handle cases where the resource pointed to is an abbreviated version of the associated resource (e.g. `CompactUser`)?


adrianthedev:
@Paul-Bob can you please also have a look over this when you get a chance? Thank you.

Paul-Bob:
Hi @sdcoffey, first of all thank you for the amazing PR. 

I've been testing it and came across a few issues.  The modal that opens for the belongs_to creation seems a little unpolished, for example, when creating an user or some record that have a big resource it expands to full height, maybe we should assure some margins.

My main concern revolves around the form errors. They no longer seem to show up when submitting a form and some validation fails. It happens due turbo-stream I think. We should focus on solving this error rendering issue first

Paul-Bob (changes requested):
Let's assure that errors render on the form when validations fails

app/views/avo/partials/_keep_modal_open.turbo_stream.erb
```diff
@@ -0,0 +1,3 @@
+<%= turbo_stream.append "alerts" do %>
+  <%= render Avo::FlashAlertsComponent.new flashes: flash %>
+<% end %>
```
Paul-Bob, line 3: Maybe we should rename this stream to `.../_flash_alerts.turbo_stream.erb` i think it fits better to the file porpoise.

sdcoffey:
Thanks for the review @Paul-Bob! Will take another look at the issue with large resources and fix the missing validation errors (totally missed that). 

Will ping you again when I've finished updating

sdcoffey:
Hey @Paul-Bob,

Just had a chance to fix up this PR. Here's what's new:

1. Fixed the missing validation errors by ensuring that the div wrapping the edit panel has an id, and then changing the failure behavior to issue three turbo streams: 
    1. Replace the edit panel with one containing the form errors (works in the "bare" view as well as inside the modal)
    1. Append the flash to `#alerts`
    1. Append an error message for each model error to `#alerts`, mimicking the behavior in the root layout
1. Fixed the modal UI issues by forcing a one-column layout in the modal. Could have also used container media queries to solve this but the browser support isn't quite there yet IMO
1. Renamed `_keep_modal_open` to `_flash_alerts` as discussed     

Take another look and let me know if there's anything else

app/components/avo/fields/belongs_to_field/edit_component.rb
```diff
@@ -57,6 +57,17 @@ def field_html_action
     @field.get_html(:data, view: view, element: :input).fetch(:action, nil)
   end
 
+  def create_path(target_resource = nil)
+    return nil if @resource.blank?
+
+    helpers.new_resource_path(**{
+      via_relation: @field.id.to_s,
+      resource: target_resource || @field.target_resource,
+      via_resource_id: resource.model.id,
```
Paul-Bob, line 66: Should we use `resource.model.to_param` instead?

app/components/avo/views/resource_edit_component.rb
```diff
@@ -34,6 +36,23 @@ def resource_view_path
     helpers.resource_view_path(model: association_resource.model, resource: association_resource)
   end
 
+  def resource_edit_or_new_path
+    modal_id = "new_via_belongs_to"
+
+    if params[:via_resource_id].present?
+      # Back to edit path with param indicating turbo stream should simply close modal
+      related_resource = params[:via_belongs_to_resource_class].constantize.new
+      related_record = related_resource.model_class.find(params[:via_resource_id])
```
Paul-Bob, line 45: If we start using `resource.model.to_param` on the `create_path` we need to use [find_record_method](https://docs.avohq.io/2.0/customization.html#custom-find-method-for-show-and-edit-pages) here

Paul-Bob:
Hi @sdcoffey, amazing work! I tested it and it seems to work as expected, I let two comments about using `to_param` instead `id` when passing the record through params. I also have just one more little suggestion about the `modal_component.html.erb`. Since we're not using the heading I think we can wrap the `heading` div on a `if heading.present?` condition like this:

```ruby
<% if heading.present?%>
  <div class="p-6 text-2xl tracking-normal font-semibold text-black">
    <%= heading %>
  </div>
<% end %>
```

WDYT?

Since this is a relatively big PR it would be important to have another review from @adrianthedev.

app/components/avo/fields/belongs_to_field/edit_component.rb
sdcoffey: Good catch! Looks like that's also happening here: 

https://github.com/avo-hq/avo/blob/c94dc00eac0f8bbd4a79055868c96521305aaac2/app/components/avo/fields/belongs_to_field/show_component.rb#L8-L9

Paul-Bob (approved):

sdcoffey:
Hey all! Been a couple weeks since the last activity, wanted to bump this to see about getting it across the finish line

adrianthedev:
Yes @sdcoffey. You're right! It's on me to review it.
Sorry about the delay.

I'll give it a go tomorrow.

sdcoffey:
No worries, just wanted to make sure it didn't get lost–I'm sure y'all are busy! Thank you! 🙏

adrianthedev:
@sdcoffey it works great but I need to check the code a bit and the styling. It's on my plate for the weekend.

sdcoffey:
Sounds good! Thank for the update

adrianthedev (approved):

adrianthedev:
Thanks @sdcoffey for the contribution.
Sorry it took so long to get it merged.

adrianthedev:
Yes. It will be out on Tuesday 🙌

app/components/avo/modal_component.html.erb
```diff
@@ -3,19 +3,26 @@
...
-      </div>
-      <div class="flex justify-end items-baseline space-x-4 p-4 bg-gray-100">
-        <%= controls %>
+        <% if heading? %>
+          <div class="p-6 text-2xl tracking-normal font-semibold text-black">
+            <%= heading %>
+          </div>
+        <% end %>
+
+        <% if content? %>
```
segiddins, line 15: I think this `content?` call is causing a NoMethodError

Paul-Bob: @segiddins  can you describe the scenario? I wasn't able to reproduce it

segiddins: https://github.com/rubygems/rubygems.org/pull/4154 yeah, tests started failing on the version bump

adrianthedev: 👀

adrianthedev: I think you might need to update `view_component` on your end.
I can't put my finger on it exactly, but at some point they changed the way they figure out if the slot has any content inside.

adrianthedev: I see that view_component v3 added this helper. Are you able to update to VC 3?
I see that you are already on 2.8+ which should not bring breaking changes AFAIK.

xeron:
I'm getting a error trying to use new resource links introduced by this PR.

e.g. this URL `/admin/resources/drivers/new?via_belongs_to_resource_class=ResultResource&via_relation=driver` gives me a error:

```
No route matches {:action=>"new", :controller=>"avo/drivers"}
```

but this works just fine `/admin/resources/drivers/new`.

xeron:
@sdcoffey could you please take a look? ^^ It doesn't open a modal for me, it just goes to the link and shows a "No route" error.

adrianthedev:
@xeron was this fixed for you? I can't replicate it on my end.
If you can share a reproduction repository I'll gladly have a look.

xeron:
No, I tried to remove all my avo customizations:

* custom avo stimulus controllers
* custom avo _head and _scripts partials
* custom avo routes

also tried downgrading rails from 7.1 to 7.0.

And it still didn't work. Clicking `Create new resource` just opens a page with a routing error for me.

adrianthedev:
Can you check if you have a JS error somewhere. To me that sounds like the JS is not working. Or Turbo doesn't execute it as a Turbo request.
That should be a turbo request.

xeron:
I think I found the problem.

Enabling the following in `config/initializers/avo.rb` breaks the functionality:

```ruby
config.buttons_on_form_footers = true
```

xeron:
I've reported it as an issue for tracking purposes: https://github.com/avo-hq/avo/issues/2008

adrianthedev:
Pfff. What a weird bug!
Thanks for looking into this.
I'll add it to the roadmap.

Paul-Bob:
> > Can you check if you have a JS error somewhere. To me that sounds like the JS is not working. Or Turbo doesn't execute it as a Turbo request.
> 
> This was the 1st thing I did when I faced the issue :)
> 
> Anyway I think I found the problem.
> 
> Enabling the following in `config/initializers/avo.rb` breaks the functionality:
> 
> ```ruby
> config.buttons_on_form_footers = true
> ```

Back button without valid path was rendering on the form footer, [this PR](https://github.com/avo-hq/avo/pull/2020) fixes it. Thanks for pointing to the right point and for building a detailed issue!

lpe:
It would indeed be important to be able to disconnect this option, if possible at field level. For example I have "Task belongs_to User", I can pick a User (searchable). This certainly does not mean that I can create a User.  Maybe there's another way to disable it? Thx.

xeron:
> It would indeed be important to be able to disconnect this option

Fun fact I was thinking about the same. This feature should be opt-in or opt-out in `belongs_to` field settings.

adrianthedev:
Yup. That makes sense.
Apologies for the troubles.
We'll add a mechanism to hide that option.

Issue #1908, opened 2023-08-13
non-ASCII string in Tab name


zhongsheng:
### Context
```
  tabs do
    field :stores, as: :has_many, name: '📖 store'
    field :addresses, as: :has_many, name: 'store 🧭'
    field :others, as: :has_many, name: '其他store '
  end
```
The above codes will get same *turbo_frame_id*

String#parameterize removed all non-ASCII string [https://github.com/avo-hq/avo/blob/main/lib/avo/tab.rb](url)

zhephyn:
@Paul-Bob I've given this a shot as well.

Paul-Bob:
Hi @zhephyn thanks for giving this a shot.

Did you manage to reproduce the issue? If yes, how?

I've been trying to reproduce it locally without success

```ruby
class Avo::Resources::User < Avo::BaseResource
  def fields
    field :name

    tabs do
      field :posts, as: :has_many, name: "其他store"
      field :comments, as: :has_many, name: "store 🧭"
      field :comment, as: :has_one, name: "📖 store"
    end
  end
end
```

https://www.loom.com/share/fda1a88bd907496da426a4e3573b6b4c?sid=25be617e-af54-46b2-9b56-450e947718b1

---

@zhongsheng, could you confirm whether this issue is still occurring in the latest version of Avo? It seems to have already been fixed.

zhephyn:
Hey @Paul-Bob. Unfortunately I couldn't reproduce the issue. I kept on running into other unrelated errors when I tried.

Paul-Bob:
The original example is not the way to reproduce it, I just get a way to reproduce this:

```ruby
class Avo::Resources::User < Avo::BaseResource
  def fields
    field :name

    tabs do
      tab "其他store", lazy_load: true do
        field :tab_1 do
          "tab_1"
        end
      end

      tab "store 🧭", lazy_load: true do
        field :tab_2 do
          "tab_2"
        end
      end

      tab "📖 store", lazy_load: true do
        field :tab_3 do
          "tab_3"
        end
      end
    end
  end
end
```

1. Replace the `User` resource in our dummy app with the code above.  
2. When on the `main` branch, navigating through the tabs always loads the `tab_1` field.  
3. On the branch where you pushed the fix, it works as expected.

Let's please add some tests for this. You can use the `with_temporary_items` helper to set the user like this, and verify that when each tab is clicked, the expected field is displayed.

zhephyn:
[...]

PR #1910, opened 2023-08-14
fix non-ASCII string in Tab name

zhongsheng:
#1908 
This is my current temporary solution, and there might be better solutions in the future.


adrianthedev:
Would you be able to send some tests for it too, please?

zhongsheng:
> Would you be able to send some tests for it too, please?

I am studying the source code and  unfamiliar with rspec. I'm not quite sure where this test should be placed, in the 'system' or 'components' directory?

PR #1935, opened 2023-08-31
fix: file image doesn't render in full size

### LINKED ISSUE #1897: Vertical image is truncated in file field show view
### Describe the bug

File field with uploaded image doesn't show the full picture, top and bottom are truncated.

I think it comes from `aspect-video` class [here](https://github.com/avo-hq/avo/blob/bbe7c6e69f50a7c32dedbee42572c46b3fcae2d0/app/components/avo/fields/common/files/view_type/grid_component.html.erb#L5).

### Steps to Reproduce

1. Add attachment to the model.
2. Add the file filed in avo resource.
3. Upload an image which has hight greater than width.
4. Navigate to show view.

Operator should be able to see the full image, but it's truncated at the top and bottom.

### Models and resource files

```ruby
  has_one_attached :map do |attachable|
    attachable.variant :thumb, resize_to_limit: [500, 500]
  end
```

```ruby
field :map, as: :file, accept: 'image/*'
```

### System configuration

**Avo version**: 2.38.0
**Rails version**: 7.0.7
**Ruby version**: 3.2.2
**License type**:
 - [x] Community
 - [ ] Pro

 - [x] No

Screenshot 2023-08-10 at 7 54 33 PM

### Additional context

N/A

#### Impact

#### Urgency

Paul-Bob:
# Description

Fixes #1897 

File image was applying `aspect-video` forcing the image to render on 16/9 aspect ratio format. This PR removes that format.

# Checklist:


app/components/avo/fields/common/files/view_type/grid_component.html.erb
```diff
@@ -2,7 +2,7 @@
   <% if file.present? %>
     <div class="flex flex-col justify-between h-full">
       <% if file.representable? && is_image? %>
-        <%= image_tag helpers.main_app.url_for(file), class: 'rounded-lg object-cover w-full aspect-video' %>
+        <%= image_tag helpers.main_app.url_for(file), class: 'rounded-lg object-cover' %>
```
adrianthedev, line 5: I think we still need `w-full`, but will default to your judgement.

adrianthedev (commented):
What if someone wants to apply that aspect ratio?

How about this? 
we add all Tailwind `aspect-` classes in the whitelist file and folks can apply it when they want it with the `html` option.
Is that something we could support?

Paul-Bob:
> What if someone wants to apply that aspect ratio?
> 
> How about this? we add all Tailwind `aspect-` classes in the whitelist file and folks can apply it when they want it with the `html` option. Is that something we could support?

How would you target that specific div with `html` option? I think using `html` option it's possible to apply it oo the wrapper but not that particular div.

Paul-Bob:
About the white list file, there are too many combination, we can't guess all what users want to use (sure we can add the `aspect-[all]` but do we want to add each custom style for each host app on request?), maybe we should provide a hook where users can declare them own whitelist that will merge with ours.

Paul-Bob:
@adrianthedev the `aspect-video` it's applied only when multiple files are rendered as grid. Also renamed some components for better readability

adrianthedev (approved):

Issue #1955, opened 2023-09-28
bug: all_resources does not use a resources custom authorization_policy


ddmck:
### Describe the bug

```
config.main_menu = -> {
  section 'Resources', icon: 'resources' do
    all_resources
  end
end
```
If self.authorization_policy is set on a resource and the current user would pass the check for `index?` the resource does not show up in the navigation menu. I believe this is because https://github.com/avo-hq/avo/blob/3ada1392e6b82c9be99a7c33bad69c60c2c29d20/lib/avo/concerns/fetches_things.rb#L92C17-L92C17              
`Services::AuthorizationService.authorize user, resource.model_class, Avo.configuration.authorization_methods.stringify_keys["index"], raise_exception: false`
is missing the `policy_class: resource.authorization_policy` keyword argument. 

### Steps to Reproduce
1. set 
```
config.license = 'pro'
config.main_menu = -> {
  section 'Resources', icon: 'resources' do
    all_resources
    resource :district
  end
}
``` 
 in config/initializers/avo.rb

set 
```
class PostPolicy
  def index?
    false
  end
  ...
end
``` 
in spec/dummy/app/policies/post_policy.rb

create 
```
class PostOverridePolicy < ApplicationPolicy
  def index?
    true
  end
  ...
end
```
in spec/dummy/app/policies/post_override_policy.rb

set 
```
class PostResource < Avo::BaseResource
  self.authorization_policy = PostOverridePolicy
  ...
end
```
in spec/dummy/app/avo/resources/photo_comment_resource.rb

2. Visit '/avo'
3. Confirm that you don't see Posts in the resources section of the navigatiion.

### Expected behavior & Actual behavior

Expected: Resources that have self.authorization_policy set with a policy that should pass index? should be returned by all_resources even if the default policy would fail.
Actual: Resources that have self.authorization_policy set with a policy that should pass index? are not returned by all_resources if the default policy would fail.

### System configuration
**Avo version**: 2.41.0

**Rails version**: 7.0.4.3

**Ruby version**: 3.2.2 

**License type**:

 - [ ] Community
 - [X] Pro

**Are you using Avo monkey patches, overriding views or view components?**

 - [ ] Yes. If so, please post code samples.
 - [X] No

#### Impact

 - [ ] High impact (It makes my app un-usable.)
 - [ ] Medium impact (I'm annoyed, but I'll live.)
 - [X] Low impact (It's really a tiny thing that I could live with.)

#### Urgency

[...]

PR #1956, opened 2023-09-28
bug: Update fetches_things.rb to use custom policy if available

ddmck:
# Description

Fixes https://github.com/avo-hq/avo/issues/1955

If self.authorization_policy is set on a resource and the current user would pass the check for index? the resource does not show up in the navigation menu. 

Updating the `Services::AuthorizationService.authorize` call in `Avo::Concerns::FetchesThings#get_available_resources` appears to resolve this. 

No tests are failing locally, but I still need to figure out the test suite so that I can test this functionality 

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/lib/avo/concerns/fetches_things.rb
+++ b/lib/avo/concerns/fetches_things.rb
@@ -89,7 +89,7 @@ module Avo
         def get_available_resources(user = nil)
           valid_resources
             .select do |resource|
-              Services::AuthorizationService.authorize user, resource.model_class, Avo.configuration.authorization_methods.stringify_keys["index"], raise_exception: false
+              Services::AuthorizationService.authorize user, resource.model_class, Avo.configuration.authorization_methods.stringify_keys["index"], policy_class: resource.authorization_policy ,raise_exception: false
             end
             .sort_by { |r| r.name }
         end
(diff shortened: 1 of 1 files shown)


lib/avo/concerns/fetches_things.rb
```diff
@@ -89,7 +89,7 @@ def get_model_class_by_name(name)
         def get_available_resources(user = nil)
           valid_resources
             .select do |resource|
-              Services::AuthorizationService.authorize user, resource.model_class, Avo.configuration.authorization_methods.stringify_keys["index"], raise_exception: false
+              Services::AuthorizationService.authorize user, resource.model_class, Avo.configuration.authorization_methods.stringify_keys["index"], policy_class: resource.authorization_policy ,raise_exception: false
```
Paul-Bob, line 92:
```suggestion
      resource.authorization.class.authorize(
        user,
        resource.model_class,
        Avo.configuration.authorization_methods.stringify_keys["index"],
        raise_exception: false
      )
```

Paul-Bob:
Thanks for the investigation and for the PR @ddmck! I made a change suggestion, let me know if it make sense

Issue #1651, opened 2023-03-28
Allow resource menu items to receive params


adrianthedev:
### Feature

From [this discussion](https://discord.com/channels/740892036978442260/740893011994738751/1090017637079207946).

We'd love it if the `resource` menu items would receive the `params` option, which would be a `Hash`, and that would be appended to the generated URL.

```ruby
resource :clients, label: "Active", params: {
  filters: "THE_FILTER_QUERY"
}
```

As a block
```ruby

        resource :projects, params: -> {
          params = {}

          if current_user.is_admin?
            params[:foo] = :bar
          end

          params
        }
```

### Current workarounds

One can use a `link_to` operator.

```ruby
link_to "Active", "/avo/resource/clients?filters=THE_FILTER_QUERY"
```

bryszard:
Hello @adrianthedev, I have created a POC for this feature, but I will need some advice about:
1. Telling if it's the right direction that I took
1. Deciding what to do with the `active_link_to` problem

I explain it in [my PR here](https://github.com/bryszard/avo/pull/1) (didn't set the main repo as the target, because it's POC).

Paul-Bob:
> Additionally, I have a doubt if your intention is to allow also using the `block` for `:params`. I'm not sure, because you're only saying that it's a workaround right now.

Hi @bryszard I let you a detailed review on the PR.
I edited the issue, the intention is to allow a lambda syntax on the `params` argument. It was wrongly on the workaround section.

PR #1957, opened 2023-09-29
feature: allow params on resource menu item

bryszard:
# Description

Fixes https://github.com/avo-hq/avo/issues/1651

In this Pull Request I'm adding an option to pass `params` to the `resource` method on the `MenuBuilder`. It can be a Hash or a Proc that will return a Hash.

Whatever is passed as `params` will be added to the path of the menu item, when navigating to it.

Before this PR, there was a [POC](https://github.com/bryszard/avo/pull/1) on which @Paul-Bob helped me understand the scope of the changes to be done.

In the POC we've identified one caveat for future users of the feature - we don't support having multiple menu items for the same resource, but with different params. The reason is simple - if we have two menu items with the same path, they would be both highlighted as `active` when visited. The workarounds for that using current library `active_link_for` are imperfect and substituting it right now has no point.

As @Paul-Bob stated [here](https://github.com/bryszard/avo/pull/1#issuecomment-1728725362) - we already have a feature to achieve the goal of having different filters for the same resource applied and these are the scopes.

With this feature it will be possible to add some default params to the menu item, which don't necessarily need to be the filters.

# Checklist:

## Video QA

https://github.com/avo-hq/avo/assets/12682792/1f8720dd-24df-4042-908b-6647080cf046

## Manual review steps

1. Open the Avo app on local environment
1. Click on the `Users` resource and check the path

### DIFF
--- a/lib/avo/menu/resource.rb
+++ b/lib/avo/menu/resource.rb
@@ -3,6 +3,7 @@ class Avo::Menu::Resource < Avo::Menu::BaseItem
 
   option :resource
   option :label, optional: true
+  option :params, default: proc { {} }
 
   def parsed_resource
     Avo::App.guess_resource resource.to_s
@@ -10,5 +11,12 @@ class Avo::Menu::Resource < Avo::Menu::BaseItem
 
   def entity_label
     parsed_resource.navigation_label
+  end
+
+  def fetch_params
+    Avo::ExecutionContext.new(
+      target: params,
+      resource: resource
+    ).handle
   end
 end
--- a/lib/avo/menu/resource.rb
+++ b/lib/avo/menu/resource.rb
@@ -6,7 +6,7 @@ class Avo::Menu::Resource < Avo::Menu::BaseItem
   option :params, default: proc { {} }
 
   def parsed_resource
-    Avo::App.guess_resource resource.to_s
+    @parsed_resource ||= Avo::App.guess_resource resource.to_s
   end
 
   def entity_label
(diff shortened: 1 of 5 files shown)


bryszard:
Documentations changes - https://github.com/avo-hq/avodocs/pull/114

According to the Contributing guide I should add some tags to the PR, but I don't know how. I can't change labels.

When it comes to the Checklist point "I have commented my code, particularly in hard-to-understand areas" - I don't think there is part of the code that would require commenting. Correct me if I'm wrong.

lib/avo/menu/resource.rb
```diff
@@ -11,4 +12,11 @@ def parsed_resource
   def entity_label
     parsed_resource.navigation_label
   end
+
+  def fetch_params
+    Avo::ExecutionContext.new(
+      target: params,
+      resource: resource
```
Paul-Bob, line 19:
```suggestion
      resource: parsed_resource
```

lib/avo/menu/resource.rb
```diff
@@ -3,6 +3,7 @@ class Avo::Menu::Resource < Avo::Menu::BaseItem
 
   option :resource
   option :label, optional: true
+  option :params, default: proc { {} }
 
   def parsed_resource
     Avo::App.guess_resource resource.to_s
```
Paul-Bob, line 9:
```suggestion
    @parsed_resource ||= Avo::App.guess_resource resource.to_s
```

Paul-Bob:
Thank you for this amazing PR @bryszard! I let some tweaks suggestions and also added the tags so don't worry about it.

Paul-Bob (approved):

Issue #1926, opened 2023-08-24
Allow to customize the maximum number of suggestion on tags field


Paul-Bob:
### Feature

Currently the max number of suggestion on a tags field is hard-coded to 20. Allow to customize the maximum number of suggestion on tags field

From [here](https://discord.com/channels/740892036978442260/1144207643389468725)

PR #1988, opened 2023-10-21
feature: allow suggestions_max_items option on tags field

bryszard:
Fixes https://github.com/avo-hq/avo/issues/1926

# Description

We're using a JS library to handle tags `@yaireo/tagify`. 
There is a feature there that allows to show to the users tag suggestions - we provide a list of potential suggestions and we have some options around it. 
One of those options is `dropdown.max_items` which we have hard-coded currently to 20.

We want to allow users of Avo to set their own `max_items` limit, keeping the 20 as default.

# Checklist:

## Visual QA

https://github.com/avo-hq/avo/assets/12682792/f2890331-a9e9-4a11-8b2a-8f6713e98b36

## Manual review steps

1. Check the `Posts` resource in the dummy app
1. Remove the tags and show how many tags show up (limit is set to 2, there are 3 options, so you should see 2 and info that more tags are there but hidden)

### DIFF
--- a/lib/avo/fields/tags_field.rb
+++ b/lib/avo/fields/tags_field.rb
@@ -5,6 +5,7 @@ module Avo
       attr_reader :close_on_select
       attr_reader :delimiters
       attr_reader :enforce_suggestions
+      attr_reader :suggestions_max_items
       attr_reader :mode
 
       def initialize(id, **args, &block)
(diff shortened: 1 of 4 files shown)


lib/avo/fields/tags_field.rb
```diff
@@ -16,6 +17,7 @@ def initialize(id, **args, &block)
         add_array_prop args, :disallowed
         add_array_prop args, :delimiters, [","]
         add_array_prop args, :suggestions
+        add_string_prop args, :suggestions_max_items
```
bryszard, line 20: It's technically not a string, but there is no "add_integer_prop" method and it is working with `add_string_prop`.

bryszard:
I have encountered 2 problems:
1. I cannot add label on my own, so cannot fix the `labeler` CI issue.
2. I don't see any tests for various options of the `TagsField`, so I don't know if and where should I add tests for this feature.

Let me know how to deal with those issues.

adrianthedev (approved):
Sublime! Thank you @bryszard

PR #1992, opened 2023-10-25
fix: actions and field hydration

### LINKED ISSUE #1946: The fix for #1865 has broken default procs on custom actions
The commit at 5db08d462978ff77f4f4d0cfca0b5689923a3374 breaks the following use case:

```ruby
class CustomAction < Avo::BaseAction
  self.name = 'Custom Action'
  self.message = 'Some kind of custom action'

  field :option_a, as: :boolean, default: -> { @record.can?(:option_a) }
  field :option_b, as: :boolean, default: -> { @record.can?(:option_b) }
  field :option_c, as: :boolean, default: -> { @record.can?(:option_c) }
end
```

In this case, I've noticed that since the commit at 5db08d462978ff77f4f4d0cfca0b5689923a3374, the default value procs for `:option_b` and `:option_c` no longer have `@record` populated and the value is simply `nil`. This was working in the commit immediately before 5db08d462978ff77f4f4d0cfca0b5689923a3374. 

### System configuration
**Avo version**:  v2.37.1 and above

**Rails version**: 7.0.8

**Ruby version**: 3.2.2

**License type**:

 - [ ] Community
 - [x] Pro

 - [x] No

#### Impact

#### Urgency

Paul-Bob:
# Description

Fixes #1946

This pull request addresses primarily two key issues: the class attributes for actions and the field hydration method.

## Action class attributes
Relying on class attributes for actions is not convenient since undesired old values can remain populated on a new instance.

## Field Hydration Method
The existing code made it impossible to hydrate the field with `nil` values, resulting in undesired behavior. For instance, in cases where a field within an action relied on the `@record` via a default Proc, it would correctly receive the default value when executed in a show view. However, when executed in an index view, the field attempted to hydrate with `model: nil`, yet the prior `model` would remain hydrated, causing the value to be fetched incorrectly.

## `ActionModel` removed
`ActionModel` was not being used only to pass it as `model` on `form_with` helper, and I think that was unecessary.

# Checklist:

## Manual review steps

1. Create an action with several fields that have the `default` value as Proc reling on the `@record`
  Example: 
```ruby
class ReleaseFish < Avo::BaseAction
  field :option_a, as: :boolean, default: -> { @record&.foo? }
  field :option_b, as: :boolean, default: -> { @record&.foo? }
  field :option_c, as: :boolean, default: -> { @record&.foo? }

  def handle(...)
    succeed ":)"
  end
end
```
2. Define the method on the model
  Example:
```ruby
  def foo?
    true
  end
```
3. Register the action on a resource
4. Go to that resource index page
5. Select 2 records
6. Run the action
7. Notice that all fields are false since `@record` is blank
8. Go to one record show view
9. Run the action
10. Notice that all fields are true since `@record` is setted ( and the default proc is running with `@record` properly hydrated on all executions )
11. Return to index and run action after selecting 2 records
12. Notice the same as step 7 (Prior this commit the fields would still hydrated with the `@record` from the show view since was impossible to hydrate with nil values)


lib/avo/fields/base_field.rb
```diff
@@ -99,14 +99,20 @@ def initialize(id, **args, &block)
...
+        permited_kwargs_keys = %i[model resource action view panel_name user]
 
+        # Check for unrecognized keys
+        unrecognized_keys = kwargs.keys - permited_kwargs_keys
+        raise ArgumentError, "Unrecognized argument(s): #{unrecognized_keys.join(', ')}" if unrecognized_keys.any?
+
+        # Set instance variables with provided values
+        kwargs.each do |key, value|
+          instance_variable_set("@#{key}", value)
+        end
```
adrianthedev, line 113: why the refactor here? it feels difficult to understand at a glance

lib/avo/base_action.rb
```diff
@@ -59,10 +55,10 @@ def action_name
...
 
     def initialize(model: nil, resource: nil, user: nil, view: nil, arguments: {})
-      self.class.model = model
-      self.class.resource = resource
-      self.class.user = user
-      self.class.view = view
+      @model = model
+      @resource = resource
+      @user = user
+      @view = view
```
adrianthedev, line 61: Pfff. what a good fix!
THanks!

app/controllers/avo/actions_controller.rb
```diff
@@ -16,7 +16,6 @@ def show
       @view = :new
 
       @resource.hydrate(model: @model, view: @view, user: _current_user, params: params)
-      @model = ActionModel.new @action.get_attributes_for_action
```
adrianthedev, line 19: Not sure that this is the right decision.
I can't remember why, but at some point I know for a fact that it wasn't the right decision to use the `@model`.
Let's talk about it.

lib/avo/fields/base_field.rb
Paul-Bob: The existing code made it impossible to hydrate the field with `nil` values, resulting in undesired behavior. For instance, in cases where a field within an action relied on the `@record` via a default Proc, it would correctly receive the default value when executed in a show view. However, when executed in an index view, the field attempted to hydrate with `model: nil`, yet the prior `model` would remain hydrated, causing the value to be fetched incorrectly.

Paul-Bob: What this does is if we use `hydrate(model: nil)` will assign `@model = nil` without touching the other variables... With the old code `@model = model if model.present?` since `@model.present? == false` the `@model` will keep the last assigned `model` even if we want to force hydrate it to nil...

adrianthedev (approved):
Ok, let's merge it but let's keep an eye on it and mention something in the upgrade guide that we changed something internal and if they see anything off, they should reach out to us.

Paul-Bob:
> Ok, let's merge it but let's keep an eye on it and mention something in the upgrade guide that we changed something internal and if they see anything off, they should reach out to us.

https://github.com/avo-hq/avodocs/pull/127

PR #2016, opened 2023-11-06
fix: message don't appear on fail destroy on associations

### LINKED ISSUE #1918: Deleting a record with validation errors within has_many view hangs on reloading view
### Describe the bug
When an attempt is made to destroy a resource within a Parent resources `field :children, as: :has_many`, if the deletion fails because of a validation error the turbo frame fails to update, and is stuck on the loading view. The response is passed and looks like a proper turbo frame update to me, I couldn't identify the flaw.

### Steps to Reproduce
Drop the following single-file app into a file called `app.ru` and run `rackup app.ru`. I found the issue in my own app, then verified it in a `rails new --minimal`, and finally knocked this thing together to simplify reproduction.

- Navigate to /resources/parents/1
- Click on the delete button next to Child
- See that children has_many view hangs on loading
- Reload page, click on the delete button next to Grandchildless Child
- See that it behaves as expected

```ruby
require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'

  gem 'rails', '~> 7.0.4'
  gem 'sqlite3'
  gem 'avo', '2.38.0'
  gem 'puma'
  gem 'propshaft'
end

require 'rails/all'

database = 'development.sqlite3'
ENV['DATABASE_URL'] = "sqlite3:#{database}"
ActiveRecord::Base.establish_connection(adapter: 'sqlite3', database: database)
ActiveRecord::Base.logger = Logger.new(STDOUT)
ActiveRecord::Schema.define do
  create_table :parents, if_not_exists: true do |t|
    t.string :name
  end
  create_table :children, if_not_exists: true do |t|
    t.references :parent, null: false, foreign_key: true
    t.string :name
  end
  create_table :grandchildren, if_not_exists: true do |t|
    t.references :child, null: false, foreign_key: true
    t.string :name
  end
end

class App < Rails::Application
  config.root = __dir__
  config.eager_load = true
  config.consider_all_requests_local = true
  config.secret_key_base = 'Aswertyuioasdfghjkqwertyuiqwerty'
  config.active_storage.service_configurations = { 'local' => { 'service' => 'Disk', 'root' => './storage' } }
  config.active_record.legacy_connection_handling = false


  routes.append do
    mount Avo::Engine, at: "/"
  end
end

App.initialize!

class Parent < ActiveRecord::Base
  has_many :children
end

class Child < ActiveRecord::Base
  belongs_to :parent
  has_many :grandchildren, dependent: :restrict_with_error
end

class Grandchild < ActiveRecord::Base
  belongs_to :child
end

class ParentResource < Avo::BaseResource
  field :name, as: :text
  field :children, as: :has_many
end

class ChildResource < Avo::BaseResource
  field :parent, as: :belongs_to
  field :name, as: :text
  field :grandchildren, as: :has_many
end

class GrandchildResource < Avo::BaseResource
  field :child, as: :belongs_to
  field :name, as: :text
end

class Avo::ParentsController < Avo::ResourcesController
end

class Avo::ChildrenController < Avo::ResourcesController
end

class Avo::GrandchildrenController < Avo::ResourcesController
end

Rails.application.reload_routes!

parent = Parent.find_or_create_by! name: 'Parent'
child = Child.find_or_create_by! name: 'Child', parent: parent
_grandchild = Grandchild.find_or_create_by! name: 'Grandchild', child: child
_grandchildless_child = Child.find_or_create_by! name: 'Grandchildless Child', parent: parent

run App
```

View should refresh and error flash message should appear.

### System configuration
**Avo version**: 2.38.0

**Rails version**: 7.0.4

**Ruby version**: 3.2.2

**License type**:
 - [x] Community
 - [x] Pro

 - [x] No

image

### Additional context
The single file is handy for IDing bugs, but It is necessary to run `App.initialize!` prior to defining controllers so that `Avo::ResourcesController` is available, then you gotta `Rails.application.reload_routes!` so (I believe) the Avo dynamic router can pick them up. Just a note for anyone who might find that useful.

#### Impact

#### Urgency
I haven't found a workaround but it isn't exactly a showstopper. It will require explanation to my internal users which is a bummer.

Paul-Bob:
# Description

Fixes #1918

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/app/controllers/avo/base_controller.rb
+++ b/app/controllers/avo/base_controller.rb
@@ -504,8 +504,10 @@ module Avo
     end
 
     def destroy_fail_action
+      flash[:error] = destroy_fail_message
+
       respond_to do |format|
-        format.html { redirect_back fallback_location: params[:referrer] || resources_path(resource: @resource, turbo_frame: params[:turbo_frame], view_type: params[:view_type]), error: destroy_fail_message }
+        format.turbo_stream { render partial: "avo/partials/flash_alerts" }
       end
     end
 
--- a/app/views/avo/partials/_flash_alerts.turbo_stream.erb
+++ b/app/views/avo/partials/_flash_alerts.turbo_stream.erb
@@ -1,3 +1,3 @@
 <%= turbo_stream.append "alerts" do %>
-  <%= render Avo::FlashAlertsComponent.new flashes: flash %>
+  <%= render Avo::FlashAlertsComponent.new flashes: flash.discard %>
 <% end %>


app/controllers/avo/base_controller.rb
```diff
@@ -504,8 +504,10 @@ def destroy_success_action
     end
 
     def destroy_fail_action
+      flash[:error] = destroy_fail_message
+
       respond_to do |format|
-        format.html { redirect_back fallback_location: params[:referrer] || resources_path(resource: @resource, turbo_frame: params[:turbo_frame], view_type: params[:view_type]), error: destroy_fail_message }
+        format.turbo_stream { render partial: "avo/partials/flash_alerts" }
```
adrianthedev, line 510: Are you sure that removing the `redirect` does not break other flows?

app/views/avo/partials/_flash_alerts.turbo_stream.erb
```diff
@@ -1,3 +1,3 @@
 <%= turbo_stream.append "alerts" do %>
-  <%= render Avo::FlashAlertsComponent.new flashes: flash %>
+  <%= render Avo::FlashAlertsComponent.new flashes: flash.discard %>
```
adrianthedev, line 2: Are we sure that the `discard` works well with other flows as well?

app/controllers/avo/base_controller.rb
Paul-Bob: None that I tested or that is covered by automatic tests

app/views/avo/partials/_flash_alerts.turbo_stream.erb
Paul-Bob: I think so, it will discard the entire flash at the end of the current action

PR #2034, opened 2023-11-16
fix: detach should not destroy attachment

Paul-Bob:
# Description

[This change](https://github.com/avo-hq/avo/pull/1856/files#diff-b1f64e5509fa895bb0217e2740a262487a1344069c30bef0391e92f05c760b91) intended to make sure that validations / callbacks are called when detaching a record from a `has_many` associations. Unfortunately it introduced a bug where the record getting detached is actually deleted from DB.

[`destroy(*records)`](https://edgeapi.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-destroy)

> ...
This method will always remove record from the database ignoring the :dependent option.
...

Delete is what we want here but [`delete(*records)`](https://edgeapi.rubyonrails.org/classes/ActiveRecord/Associations/CollectionProxy.html#method-i-delete) will not run associations callbacks...

Fixes https://discord.com/channels/740892036978442260/1157144752752427048

We should find a way to trigger associations callbacks using `delete` or another strategy.

# Checklist:

### DIFF
--- a/app/controllers/avo/associations_controller.rb
+++ b/app/controllers/avo/associations_controller.rb
@@ -81,7 +81,7 @@ module Avo
       association_name = BaseResource.valid_association_name(@model, params[:related_name])
 
       if reflection_class == "HasManyReflection"
-        @model.send(association_name).destroy @attachment_model
+        @model.send(association_name).delete @attachment_model
       else
         @model.send("#{association_name}=", nil)
       end


adrianthedev (approved):

Paul-Bob:
`delete` triggered `before_remove` and `after_remove` on my tests...

yshmarov:
thanks!

Issue #1938, opened 2023-09-04
Assign Stimulus controllers to actions 


Paul-Bob:
### Feature

Similar to how we [assign stimulus controllers to resources](https://docs.avohq.io/2.0/stimulus-integration.html#assign-stimulus-controllers-to-resource-views) we should be able to assign stimulus controllers to actions.

Another possibility is to reuse the controller assigned to the resource and be able to use it on all the actions used by that resource.

Paul-Bob:
From [here](https://discord.com/channels/740892036978442260/1147268129421004871)

bryszard:
I'll make my attempt to solve this one.

Paul-Bob:
Thank you @bryszard let us know if you need any help!

PR #2039, opened 2023-11-18
feature: allow assigning Stimulus controllers to actions

bryszard:
# Description

Fixes [# 1938](https://github.com/avo-hq/avo/issues/1938)

## Context

As explained in the linked issue, we want to be able to assign Stimulus controllers to actions, similarly as we do to the resources.

I was considering reusing the controller assigned to the resource, but I've decided that would make the feature less flexible and less understandable. Why should we be binding the JS logic on action to a given resource? What if we'd like to reuse an action for different resources?

## Changes

In order to enable the feature I had to do the following:
1. Add `stimulus_controllers` class attribute and `get_stimulus_controllers` instance method to `BaseAction`. 
I've considered reusing the concern `HasStimulusControllers`, but it would be an overfit in here because as far as I understand we don't need to add view type specific attributes.
2. Pass `action` to the `BaseField` in order to later use it while rendering.
3. Add Stimulus data attributes - `controller` to the top-level action element and `target` to all fields in the action form.

As an example I've reused existing Stimulus controller `course_resource_controller`. 
I've renamed to `city_in_country_controller`, because it is not specific to course resource.

## Out of scope

What are differences from the Stimulus controllers added for resources:
1. I don't add default controllers per view, like we do for `resource-edit`, `resource-show`, and `resource-index`.
2. I don't pass the `view` attribute to the DOM. The main reason is that currently standalone action is getting the `view` value from [here](https://github.com/avo-hq/avo/blob/3f6110b231407712d423ab7ea6616ebaa7aed6ae/app/controllers/avo/actions_controller.rb#L57) and it will be always `:new`. To get the "real" value I believe some refactoring would be required.

Let me know if you think we should do it differently. 

# Checklist:

## Visual QA

https://github.com/avo-hq/avo/assets/12682792/c995a51e-9fd8-4d41-ae30-acd7d11f6a2f

## Manual review steps

1. Go to the dummy app to Courses page
2. Check the "Show current time" action
3. Check if the action also works on show and edit pages


bryszard:
@Paul-Bob I have a problem with failing `i18n_spec`, but the problem does not seem to be related to my changes:

```
Missing translations (6) | i18n-tasks v1.0.12
+--------+------------------------------------------+-------------------------------------------------------------------------------------------+
| Locale | Key                                      | Value in other locales or source                                                          |
+--------+------------------------------------------+-------------------------------------------------------------------------------------------+
|   nb   | avo.records_selected_from_all_pages_html | en All records selected from all pages                                                    |
|   nb   | avo.x_records_selected_from_page_html    | en <span class="font-bold text-gray-700">%{selected}</span> records selected on this page |
|   nn   | avo.records_selected_from_all_pages_html | en All records selected from all pages                                                    |
|   nn   | avo.x_records_selected_from_page_html    | en <span class="font-bold text-gray-700">%{selected}</span> records selected on this page |
|   tr   | avo.records_selected_from_all_pages_html | en All records selected from all pages                                                    |
|   tr   | avo.x_records_selected_from_page_html    | en <span class="font-bold text-gray-700">%{selected}</span> records selected on this page |
+--------+------------------------------------------+-------------------------------------------------------------------------------------------+
```

Paul-Bob:
> @Paul-Bob I have a problem with failing `i18n_spec`, but the problem does not seem to be related to my changes:
> 
> ```
> Missing translations (6) | i18n-tasks v1.0.12
> +--------+------------------------------------------+-------------------------------------------------------------------------------------------+
> | Locale | Key                                      | Value in other locales or source                                                          |
> +--------+------------------------------------------+-------------------------------------------------------------------------------------------+
> |   nb   | avo.records_selected_from_all_pages_html | en All records selected from all pages                                                    |
> |   nb   | avo.x_records_selected_from_page_html    | en <span class="font-bold text-gray-700">%{selected}</span> records selected on this page |
> |   nn   | avo.records_selected_from_all_pages_html | en All records selected from all pages                                                    |
> |   nn   | avo.x_records_selected_from_page_html    | en <span class="font-bold text-gray-700">%{selected}</span> records selected on this page |
> |   tr   | avo.records_selected_from_all_pages_html | en All records selected from all pages                                                    |
> |   tr   | avo.x_records_selected_from_page_html    | en <span class="font-bold text-gray-700">%{selected}</span> records selected on this page |
> +--------+------------------------------------------+-------------------------------------------------------------------------------------------+
> ```

It's not related @bryszard, don't worry. We're missing some translation on a new feature

bryszard:
@Paul-Bob Ok, in that case I think it's ready for review :).

bryszard:
Thank you, Paul :). So, it will be merged only to version 2.0? Or also copied to 3.0?

Paul-Bob:
> Thank you, Paul :). So, it will be merged only to version 2.0? Or also copied to 3.0?

Everything that is merged into `2.x` is also merged into `3.x`. I'll "copy" your PR and apply it to `3.x` after we merge it here.

adrianthedev (approved):
Look good! Thanks @bryszard! This is going to improve actions so much!

PR #2048, opened 2023-11-22
fix: create new record from belongs_to associations

### LINKED ISSUE #2045: Create new record from belongs_to associations crashes the app
avo-hq/avo#1876 introduced a bug.

if you're on a `New` page and you have a `belongs_to` association it will crash the page because it will try to build a `create` path for a record that has not been created yet.

Paul-Bob:
# Description

` via_resource_id: resource.record.to_param` when `resource.record` is not a persistent yet `to_param` will return `nil` and that's fine. But in case you use `prefix_id` or some other ID manipulation it will break if the attributes used on the manipulation are blank.

Fixes #2045 

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/app/components/avo/fields/belongs_to_field/edit_component.rb
+++ b/app/components/avo/fields/belongs_to_field/edit_component.rb
@@ -63,7 +63,7 @@ class Avo::Fields::BelongsToField::EditComponent < Avo::Fields::EditComponent
     helpers.new_resource_path(**{
       via_relation: @field.id.to_s,
       resource: target_resource || @field.target_resource,
-      via_record_id: resource.record.to_param,
+      via_record_id: resource.record.persisted? ? resource.record.to_param : nil,
       via_belongs_to_resource_class: resource.class.name
     }.compact)
   end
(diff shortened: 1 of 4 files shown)


adrianthedev (changes requested):
Can you please ccheck searchable as well? And add proper tests to cover these scenarios?

app/components/avo/fields/belongs_to_field/edit_component.rb
```diff
@@ -63,7 +63,7 @@ def create_path(target_resource = nil)
     helpers.new_resource_path(**{
       via_relation: @field.id.to_s,
       resource: target_resource || @field.target_resource,
-      via_record_id: resource.record.to_param,
+      via_record_id: resource.record.persisted? ? resource.record.to_param : nil,
```
adrianthedev, line 66: Will this hide the link from the `html` template? If not, then what kind of link will that be? How will that record be created if it doesn't have an associated id?

Issue #1960, opened 2023-10-03
Show special notification for when a resource is missing


adrianthedev:
From [here](https://discord.com/channels/740892036978442260/1138182212655788054/1157947567561261066).

We know it's missing. We just need better signaling and instructions on what they need to do.

Paul-Bob:
When rendering an association field and it's resource is missing we should raise an custom error that includes 

- Information about that ressource is missing for that particular field
- Example how to generate the resource (ex: `rails generate avo:resource use`)
- Quick referal to `use_resource` option and maybe a link to it's documentation

This error is just an example:
`Avo -> Users resources missig while rendering 'field :users, as: :has_many'. You can generate that resource running 'rails generate avo:resource user' or you can use the 'use_resource' option to specify the resource to be used on the field (maybe a link to docs).`

bryszard:
I'll take this one, if I may :)

PR #2062, opened 2023-11-26
Feature/better error for missing resource

bryszard:
# Description

Fixes [# 1960](https://github.com/avo-hq/avo/issues/1960)

We're adding a better handling for errors when we encounter a missing resource on association fields.

Right now the error is unhelpful and requires developer to understand the internal logic of Avo. 
We want to give a better Dev Experience and point to solutions.

# Checklist:

## Screenshots

Error:

UPDATED WITH NEW MESSAGE
image

## Video

BEFORE APPLYING REVIEW SUGGESTIONS
https://github.com/avo-hq/avo/assets/12682792/3b4f14f8-700a-4770-ba37-cecd66683052

## Manual review steps

1. Go to the dummy app to Teams page and choose any Team, add `show_location_field=1` query param
1. On the Team show page go to the bottom and observe what happens with the slot for `locations` field.
1. Go to the Courses page and choose any Course, add `show_location_field=1` query param
1. On the Course show page go to the bottom and observe what happens with the slot for `locations` field.
1. Go to the Stores page , add `show_location_field=1` query param and see the error
1. Go to the Events page, add `show_location_field=1` query param and see the error


adrianthedev (commented):
This looks great!

I think this catch can be added to `BelongsToField.target_resource`, `HasBaseField.field_resource` and `HasBaseField.target_resource`.

spec/system/avo/exposed_methods_base_controller_spec.rb
```diff
@@ -1,6 +1,13 @@
 require "rails_helper"
 
 RSpec.feature "ExposedMethodsBaseController", type: :system do
+  around do |example|
+    old_value = Capybara.raise_server_errors
+    Capybara.raise_server_errors = false
```
bryszard, line 6: I'm not sure how those errors (our new custom ones) are triggered in the test env, because in the development it is not triggered when course is created, updated or deleted.

bryszard:
@Paul-Bob I'm ready for review, although I have a doubt:

Since I've merged the newest `main` I have a problem with the display of the dummy app. The resources are not grouped correctly as they used to and as they do when I check the [Demo App](https://main.avodemo.com/avo/dashboards/dashy). They are all displayed in one non-collapsable group in alphabetical order:

image

Do you know what could be the problem?

Paul-Bob:
@bryszard 

> The resources are not grouped correctly as they used to

They are grouped correctly give a read here https://docs.avohq.io/3.0/resource-panels.html

We refactored the way that items are fetched from resource on avo3 giving more customization

Let me know if the docs are clear or if there are some thinks that can be better explained

adrianthedev:
I'm checking your commits and you're a legend @bryszard.
Thank you for taking the time to fix these 💪
We appreciate it.

spec/system/avo/exposed_methods_base_controller_spec.rb
Paul-Bob: Actually the error in development is there but suppressed on the frame.

image

I'm not totally convinced by `Capybara.raise_server_errors = false` strategy since it can omit other errors that we want the test to cover.

Can we remove the capybara server errors suppression and change the dummy app to avoid this breaks?

If some fields are needed on a specific spec we always can use the `with_temporary_items` helper

Paul-Bob (changes requested):
Great PR @bryszard! Thank you!!

I let a comment about removing the error suppression on tests, anything else it's great!

spec/system/avo/skip_show_view_spec.rb
```diff
@@ -4,6 +4,13 @@
   let!(:course) { create :course }
   let!(:post) { create :post }
 
+  around do |example|
+    old_value = Capybara.raise_server_errors
+    Capybara.raise_server_errors = false
+    example.run
+    Capybara.raise_server_errors = old_value
+  end
```
adrianthedev, line 12: I don't think we should disable errors like this.

spec/system/avo/resource_sidebar_spec.rb
```diff
@@ -4,6 +4,13 @@
   let(:fish) { create :fish }
   let(:team) { create :team }
 
+  around do |example|
+    old_value = Capybara.raise_server_errors
+    Capybara.raise_server_errors = false
+    example.run
+    Capybara.raise_server_errors = old_value
+  end
```
adrianthedev, line 12: I don't think we should disable errors like this.

adrianthedev (changes requested):
I tweaked the message and the tests.

I noticed that you disabled the error catching, but I think it's important to not do that.
We can check if an error was raised. I pushed a sample on the test you created.

On the last two specs we should not disable the error catching.
We should take a different approach. Maybe like I did on events?
Show the field only if a `param` is present?
This way we won't get random crashes when we do other tests or navigate the app, but still be able to test that scenario.

One thing that I'm a bit afraid of is coupling of the database tables.
For example, you can't have an event without a location. So in reality in the UI we will never be able to create an event because it will fail on the new page (no location resource present).
How about making the location optional to the event?
Maybe the same with location and stores?

Thsi way they are not so coupled and will give us some leeway with creating records.



## To do:

- [ ] Make models less coupled
- [ ] remove the error catching from the two unrelated tests
- [ ] show the error fields only if a param is present so we can test them

This is solid progress!
Thank you!

Let's address the issues above and we'll be able to merge it very soon.

adrianthedev:
BTW @bryszard, are you on our [Discord](https://avo.cool/chat)?

bryszard:
> BTW @bryszard, are you on our [Discord](https://avo.cool/chat)?

Yes, Adrian, I am there :)

bryszard:
Thank you guys for the review. I'm back from holiday and will be able to adjust it in the next days.

spec/system/avo/exposed_methods_base_controller_spec.rb
bryszard: Sure thing, we can do that. 

My intention though was to have an example of how this error would look like for each of the association type in the dummy app. Maybe we don't need to keep it and only trigger it in specs using the `with_temporary_items` helper.

bryszard:
> @bryszard
> 
> > The resources are not grouped correctly as they used to
> 
> They are grouped correctly, give a quick read here https://docs.avohq.io/3.0/resource-panels.html
> 
> We refactored the way that items are fetched from resource on avo3 allowing more customization
> 
> Let me know if the docs are clear or if there are some thinks that can be better explained
> 
> PS: Thank you for the contribution. I'll review it during this week

@Paul-Bob I think I found the reason - you've moved the `Avo::Menu::Builder` to the `avo-menu` plugin and it's under Pro licence (https://docs.avohq.io/3.0/menu-editor.html). When I run the dummy app I don't have it.

bryszard:
@adrianthedev I've applied all the suggestions, but there is one failing spec:

After moving the `model_missing_resource_spec` to `spec/features` now for some reason it is raising `#<ActionController::RoutingError: No route matches [GET] "/admin/resources/events">`. 

The problem is that I cannot reproduce it locally with `bin/test ./spec/features/avo/model_missing_resource_spec.rb`. I also don't see a problem with that route when running `bin/dev`. Could you help me here? :)

Paul-Bob:
> > @bryszard
> > > The resources are not grouped correctly as they used to
> > 
> > 
> > They are grouped correctly, give a quick read here https://docs.avohq.io/3.0/resource-panels.html
> > We refactored the way that items are fetched from resource on avo3 allowing more customization
> > Let me know if the docs are clear or if there are some thinks that can be better explained
> > PS: Thank you for the contribution. I'll review it during this week
> 
> @Paul-Bob I think I found the reason - you've moved the `Avo::Menu::Builder` to the `avo-menu` plugin and it's under Pro licence (https://docs.avohq.io/3.0/menu-editor.html). When I run the dummy app I don't have it.
> 
> ```
> irb(main):006> Avo::Menu::Builder
> (irb):6:in `<main>': uninitialized constant Avo::Menu (NameError)
> ```

Ah you meant the menu, yes, on avo versions > 3 the main branch contains mainly community version features, the pro/advanced features are distributed by packages.

PR #2102, opened 2023-12-13
fix: `belongs_to` index field name

### LINKED ISSUE #2099: `belongs_to` fields not applying `name` option on index view
### Describe the bug

`belongs_to` fields are not applying `name` option on index view.

`field :commentable, as: :belongs_to, polymorphic_as: :commentable, types: [::Post, ::Project], name: "Test"`

### Index
image

### Show
image

Paul-Bob:
# Description

`belongs_to` name on index was hard-coded.

Fixes #2099 

# Checklist:

### DIFF
--- a/lib/avo/fields/belongs_to_field.rb
+++ b/lib/avo/fields/belongs_to_field.rb
@@ -259,8 +259,8 @@ module Avo
         nil
       end
 
-      def name
-        return polymorphic_as.to_s.humanize if polymorphic_as.present? && view.index?
+      def default_name
+        return polymorphic_as.to_s.humanize if polymorphic_as.present?
 
         super
       end


lib/avo/fields/belongs_to_field.rb
```diff
@@ -259,8 +259,8 @@ def get_model
         nil
       end
 
-      def name
-        return polymorphic_as.to_s.humanize if polymorphic_as.present? && view.index?
```
adrianthedev, line 263: Removing `&& view.index?` is ok? I mean it does seem to have it's place...

Paul-Bob: I think so, why we would apply the polymorphic_as default name only on index? That means that on other views would fallback to 

```ruby
def default_name
    @id.to_s.humanize(keep_id_suffix: true)
  end
```

PR #2114, opened 2023-12-15
Add base_field_translation_key to BaseResouce allowing fields to be nested localized

### LINKED ISSUE #1593: Automatic translation lookup
### Feature
I wonder if we could set some automatic implicit translation lookup rules. For instance, couldn't we set `translation_key` for each field to go look into `avo.field_translations.RESOURCE_NAME.FIELD_NAME` instead of having to specify the same route for every field, and every resource ?

Also, I don't believe `Actions` and `Filters` have a specific way of handling translations (I manually resort to calling `I18n` and a custom path). I think this could be a cool opportunity to add global, implicit I18n support to avo in most parts of the gem :)

Unless I'm completely off and all this already exists, but I think I re-read documentation quite a few times already 😅

What do you think ?

### Current workarounds
Right now it looks like this:
```ruby
  field :ease, as: :number, hide_on: :index, translation_key: "avo.fields_translations.action_suggestion.ease"
  field :cost, as: :number, hide_on: :index, translation_key: "avo.fields_translations.action_suggestion.cost"
  field :link, as: :text, hide_on: :index, translation_key: "avo.fields_translations.action_suggestion.link"
  field :created_at, as: :date_time, hide_on: [:index, :edit, :new], translation_key: "avo.fields_translations.action_suggestion.created_at"
```

If we set some clear, logic lookup rule (just like some parts of Rails do), we could get rid of having to specify `translation_key` everywhere, as long as one would respect the said rule, but still allow manual override.

djalmaaraujo:
# Description

Currently Avo supports users to Localize Resources or Fields as in a root-level perspective. You can define `translation_key` for resources or each field, but you cannot tell avo to translate fields from a specific nested base point.

The proposed change allow users to continue to use the regular approach, but also adds the support for specifying where the fields from a specific resource will be localized.

This is useful for migrating big applications with other Admin capabilities, but also it allows users to have the same field name in different locations. 

Fixes #1593

# Checklist:

## Manual review steps

1. Pull the branch
2. Change your `avo.en.yml` locale file

```yml
# lib/generators/avo/templates/locales/avo.en.yml
field_translations:
      status: Status Root
      project:
        status:
          zero: Status Nesteds
          other: Status Nesteds
          one: Status Nested
```

3. Inside `class Avo::Resources::Project < Avo::BaseResource`, write:

```rb
self.base_field_translation_key = "avo.field_translations.project"
```

4. Go to the dummy project /projects path
5. You should expect to see Status Nested as the table column instead of just Status

Output:

image

### DIFF
--- a/lib/avo/base_resource.rb
+++ b/lib/avo/base_resource.rb
@@ -48,6 +48,7 @@ module Avo
     class_attribute :includes, default: []
     class_attribute :authorization_policy
     class_attribute :translation_key
+    class_attribute :base_field_translation_key
     class_attribute :default_view_type, default: :table
     class_attribute :devise_password_optional, default: false
     class_attribute :scopes_loader
--- a/lib/avo/fields/base_field.rb
+++ b/lib/avo/fields/base_field.rb
@@ -95,7 +95,7 @@ module Avo
       def translation_key
         return @translation_key if @translation_key.present?
 
-        return "#{@resource.base_field_translation_key}.#{@id}" if @resource.base_field_translation_key.present?
+        return "#{@resource.base_field_translation_key}.#{@id}" if @resource.try(:base_field_translation_key)
 
         "avo.field_translations.#{@id}"
       end
--- a/lib/avo/fields/base_field.rb
+++ b/lib/avo/fields/base_field.rb
@@ -93,7 +93,11 @@ module Avo
       end
 
       def translation_key
-        @translation_key || "avo.field_translations.#{@id}"
+        return @translation_key if @translation_key.present?
+
+        return "#{@resource.base_field_translation_key}.#{@id}" if @resource.base_field_translation_key.present?
+
+        "avo.field_translations.#{@id}"
       end
 
       def translated_name(default:)
--- a/lib/generators/avo/templates/locales/avo.en.yml
+++ b/lib/generators/avo/templates/locales/avo.en.yml
@@ -47,6 +47,11 @@ en:
         one: person
         other: people
         zero: people
+      project:
+        status:
+          one: Status Nested
+          other: Status Nesteds
+          zero: Status Nesteds
     filter_by: Filter by
     filters: Filters
     go_back: Go back
--- a/lib/generators/avo/templates/locales/avo.en.yml
+++ b/lib/generators/avo/templates/locales/avo.en.yml
@@ -47,11 +47,6 @@ en:
         one: person
         other: people
         zero: people
-      project:
-        status:
-          one: Status Nested
-          other: Status Nesteds
-          zero: Status Nesteds
     filter_by: Filter by
     filters: Filters
     go_back: Go back
--- a/lib/generators/avo/templates/locales/avo.en.yml
+++ b/lib/generators/avo/templates/locales/avo.en.yml
@@ -47,6 +47,12 @@ en:
         one: person
         other: people
         zero: people
+      status: Status
+      project:
+        status:
+          zero: Status Nesteds
+          other: Status Nesteds
+          one: Status Nested
     filter_by: Filter by
     filters: Filters
     go_back: Go back
--- a/lib/generators/avo/templates/locales/avo.en.yml
+++ b/lib/generators/avo/templates/locales/avo.en.yml
@@ -52,7 +52,6 @@ en:
           one: Status Nested
           other: Status Nesteds
           zero: Status Nesteds
-      status: Status
     filter_by: Filter by
     filters: Filters
     go_back: Go back
--- a/lib/generators/avo/templates/locales/avo.en.yml
+++ b/lib/generators/avo/templates/locales/avo.en.yml
@@ -47,12 +47,12 @@ en:
         one: person
         other: people
         zero: people
-      status: Status
       project:
         status:
-          zero: Status Nesteds
-          other: Status Nesteds
           one: Status Nested
+          other: Status Nesteds
+          zero: Status Nesteds
+      status: Status
     filter_by: Filter by
     filters: Filters
     go_back: Go back
--- a/config/i18n-tasks.yml
+++ b/config/i18n-tasks.yml
@@ -119,6 +119,7 @@ ignore_missing:
   - 'avo.resource_translations.user'
   - 'avo.resource_translations.team_members'
   - 'avo.x_items_more'
+  - 'avo.field_translations.project.status'
 
 # - 'errors.messages.{accepted,blank,invalid,too_short,too_long}'
 # - '{devise,simple_form}.*'
--- a/spec/dummy/app/avo/resources/project.rb
+++ b/spec/dummy/app/avo/resources/project.rb
@@ -1,6 +1,8 @@
 class Avo::Resources::Project < Avo::BaseResource
   self.title = :name
 
+  self.base_field_translation_key = "avo.field_translations.project"
+
   self.search = {
     query: -> {
       query.ransack(id_eq: params[:q], name_cont: params[:q], country_cont: params[:q], m: "or").result(distinct: false)
--- a/spec/dummy/app/avo/resources/project.rb
+++ b/spec/dummy/app/avo/resources/project.rb
@@ -1,8 +1,6 @@
 class Avo::Resources::Project < Avo::BaseResource
   self.title = :name
 
-  self.base_field_translation_key = "avo.field_translations.project"
-
   self.search = {
     query: -> {
       query.ransack(id_eq: params[:q], name_cont: params[:q], country_cont: params[:q], m: "or").result(distinct: false)
--- a/spec/system/avo/base_field_spec.rb
+++ b/spec/system/avo/base_field_spec.rb
@@ -1,7 +1,7 @@
 require "rails_helper"
 
 RSpec.feature "Base Fields", type: :system do
-  let(:project) { create :project, :with_files }
+  let!(:project) { create :project }
 
   after :each do
     Avo::Resources::Project.restore_items_from_backup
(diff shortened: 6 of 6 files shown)


djalmaaraujo:
@adrianthedev @Paul-Bob I think the tests should be in a unit-test matter, not in a browser related test, but I could not find a better example. Feel free to guide me on the best alternative.

lib/generators/avo/templates/locales/avo.en.yml
```diff
@@ -47,6 +47,11 @@ en:
         one: person
         other: people
         zero: people
+      project:
```
djalmaaraujo, line 50: If I could find a way to inject this during the test runtime, we don't need to update all locale files. The good thing about updating the locale is that is a clear example to others.

lib/avo/base_resource.rb
```diff
@@ -48,6 +48,7 @@ def current_user
     class_attribute :includes, default: []
     class_attribute :authorization_policy
     class_attribute :translation_key
+    class_attribute :base_field_translation_key
```
Paul-Bob, line 51: Would be possible to add a default to the resource name? Ex `Avo::Resources::Project` to have as default `base_field_translation_key == "avo.field_translations.project"`

When fetching the translation can we check something like (pseudo code) `if "avo.field_translations.project.some_field".exists_translation? else ...`

This way would only be necessary to add the `base_field_translation_key` when want a different key. It would work as default if there are nested fields under a known resource scope.

lib/generators/avo/templates/locales/avo.en.yml
Paul-Bob: IMO we can keep this only on the `avo.en.yml` and use it as example and as test

djalmaaraujo: @Paul-Bob How to avoid the I18n tests failures? Because the CI complains if I don't add to the other ymls.

Paul-Bob: `"i18n-tasks"` should have an configuration to ignore some translation keys

lib/avo/base_resource.rb
djalmaaraujo: @Paul-Bob It's a good idea. So based on your suggestion, the "on/off" of this feature would be having the real translation in the yml file, correct?

If yes, we would never have the "missing translation key" error that rails or avo throws for a missing key. Is this what we want here? Am I missing something?

lib/generators/avo/templates/locales/avo.en.yml
Paul-Bob: I think is the `ignore_missing:` list on `config/i18n-tasks.yml`

djalmaaraujo: image

lib/avo/base_resource.rb
Paul-Bob: Your concern makes sense and kinda invalidates my suggestion, let's keep it simple and keep the current implementation since we want to have the "missing translation key" error raised

lib/generators/avo/templates/locales/avo.en.yml
Paul-Bob: Adding `  - 'avo.field_translations.project.status'
` to `L122` in `config/i18n-tasks.yml` don't fix it?

djalmaaraujo: @Paul-Bob I found a way to inject only for the spec: aee1e60825f511f7911f9f92eb3ed9e59612c757. I hope you approve. It does not have to mess with the locale files.

lib/avo/base_resource.rb
djalmaaraujo: @Paul-Bob I am not 100% sure that this will happen, but since you have a better understanding of the code-base and you agreed, that's great. :)

lib/generators/avo/templates/locales/avo.en.yml
djalmaaraujo: @Paul-Bob Oh, yes, I can do that: ignore_missing property, I just saw you comment. let me know what approach you prefer, the inject or the `ignore_missing`. I will adjust.

djalmaaraujo: @Paul-Bob Check again 👍

Paul-Bob: Looks great @djalmaaraujo thank you!

djalmaaraujo: @Paul-Bob What are the next steps?

Paul-Bob: @djalmaaraujo we need to review and locally test it before merging. Thank you for the contribution!

lib/avo/fields/base_field.rb
```diff
@@ -93,7 +93,11 @@ def initialize(id, **args, &block)
       end
 
       def translation_key
-        @translation_key || "avo.field_translations.#{@id}"
+        return @translation_key if @translation_key.present?
+
+        return "#{@resource.base_field_translation_key}.#{@id}" if @resource.try(:base_field_translation_key)
+
+        "avo.field_translations.#{@id}"
```
owalmirneto, line 100: In a palliative way and trying to avoid DRY I used it this way:

```yml
pt-BR:
  activerecord:
    models:
      user: &user_names
        zero: "Usuários"
        one: "Usuário"
        other: "Usuários"
    attributes:
      user: &user_fields
        name: "Nome"
        email: "E-mail"
        password: "Senha"
  avo:
    resource_translations:
      <<: *user_names
    field_translations:
      <<: *user_fields
```

however I think you should consider translation from model, maybe like this:


```suggestion
        return @translation_key if @translation_key.present?

        return "activerecord.attributes.#{@resource.record.model_name.singular}.#{@id}" if @resource.record.respond_to?(@id)

        "avo.field_translations.#{@id}"
```

If you accept this suggestion maybe you should remove `class_attribute :base_field_translation_key` at `lib/avo/base_resource.rb:51`

---

PS: I know it'll change the standard, but it'll follow the rails conversion and best of all, we wouldn't need to rewrite the translation of the models (For who are migrating from other projects)

adrianthedev: Hey guys, let me pitch in with an opinion.

I'm going to preface this by saying that I did not know about Rails' way of translating models (`en.active_record.models.user`) when I first wrote the i18n part for Avo.

But, after a while of spending time with Avo, I think it's better that we didn't use that as a resource is not always a one-to-one representative.
One can use on model with multiple resources, thus needing multiple translations for something that active record only has once.

I do see that if you already translated the `activerecord` model, you wouldn't want to translate it again.
But there is a solution for that. 
You could se the [`translation_key`](https://docs.avohq.io/3.0/localization.html#localizing-resources) on a resource with the translation key that you already have `self.translation_key = `activerecord.models.user`.
Avo will use that to translate the resource.

adrianthedev (changes requested):
@djalmaaraujo, amazing work! Thank you for your contribution.

How about using `field_translation_keys` instead of `base_field_translation_key`?
I think it would fall in line with what we have now `transaltion_key`.

Also, I think we should provide a different key for our example.
Right now it's `field_translations.project.status`. This might be in conflict if there were a `project` field somewhere.
It's no biggie, but showing good defaults is a great tactic.

Otherwise, great PR! 
I have noticed the need for this while building Avo apps myself.

lib/avo/fields/base_field.rb
djalmaaraujo: Hello. When you say translate the resource, you are talking about the resource name. If you intend to translate fields, you need to repeat translation_key a lot. The issue here is that fields are in the root of fields_translations. This is very different from the community standards and Avo is here to simplify things and speed up development, right?

This is the pitch here. 😇

djalmaaraujo:
The name you suggested is great. I don't understand the issue with the project you mentioned. You can have both project and project status. The idea of the example is that you would nest project as a "group/model".

I will do one or two tests before asking for another review. 🤙

lib/avo/fields/base_field.rb
adrianthedev: Ok. I think I got it.

The pitch here is "give me a setting that will make Avo i18n behave like Rails' so I don't have to manually declare the translation keys for all my fields".

Am I correct with that statement?

adrianthedev:
The problem with `field_translations.project.status ` is that someone might have the `project` field and that will be in conflict with the proposed naming scheme.

djalmaaraujo:
> The problem with `field_translations.project.status ` is that someone might have the `project` field and that will be in conflict with the proposed naming scheme.

I might not get this right, but isnˋt the conflict ok? For example, if you have a global field translation for project, when you are on the specific resource, it will be overwritten by the specific, no? Again, I may be not getting what you are saying.

adrianthedev:
The conflict is not an override from global to specific.
The docs: https://docs.avohq.io/3.0/localization.html#localizing-fields

Let me show you the way the key is used and where the conflict might appear:

```yml
# How it is now
es:
  avo:
    field_translations: # the field translations holds the fields
      file: # field name
        zero: 'archivos'
        one: 'archivo'
        other: 'archivos'

# the override will break that
es:
  avo:
    field_translations:
      project: # this key is not a field anymore, but refers to a resource
        file:
          zero: 'archivos'
          one: 'archivo'
          other: 'archivos'
      project: # if this is a field it will clash with the `project` specified above
        zero: 'archivos'
        one: 'archivo'
        other: 'archivos'

# Proposed resolution
es:
  avo:
    resource_translations: # we are in the resource translation key, not `field_translation`
      project: # the resource
        zero: 'usuarios' # usual zero
        one: 'usuario' # usual one
        other: 'usuarios' # usual other
        fields: # add specific resource overrides here
          file:
            zero: 'archivos'
            one: 'archivo'
            other: 'archivos'
```
I propose using the `resource_translations.RESOURCE_NAME.fields.FIELD_NAME` key for resource-specific overrides

adrianthedev:
@djalmaaraujo, I'd so love to see this go past the finish line! 
It solves a painful problem.

PR #2189, opened 2023-12-18
fix: applied filters when filters are dynamic

Paul-Bob:
# Description

Since we removed the rescue on record to catch relevant errors we should ensure that `params[:filters]` is a string before trying to decode it. When using dynamic filters `params[:filters]` is a Hash

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/lib/avo/filters/base_filter.rb
+++ b/lib/avo/filters/base_filter.rb
@@ -65,7 +65,11 @@ module Avo
 
       # Fetch the applied filters from the params
       def applied_filters
+        # Return empty hash if no filters are present
         return {} if (filters_from_params = params[PARAM_KEY]).blank?
+
+        # Return empty hash if the filters are not a string, decode_filters method expects a Base64 encoded string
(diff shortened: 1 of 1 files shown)


lib/avo/filters/base_filter.rb
```diff
@@ -66,6 +66,7 @@ def applied_or_default_value(applied_filters)
       # Fetch the applied filters from the params
       def applied_filters
         return {} if (filters_from_params = params[PARAM_KEY]).blank?
+        return {} if !filters_from_params.is_a?(String)
```
adrianthedev, line 69: Let's add a comment on why this is here.
And make this the default pattern from now on.

adrianthedev (changes requested):
Do we need tests for this?

Paul-Bob:
> Do we need tests for this?

Not for this particular error. We should implement tests for usage of simultaneous dynamic and regular filters when working on this issue https://github.com/avo-hq/avo-3/issues/472

PR #2269, opened 2023-12-20
fix: actions download

### LINKED ISSUE #2184: Fix `ActionController::UnknownFormat` on download actions
### Describe the bug
When early returning an error or something on an action that have `  self.may_download_file = true` an `ActionController::UnknownFormat` error is raised.

Action code:
```ruby
class Avo::Actions::ExportCsv < Avo::BaseAction
  self.name = "Export csv"
  self.no_confirmation = false
  self.may_download_file = true

  def handle(**args)
    return error "Error"
  end
end

```

Paul-Bob:
# Description

Fixes #2184 

- `may_download_file` become deprecated and is no longer being used, can safely be removed from all actions.
- Fixes early error messages return on download actions
- Fixes keep modal open feature on download actions
- Refactors the way that actions download works
  - Use turbo stream to a custom stream action that triggers the download using [FileSaver.js lib](https://www.npmjs.com/package/file-saver)

# Checklist:

  - https://github.com/avo-hq/avodocs/pull/154

## Manual review steps

1. Step 1
1. Step 2


app/javascript/js/controllers/modal_controller.js
```diff
@@ -1,20 +0,0 @@
-import { Controller } from '@hotwired/stimulus'
-
-export default class extends Controller {
-  static targets = ['modal']
-
-  close() {
-    this.modalTarget.remove()
-
-    document.dispatchEvent(new Event('actions-modal:close'))
-  }
```
Paul-Bob, line 10: this still being used

Paul-Bob: add tests for it...

app/views/avo/actions/download.turbo_stream.erb
```diff
@@ -0,0 +1,16 @@
+<%# Append a div that triggers the download_controller.js %>
+<%= turbo_stream.append "actions_show" do %>
+  <%= content_tag :div,
+    data: {
+      controller: "download",
+      content: @response[:path],
+      filename: @response[:filename],
+    } do %>
+  <% end %>s
```
Paul-Bob, line 9:
```suggestion
  <% end %>
```

adrianthedev (approved):
If we have tests supporting this use case, then it's good to merge.

Thank you!

Paul-Bob:
> If we have tests supporting this use case, then it's good to merge.
> 
> Thank you!

We have several tests around actions, download, keeping modal open, alerts, etc... Are all being already tested.

PR #2279, opened 2023-12-21
fix: invalid date_time

### LINKED ISSUE #2261: Invalid DateTime
### Describe the bug
A date field is sometimes displayed as Invalid DateTime.

See Loom: https://www.loom.com/share/bfab89cd438648a2a8f174dc28c5cb0d?sid=2e5b3208-0019-4bc7-86b2-4f5e0ff46310

### System configuration
**Avo version**:
3.1.7

**Rails version**:
7.1.2

**Ruby version**:
3.2.2

**License type**:

 - [ ] Community
 - [ ] Pro
 - [x] Advanced

 - [x] No

#### Impact

 - [x] High impact (It makes my app un-usable.)

#### Urgency

 - [x] High urgency (I can't continue development without it.)

Paul-Bob:
# Description

Fields of type `has_one` was eager loading on tabs.

This was causing the `date_time` field's related JS to broke when the field panel was loading on a hidden tab, resulting on the string `Invalid DateTime`.

More specifically the `this.initialValue` on the date field stimulus controller was getting some weird spaces around the value when the field was rendering hidden, we also trim that value now. 

Fixes #2261

# Checklist:

## Manual review steps

1. Step 1
1. Step 2


lib/avo/fields/concerns/frame_loading.rb
```diff
@@ -0,0 +1,19 @@
...
+
+        def turbo_frame_loading = kwargs[:turbo_frame_loading]
+
+        def loading
+          turbo_frame_loading || params[:turbo_frame_loading] || "eager"
+        end
+      end
+    end
+  end
+end
```
Paul-Bob, line 16:
```suggestion
end

```

lib/avo/fields/concerns/frame_loading.rb
```diff
@@ -0,0 +1,19 @@
...
+
+
+
```
Paul-Bob, line 19:
```suggestion
end

```

adrianthedev (changes requested):
Could you please add some tests and a scenario where one can try this?

Paul-Bob:
> Could you please add some tests and a scenario where one can try this?

Done, the bug was happening when a `has_one` field was rendering on a secondary tab. Added the comment on user resource that contains a date time field to simulate the bug.

Paul-Bob:
The initial value on JS has some white spaces around it when the field was rendering hidden. We trim that value now

app/components/avo/fields/has_one_field/show_component.html.erb
```diff
@@ -1,5 +1,5 @@
 <% if @field.value %>
-  <turbo-frame id="<%= @field.turbo_frame %>" src="<%= @field.frame_url %>" target="_top" class="block">
+  <turbo-frame id="<%= @field.turbo_frame %>" src="<%= @field.frame_url %>"  loading=<%= loading %> target="_top" class="block">
```
adrianthedev, line 2:
```suggestion
  <turbo-frame id="<%= @field.turbo_frame %>" src="<%= @field.frame_url %>"  loading=<%= turbo_frame_loading %> target="_top" class="block">
```

lib/avo/fields/concerns/frame_loading.rb
```diff
@@ -0,0 +1,16 @@
...
+  module Fields
+    module Concerns
+      module FrameLoading
+        extend ActiveSupport::Concern
+
+        def turbo_frame_loading = kwargs[:turbo_frame_loading]
+
+        def loading
+          turbo_frame_loading || params[:turbo_frame_loading] || "eager"
+        end
```
adrianthedev, line 11:
```suggestion
        def turbo_frame_loading
          kwargs[:turbo_frame_loading] || params[:turbo_frame_loading] || "eager"
        end
```

lib/avo/fields/concerns/frame_loading.rb
```diff
@@ -0,0 +1,14 @@
+module Avo
+  module Fields
+    module Concerns
+      module FrameLoading
+        extend ActiveSupport::Concern
+
+        def turbo_frame_loading
```
adrianthedev, line 7: Do you think this change will impact any other components @Paul-Bob?

adrianthedev (approved):

lib/avo/fields/concerns/frame_loading.rb
Paul-Bob: This code was extracted from `Avo::Fields::HasManyField::ShowComponent` to `Avo::Fields::Concerns::FrameLoading`. 

`Avo::Fields::HasManyField::ShowComponent` includes now `Avo::Fields::Concerns::FrameLoading` so any component that inherit from it and the component itself still have access to the code.

The extraction was made just to reuse the same code inside `Avo::Fields::HasOneField::ShowComponent`.

Any other unrelated components are not being impacted.

Paul-Bob: @adrianthedev I misunderstood your question....

Yes, we should also apply this change on `app/components/avo/fields/has_many_field/show_component.html.erb`

PR #2297, opened 2023-12-28
fix: secret key base warning

### LINKED ISSUE #2090: Fix `Rails.application.secrets.secret_key_base`
`Rails.application.secrets.secret_key_base` will be removed in rails `7.2`. We should assure that our `lib/avo/services/encryption_service.rb` do not call `Rails.application.secrets.secret_key_base` when rails version `>= 7.2`.

Also make sure that all avo packages use `ENV["SECRET_KEY_BASE"] || Rails.application.credentials.secret_key_base` to avoid warning logs.
```[tasklist]
### Tasks
- [x] Add `>= 7.2` condition before calling `Rails.application.secrets.secret_key_base`
- [x] Ensure `ENV["SECRET_KEY_BASE"] || Rails.application.credentials.secret_key_base` usage on all avo packages spec.
```

Paul-Bob:
# Description

Fixes #2090

- Only fetch secret key from `Rails.application.secrets.secret_key_base` if rails version is less than `7.2.0`
- Raise custom error when not on production
- Keep same behavior on production

# Checklist:

### DIFF
--- a/spec/dummy/config/boot.rb
+++ b/spec/dummy/config/boot.rb
@@ -3,6 +3,8 @@ ENV["BUNDLE_GEMFILE"] ||= File.expand_path("../../../Gemfile", __dir__)
 
 ENV["AVO_IN_DEVELOPMENT"] = "1"
 
+ENV["SECRET_KEY_BASE"] = "130b8d0a74b5b73bfb2d0505c3de8250"
+
 require "bundler/setup" if File.exist?(ENV["BUNDLE_GEMFILE"])
 require "bootsnap/setup" unless ENV["CI"] # Speed up boot time by caching expensive operations.
 $LOAD_PATH.unshift File.expand_path("../../../lib", __dir__)
--- a/lib/avo/services/encryption_service.rb
+++ b/lib/avo/services/encryption_service.rb
@@ -29,14 +29,27 @@ module Avo
 
       def encryption_key
         secret_key_base[0..31]
-      rescue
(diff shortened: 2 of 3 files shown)


spec/dummy/config/boot.rb
```diff
@@ -3,6 +3,8 @@
 
 ENV["AVO_IN_DEVELOPMENT"] = "1"
 
+ENV["SECRET_KEY_BASE"] = "130b8d0a74b5b73bfb2d0505c3de8250"
```
adrianthedev, line 6: Please remind me, why do we need the `ENV["SECRET_KEY_BASE"]`?

adrianthedev (approved):
Nice! GG!

spec/dummy/config/boot.rb
Paul-Bob: For features that uses the encryption service to work, for example select all that encrypts the `@query`

adrianthedev: Oh, yeah. Makes sense. I was thinking about something else.
GOod to merge!

Paul-Bob: Without this variable that features would not work properly on tests or dev. Another solution would be to configure it on development and test credential files.

Anyway `130b8d0a74b5b73bfb2d0505c3de8250` is just a random string, it can be anything. What is important is to use on decryption the same string used on encryption

PR #2299, opened 2023-12-29
fix: default overriding value when creation fails

### LINKED ISSUE #1862: Setting a default value on a field makes it reset when validation errors occur
### Describe the bug

When using a "text" field, setting a default value like so:

```ruby
field :name, as: :text, default: "test"
```

makes the field reset when a validation error occur. (I haven't tested on other field types)

### Steps to Reproduce
1. Add a text field with a default value
2. Create a new record
3. Modify the default value on the record creation page
4. Save and make sure a validation error happens. 

Expected:
The field retains its new value

Actual:
The field resets to its default value

### System configuration
**Avo version**:
2.37.0

**Rails version**:
7.0.4

**Ruby version**:
3.0.4

**License type**:
 - [ ] Community
 - [x] Pro

 - [x] No

https://github.com/avo-hq/avo/assets/377682/58a9708f-b17d-4c67-82b8-ddeab517c929

#### Impact

#### Urgency

Paul-Bob:
# Description

Fixes #1862 

The field's default value should not be applied when the record already have a value for that particular `field.id` 

# Checklist:

### DIFF
--- a/app/components/avo/items/switcher_component.rb
+++ b/app/components/avo/items/switcher_component.rb
@@ -56,7 +56,8 @@ class Avo::Items::SwitcherComponent < Avo::BaseComponent
     Avo::TabGroupComponent.new(
       resource: @resource,
       group: item.hydrate(view: view),
-      index: index, params: params,
+      index: index,
+      params: params,
       form: form,
       view: view
     )


app/components/avo/items/switcher_component.rb
```diff
@@ -53,12 +53,29 @@ def in_reflection?
   end
 
   def tab_group_component
-    Avo::TabGroupComponent.new resource: @resource, group: item.hydrate(view: view), index: index, params: params, form: form, view: view
+    Avo::TabGroupComponent.new(
+      resource: @resource,
+      group: item.hydrate(view: view),
+      index: index, params: params,
```
adrianthedev, line 59:
```suggestion
      index: index,
      params: params,
```

app/components/avo/items/switcher_component.rb
```diff
@@ -53,12 +53,29 @@ def in_reflection?
...
+      form: form,
+      view: view
+    )
   end
 
   def field_component
-    final_item = item.dup.hydrate(resource: @resource, record: @resource.record, user: resource.user, view: view)
-    final_item.component_for_view(@view).new(field: final_item, resource: @resource, index: index, form: form, turbo_frame_loading: :lazy, **@field_component_extra_args)
+    item.hydrate(
```
adrianthedev, line 66: Are you sure that the `dup` wasn't there for a reason?
When I add `dup`, it stands for something.

app/components/avo/items/switcher_component.rb
```diff
@@ -53,12 +53,29 @@ def in_reflection?
...
+      resource: @resource,
+      record: @resource.record,
+      user: resource.user,
+      view: view,
+    ).component_for_view(@view).new(
+      field: item,
```
adrianthedev, line 72: again, in the past code we were sending the `dup` item back.

adrianthedev (changes requested):
I left a few comments about duplicating an object.
Please tell me if I'm wrong about that.

app/components/avo/items/switcher_component.rb
Paul-Bob: Old code:
```ruby
final_item = item.dup.hydrate(resource: @resource, record: @resource.record, user: resource.user, view: view)
final_item.component_for_view(@view).new(field: final_item, resource: @resource, index: index, form: form, turbo_frame_loading: :lazy, **@field_component_extra_args)
```

The "old" code was sending the `component_for_view` new instance back.

This would be useful if we want to keep 1 "dehydrated" item (called `item`) and 1 hydrated item (called `final_item`)

Maybe when this was implement there were a necessity to keep 1 "dehydrated" item, right now, from my analysis there is no need to keep the `item` "dehydrated"  and it may have some little impact on the performance since each field instance is getting duplicated.

Paul-Bob:
> I left a few comments about duplicating an object. Please tell me if I'm wrong about that.

I made an careful analysis and I think the duplication is unnecessary, maybe it was necessary on first iteration.

app/components/avo/items/switcher_component.rb
adrianthedev: what I meant was that the `dup`ing would ensure that any changes that would be applied to the `dup`ed object would not seep into the original one that was being used somewhere else.
It's a different object in memory.

Paul-Bob: Reverted it just in case

adrianthedev (approved):

PR #2314, opened 2024-01-03
feature: delegate missing to view_context

### LINKED ISSUE #2300: Delegate missign methods to `view_context` in `ExecutionContext`

Paul-Bob:
# Description

Fixes #2300

Delegating missing methods to view_context on Avo::ExecutionContext allow to direclty use `@view_context` methods without explicitly call it:

`body: simple_format(record.description)`

VS

`body: view_context.simple_format(record.description)`

# Checklist:

### DIFF
--- a/lib/avo/execution_context.rb
+++ b/lib/avo/execution_context.rb
@@ -4,6 +4,8 @@ module Avo
 
     attr_accessor :target, :context, :params, :view_context, :current_user, :request, :include, :main_app, :avo, :locale
 
+    delegate_missing_to :@view_context
+
     def initialize(**args)
       # Extend the class with custom modules if required.
       if args[:include].present?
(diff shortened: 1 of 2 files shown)


lib/avo/execution_context.rb
```diff
@@ -4,6 +4,8 @@ class ExecutionContext
 
     attr_accessor :target, :context, :params, :view_context, :current_user, :request, :include, :main_app, :avo, :locale
 
+    delegate_missing_to :@view_context
```
adrianthedev, line 7: Isn't this cleaner? Is there a performance impact for doing it like this?

```suggestion
    delegate_missing_to :view_context
```

adrianthedev (approved):

lib/avo/execution_context.rb
Paul-Bob: I think there is a performance impact because `view_context` it's just:
```ruby
def view_context
  @view_context
end
```

So it's an extra unnecessary call

https://www.greyblake.com/blog/ruby-perfomance-tricks/#use-instance-variables

I'm surprised about this :point_up:  , never thought that's "about two times faster"

If is that much faster we should consider refactor and use the instance variable when available instead the reader method.

adrianthedev: Good to know! Let's keep that in mind!

PR #2323, opened 2024-01-04
feature: remove all scope ( V2 )

### LINKED ISSUE #2276: Add the ability to remove the "All" default scope
Let's just do `remove_scope_all` and leave `default_scope` for later.
```[tasklist]
### Tasks
- [x] Implement feature
- [ ] Add tests
- [ ] Make docs
```

Paul-Bob:
# Description

Fixes #2276

Check https://github.com/avo-hq/avo-advanced/pull/25 for details


lib/avo/loaders/loader.rb
```diff
@@ -1,7 +1,7 @@
 module Avo
   module Loaders
     class Loader
-      attr_accessor :bag
+      attr_accessor :bag, :remove_scope_all
```
adrianthedev, line 4: Let's extract this to a separate `ScopeLoader` class.

app/components/avo/views/resource_index_component.rb
```diff
@@ -164,7 +164,8 @@ def scopes_list
       resource: resource,
       turbo_frame: turbo_frame,
       parent_record: parent_record,
-      query: query
+      query: query,
+      remove_scope_all: resource.entity_loader(:scope).remove_scope_all
```
adrianthedev, line 168: Can't we pass the loader and have the component ask the loader if it should show the default scope or not?

I feel that adding arbitrary options to it every time we change the API is not the way to go.

adrianthedev (changes requested):

app/components/avo/views/resource_index_component.rb
Paul-Bob: We can and it's better for scalability

PR #2332, opened 2024-01-06
feature: better sidebar scroll positioning

adrianthedev:
# Description

Fixes an issue where the sidebar position would not be remembered between visits.
We also scroll to the first active sidebar item when it's not in view.

# Checklist:

## Screenshots

https://github.com/avo-hq/avo/assets/1334409/d7879ddb-544b-4f5c-ba46-ee43517e706f

## Manual review steps

1. Make the browser window shorter in height so the sidebar menu gets scrollable
1. select a menu item that below the fold
2. navigate to a different page
3. observe how the scroll stays the same as before the navigation

#### Scroll to the first hidden active item from the sidebar

1. do the same and select a menu item below the fold
2. refresh the page
3. observe how the browser automatically scrolls that selected item into view


Paul-Bob (approved):
Maybe we should use _getters_ to improve readability for `document.querySelector('.avo-sidebar .mac-styled-scrollbar')` and `document.querySelector('.avo-sidebar .mac-styled-scrollbar a.active')`

Otherwise GG! This should improve UX by far on apps with a large menu!

PR #2362, opened 2024-01-14
Set z-index for navbar to the highest value, equal only to flash alerts

bryszard:
# Description

The problem is that our warning from `_custom_tools_alert.html.erb` is currently having an equal precedence (z-index = 50) as the navbar (z-index = 50) which results in this:

image

I'm setting the z-index of the navbar to 100, which is the same value as flash alerts and modals currently have. Meaning that only those components are allowed to be displayed over the navbar.

After the fix:

image

# Checklist:

## Manual review steps

1. Run the dummy app and scroll a bit


Paul-Bob (approved):
Hey @bryszard! Thank you for fixing this!

adrianthedev:
Thanks @bryszard.
It's annoying but we'll end up moving that alert to somewhere else.

Issue #2014, opened 2023-11-06
highlight search results


yshmarov:
### Feature

use `highlight` to highlight search results: https://apidock.com/rails/v5.2.3/ActionView/Helpers/TextHelper/highlight

### Screenshots or screen recordings

current search results:
<img width="727" alt="Screenshot 2023-11-06 at 09 03 16" src="https://github.com/avo-hq/avo/assets/13472945/4808e951-5eb6-4202-9172-67c4fee64624">
desired state with highlighting:
<img width="594" alt="Screenshot 2023-11-06 at 09 03 03" src="https://github.com/avo-hq/avo/assets/13472945/617364cd-c0a8-4c63-a37e-3fa4908f6e80">

GajendraR:
i want to work on it

adrianthedev:
You got it @GajendraR!

As an approach, I would recommend returning the search results with the string highlighted (in the `search_controller.rb` and then adding a CSS rule to apply that styling.

Let me know if you get stuck or need more help navigating the codebase.

We appreciate your help.

PR #2370, opened 2024-01-15
Mark highlight search

gabrielgiroe1:
# Description

Fixes https://github.com/avo-hq/avo/issues/2014

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/app/controllers/avo/search_controller.rb
+++ b/app/controllers/avo/search_controller.rb
@@ -1,4 +1,4 @@
-require_dependency "avo/application_controller"
+require_dependency 'avo/application_controller'
 
 include ActionView::Helpers::TextHelper
 
@@ -11,8 +11,8 @@ module Avo
 
     def show
       render json: search_resources([resource])
-    rescue => error
-      render_search_error(error)
+    rescue StandardError => e
+      render_search_error(e)
     end
 
     private
@@ -56,28 +56,21 @@ module Avo
 
       header = resource.plural_name
 
-      if results_count > 0
-        header = "#{header} (#{results_count})"
-      end
+      header = "#{header} (#{results_count})" if results_count > 0
 
       result_object = {
         header: header,
-        help: resource.fetch_search(:help) || "",
-        results: highlight_search_results(results, params[:q]),
+        help: resource.fetch_search(:help) || '',
+        results: results.map do |result|
+                   result.transform_values do |value|
+                     highlight(value.to_s, params[:q])
+                   end
+                 end,
         count: results.count
       }
 
       [resource.name.pluralize.downcase, result_object]
     end
-
-    def highlight_search_results(results, search_term)
-      results.map do |result|
-        result.transform_values do |value|
-          highlight(value.to_s, search_term)
-        end
-      end
-    end
-
     # When searching in a `has_many` association and will scope out the records against the parent record.
     # This is also used when looking for `belongs_to` associations, and this method applies the parents `attach_scope` if present.
     def apply_scope(query)
@@ -131,8 +124,6 @@ module Avo
       end
     end
 
-    private
-
     def fetch_result_information(record, resource, item)
       {
         _id: record.id,
@@ -142,11 +133,11 @@ module Avo
     end
 
     def should_apply_has_many_scope?
-      params[:via_association] == "has_many" && @resource.class.search_query.present?
+      params[:via_association] == 'has_many' && @resource.class.search_query.present?
     end
 
     def should_apply_attach_scope?
-      params[:via_association] == "belongs_to" && attach_scope.present?
+      params[:via_association] == 'belongs_to' && attach_scope.present?
     end
 
     def should_apply_any_scope?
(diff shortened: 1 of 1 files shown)


app/controllers/avo/search_controller.rb
```diff
@@ -61,13 +63,21 @@ def search_resource(resource)
       result_object = {
         header: header,
         help: resource.fetch_search(:help) || "",
-        results: results,
+        results: highlight_search_results(results, params[:q]),
```
adrianthedev, line 66: Instead of applying the highlight here and doing another loop, can't we do it straight where we write the initial object?

https://github.com/avo-hq/avo/pull/2370/files#diff-1a43067095b17211ecd0b229a1ad7bc86cc97b3d16a7044d78a2a0455463f2c6R139

adrianthedev (changes requested):
I left a comment to improve performance a bit.

app/controllers/avo/search_controller.rb
gabrielgiroe1: Yes, we can do that. I will update it. Thanks

app/controllers/avo/search_controller.rb
```diff
@@ -9,8 +11,8 @@ class SearchController < ApplicationController
 
     def show
       render json: search_resources([resource])
-    rescue => error
-      render_search_error(error)
+    rescue StandardError => e
+      render_search_error(e)
     end
```
adrianthedev, line 16: I don't think this is related to our PR.

app/controllers/avo/search_controller.rb
```diff
@@ -121,8 +126,6 @@ def apply_search_metadata(records, avo_resource)
       end
     end
 
-    private
```
adrianthedev, line 124: is this required?

app/controllers/avo/search_controller.rb
```diff
@@ -60,14 +62,17 @@ def search_resource(resource)
...
+                   result.transform_values do |value|
+                     highlight(value.to_s, params[:q])
+                   end
+                 end,
         count: results.count
       }
 
       [resource.name.pluralize.downcase, result_object]
     end
-
```
adrianthedev, line 70: Is this required?

app/controllers/avo/search_controller.rb
```diff
@@ -60,14 +62,17 @@ def search_resource(resource)
 
       result_object = {
         header: header,
-        help: resource.fetch_search(:help) || "",
-        results: results,
+        help: resource.fetch_search(:help) || '',
+        results: results.map do |result|
+                   result.transform_values do |value|
+                     highlight(value.to_s, params[:q])
+                   end
```
adrianthedev, line 69: I meant to apply the highlight in `fetch_result_information` under the `_label` key.

This way we avoid doing another loop over the results.

adrianthedev (changes requested):

app/controllers/avo/search_controller.rb
gabrielgiroe1: No, I changed how it was before in Refactoring PR.

app/controllers/avo/search_controller.rb
Paul-Bob: My bad adding that on past commits, there is already a `private` scope on L16 on that file. This one can be safely removed

app/controllers/avo/search_controller.rb
```diff
@@ -1,5 +1,7 @@
 require_dependency "avo/application_controller"
 
+include ActionView::Helpers::TextHelper
```
Paul-Bob, line 3: We should add this include inside `Avo::SearchController`.

Context: https://stackoverflow.com/a/5161484/9067704

app/controllers/avo/search_controller.rb
gabrielgiroe1: I will change.

app/controllers/avo/search_controller.rb
gabrielgiroe1: I will update it. Thank you.

app/controllers/avo/search_controller.rb
```diff
@@ -3,14 +3,15 @@
...
 
     before_action :set_resource_name, only: :show
     before_action :set_resource, only: :show
 
     def show
       render json: search_resources([resource])
-    rescue => error
-      render_search_error(error)
+    rescue StandardError => e
+      render_search_error(e)
```
adrianthedev, line 14: Let's remove this please.

app/controllers/avo/search_controller.rb
```diff
@@ -60,7 +61,7 @@ def search_resource(resource)
 
       result_object = {
         header: header,
-        help: resource.fetch_search(:help) || "",
+        help: resource.fetch_search(:help) || '',
```
adrianthedev, line 64: And this please.

adrianthedev (approved):
Let's remove those two and then it's good to be merged.

Paul-Bob (approved):

app/controllers/avo/search_controller.rb
adrianthedev: Good spotting!

PR #2382, opened 2024-01-18
security: v2 CVE-2024-22191

Paul-Bob:
# Description

Addresses https://github.com/avo-hq/avo/security/advisories/GHSA-ghjv-mh6x-7q6h

Fixes a security issue where the output for the `key_value` field is not sanitized.

### DIFF
--- a/app/javascript/js/controllers/fields/key_value_controller.js
+++ b/app/javascript/js/controllers/fields/key_value_controller.js
@@ -1,4 +1,5 @@
 /* eslint-disable max-len */
+import * as DOMPurify from 'dompurify'
 import { Controller } from '@hotwired/stimulus'
 import { castBoolean } from '../../helpers/cast_boolean'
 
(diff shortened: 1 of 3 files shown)


tamaloa:
Hi, i just stumbled across here because my updated avo (2.47) is still reported as vulnerable. I assume it takes some time for the ruby-vulnerability-db to pick up the change but maybe the avo security-advisory should be updated to reflect that only `<= 2.46.0` is affected?

Screenshot from 2024-01-18 13-29-40

tamaloa:
First off: Thanks for backporting the fix - we are just not yet ready for 2 -> 3 upgrade :)

I am also a bit confused regarding what path a security-advisory takes until it ends up in the tool you are using (bundler-audit in our case which uses ruby-advisory-db which uses github-advisories). But it seems it worked out for this CVE :)

Regarding ruby-advisory-db: I noticed they have the next advisory which would affect the avo 2.x branch lined up (https://github.com/rubysec/ruby-advisory-db/pull/736) -> "Possible XSS in Action messages". This is taken from the linked github-advisory (https://github.com/advisories/GHSA-g8vp-2v5p-9qfh) which also flags avo 2.x as vulnerable (affected versions:` < 3.0.2`). But your own security-advisory (https://github.com/avo-hq/avo/security/advisories/GHSA-g8vp-2v5p-9qfh) only mentions one specific release (`3.0.0.pre12`) to be vulnerable.

So is the 2.x branch vulnerable to "Possible XSS in Action messages"? If not it would be great if the advisories were changed :)

Sorry this does not really belong into this PR but i could not find a better place to comment on this.

Issue #2347, opened 2024-01-09
Grid view: hide or customize poster image


sedubois:
### Feature
The [grid view](https://docs.avohq.io/3.0/grid-view.html#custom-style) allows to customize the CSS for the title and body, but not for the poster image itself. It results in an "avocado" picture being repeated on each and every grid item. An option to hide this default cover picture would be welcome. Additionally, the cover picture could be configured with custom CSS like for the grid element title and body. 

### Current workarounds
None known.

### Screenshots or screen recordings
![Screenshot 2024-01-09 at 11 46 31](https://github.com/avo-hq/avo/assets/4217871/8db99d4f-b2bc-41d7-a517-d67a83308d9e)

adrianthedev:
### Approach

I think it's not too difficult to retrofit this to the current DSL.

```ruby
self.grid_view = {
  card: -> do
    {
      title: record.name
    }
  end
}
```

If the returned object does not have the `cover_url` key, it will be omitted. the `.key? :cover_url` method can be used.

goamaral:
Hello, I am creating a Pull request to implement this but there is an undocumented workaround. The cover wrapper class and style can be customized. So you can hide the cover wrapper in this cases.

```
    html: -> do
      {
        cover: {
          index: {
            wrapper: {
              classes: "hidden"
            }
          }
        }
      }
    end
```

Paul-Bob:
Playing with it a bit and testing @goamaral suggestion I end up with this solution:

```ruby
self.grid_view = {
    card: -> do
      {
        cover_url:
          if record.cover_photo.attached?
            main_app.url_for(record.cover_photo.url)
          end,
        title: record.name,
        body: helpers.extract_excerpt(record.body)
      }
    end,
    html: -> do
      {
        cover: {
          index: {
            wrapper: {
              classes: record.cover_photo.attached? ? "" : "hidden"
            }
          }
        }
      }
    end
  }
```

![image](https://github.com/avo-hq/avo/assets/69730720/291674da-948a-4e5f-b335-538ffa02221a)

adrianthedev:
One can use these CSS rules to hide the image, collapse the container, and still keep the checkbox in view.

```css
/* remove the `[data-resource-name="posts"]` selector to have that applied globally on every resource */
/* change `posts` to a different resource that you need */
[data-component-name="avo/index/grid_item_component"][data-resource-name="posts"] > div:first-child {
  /* we need to override the pb-3/4 class */
[...]

PR #2393, opened 2024-01-20
Hide grid view cover if cover_url not defined (#2347)

goamaral:
# Description
At the moment the grid view cover handles two scenarios
- If cover_url is present, that url is used to display the cover image
- If not, a fallback image is used
We want to extend the second case distinguishing a not defined cover_url from a falsy value.

I noticed that the grid view cover empty state component was really similar to the code at link_to_cover.
So I took the initiative to merge them (the link_to_cover return did not have a wrapping div but I think this is fine 
because adding it does not affect anything visually).

Updated the avo docs to document how add custom styling to the cover. (https://github.com/avo-hq/avodocs/pull/165)

Fixes [# 2347](https://github.com/avo-hq/avo/issues/2347)

# Checklist:
- [X] I have made corresponding changes to the [documentation](https://github.com/avo-hq/avodocs)

## Screenshots & recording
image
image

## Manual review steps
1. Open Posts in grid view
2. Check if posts with image, show the image
3. Check if posts without image, show the fallback image (avocado)
4. Open Fish in grid view
5. Check if all posts have no cover


goamaral:
There is still a problem to solve. Since my solution was to not render the cover if the :cover_url was not defined, we lost the ability to select the item.

Other solution would be to early return render_cover unless @card.key?(:cover_url) but we would be left with
image
We can shorten that empty space, but because the checkbox only appears on hover we would have to leave an empty space above the title (case 1 kind of weird) or move the checkbox in this cases to the div where the title is but the checkbox would appear above the title (case 2)
case 1
image
case 2
image

Neither cases are good, any suggestions?

goamaral:
Ended up extending `item_selector_input` and moving the checkbox to the right
image
image

The changes I made to `item_selector_input` should be backwards compatible

adrianthedev:
Thanks for the PR, @goamaral, but I'm really leaning toward surfacing the workaround that you posted earlier.
I mean, it has the same result without any new code or changes for the user.

goamaral:
The only problem with the workaround is that the card becomes unselectable because the checkbox is the cover div.
If you think this is not that relevant, we can close this. But I think the docs should be updated at least.
https://github.com/avo-hq/avodocs/pull/165

adrianthedev:
One can use these CSS rules to hide the image, collapse the container, and still keep the checkbox in view.

```css
/* remove the `[data-resource-name="posts"]` selector to have that applied globally on every resource */
/* change `posts` to a different resource that you need */
[data-component-name="avo/index/grid_item_component"][data-resource-name="posts"] > div:first-child {
  /* we need to override the pb-3/4 class */
  padding-bottom: 2rem !important;
}

[data-component-name="avo/index/grid_item_component"][data-resource-name="posts"] > div:first-child > a {
  /* hide the link and image */
  display: none;
}
```

adrianthedev:
Closing for now. Feel free to reopen if it's needed.

PR #2415, opened 2024-01-26
fix: destroy when using forms and display fields API

### LINKED ISSUE #2403: def display_fields / form_fields do not cover detach
If you declare `display_fields' and `form_fields' like this

```ruby
def display_fields
 field :title as: :text
 field :users as: :has_and_belongs_to_many
end

def form_fields
 field :title, as: :text
 field :users as: :has_and_belongs_to_many
end
```

You cannot detach a `user` resource because you get the error: `use_resource` for NIL class. This may be because the DELETE request for the association does not have the field. To "fix" this, you need to declare the "fields" method only with the association, such as

```ruby
def fields
 field :users as: :has_and_belongs_to_many
end
```

With the "fields" method, the detach will work. You may need to check that a detach is also covered in "form_fields" so that you do not have to declare the "fields" method just for the detach. We are trying to make the code cleaner and easier to read by separating the different views into the corresponding field methods. It would be nice to not have the "fields" method for some resources and use the form/display fields completely.

Paul-Bob:
# Description

Fixes #2403 

The detach button request has been updated to no longer hydrate the resource using the `destroy` view, as it was identified as a problematic approach when using `def display_fields` or `def form_fields` API since those methods wouldn't get called. 

To address this, we have modified the process to include passing the current view as parameters during the detach request. This ensures that the appropriate view is properly hydrated into the resource.

# Checklist:

### Before
[before.webm](https://github.com/avo-hq/avo/assets/69730720/7696b801-fed5-4c8b-bdc4-dd69accbc923)

### After
[after.webm](https://github.com/avo-hq/avo/assets/69730720/4b12ce52-f9d0-4c3b-8a7c-cf1ee9a35cbb)

## Manual review steps

1. Set a resource with an association field using only `def display_fields` and `def form_fields` API
2. Visit resource's show view
3. Detach a record from the association field
4. Before was raising error, notice that now it works.

### DIFF
--- a/app/controllers/avo/application_controller.rb
+++ b/app/controllers/avo/application_controller.rb
@@ -119,7 +119,12 @@ module Avo
     def set_related_resource
       raise Avo::MissingResourceError.new(related_resource_name) if related_resource.nil?
 
-      @related_resource = related_resource.new(params: params, view: action_name.to_sym, user: _current_user, record: @related_record).detect_fields
+      @related_resource = related_resource.new(
+        params: params,
+        view: params[:view] || action_name.to_sym,
+        user: _current_user,
+        record: @related_record
+      ).detect_fields
     end
 
     def set_record
--- a/app/controllers/avo/application_controller.rb
+++ b/app/controllers/avo/application_controller.rb
@@ -123,6 +123,10 @@ module Avo
 
       # Get view from params unless actions is index or show or forms...
       # Else, for example for detach action we want the view from params to can fetch the correct fields
+      # This logic avoid the following scenario:
+      # When a has many field is rendered the action is index and params[:view] is show or edit but we want to
+      # keep @view as index for the related_resource
+      # Same do not happen with other actions except the list below.
       view = if action_view.in?([:index, :show, :new, :edit, :create])
         action_view
       else
--- a/app/controllers/avo/application_controller.rb
+++ b/app/controllers/avo/application_controller.rb
@@ -119,9 +119,19 @@ module Avo
     def set_related_resource
       raise Avo::MissingResourceError.new(related_resource_name) if related_resource.nil?
 
+      action_view = action_name.to_sym
+
+      # Get view from params unless actions is index or show or forms...
+      # Else, for example for detach action we want the view from params to can fetch the correct fields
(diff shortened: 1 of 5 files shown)


app/controllers/avo/application_controller.rb
```diff
@@ -119,7 +119,22 @@ def detect_fields
...
-      @related_resource = related_resource.new(params: params, view: action_name.to_sym, user: _current_user, record: @related_record).detect_fields
+      action_view = action_name.to_sym
+
+      # Get view from params unless actions is index or show or forms...
+      # Else, for example for detach action we want the view from params to can fetch the correct fields
+      view = if action_view.in?([:index, :show, :new, :edit, :create])
+        action_view
+      else
+        params[:view] || action_view
+      end
```
Paul-Bob, line 130: This logic avoid the following scenario:

When a has many field is rendered the action is `index` and `params[:view]` is `show` or `edit` but we want to keep `@view` as `index` for the `related_resource` 

Same do not happen with other actions except the list on L126

adrianthedev: Please add this as a comment in the code so we know why it's there.

adrianthedev (approved):

app/controllers/avo/application_controller.rb
```diff
@@ -119,7 +119,22 @@ def detect_fields
     def set_related_resource
       raise Avo::MissingResourceError.new(related_resource_name) if related_resource.nil?
 
-      @related_resource = related_resource.new(params: params, view: action_name.to_sym, user: _current_user, record: @related_record).detect_fields
+      action_view = action_name.to_sym
+
+      # Get view from params unless actions is index or show or forms...
+      # Else, for example for detach action we want the view from params to can fetch the correct fields
```
Paul-Bob, line 125:
```suggestion
      # Get view from params unless actions is index or show or forms...
      # Else, for example for detach action we want the view from params to can fetch the correct fields
      # This logic avoid the following scenario:
      # When a has many field is rendered the action is index and params[:view] is show or edit but we want to
      # keep @view as index for the related_resource
      # Same do not happen with other actions except the list below.
```

PR #2417, opened 2024-01-27
Use Rails.application.secret_key_base directly

segiddins:
This helps avoid the rails 7.1 deprecation warning about using `Rails.application.secrets` when no secret key base is explicitly configured, when `Rails.env.local? || ENV["SECRET_KEY_BASE_DUMMY"]`

# Description

Fixes # (issue)

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/lib/avo/services/encryption_service.rb
+++ b/lib/avo/services/encryption_service.rb
@@ -32,15 +32,11 @@ module Avo
       private
 
       def encryption_key
-        secret_key_base[0..31]
+        Rails.application.secret_key_base[0..31]
       rescue
         # This will fail the decryption process.
         # It's here only to keep Avo from crashing
         SecureRandom.random_bytes(32)
-      end
-
-      def secret_key_base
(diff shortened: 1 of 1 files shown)


adrianthedev:
It [seems](https://apidock.com/rails/v6.0.0/Rails/Application/secret_key_base#:~:text=In%20all%20other%20environments%2C%20we%20look%20for%20it%20first%20in%20ENV%5B%E2%80%9CSECRET_KEY_BASE%E2%80%9D%5D%2C%20then%20credentials.secret_key_base%2C%20and%20finally%20secrets.secret_key_base.%20For%20most%20applications%2C%20the%20correct%20place%20to%20store%20it%20is%20in%20the%20encrypted%20credentials%20file.) that `Rails.application.secret_key_base` does what we already did in the removed method.
I didn't know that.
Thanks @segiddins!

jpawlyn:
Would it be ok to have a similar PR against `main` to avoid the deprecation warning for Avo 3 ? If so happy to do it.

PR #2423, opened 2024-01-29
fix: searchable belongs_to field keep html tags on selected option

### LINKED ISSUE #2416: `<mark>` tag persists when selecting on a belongs association with searchable true
### Describe the bug

[mark_persist_bug.webm](https://github.com/avo-hq/avo/assets/69730720/d1d01ac5-6ea5-45df-867a-97d4af04f707)

## Approach
Remove the `<mark>` and `</mark>` from the string after selecting the option.

Paul-Bob:
# Description

Fixes #2416

Clear all HTML tags before updating the belongs_to searchable field with selected option.

# Checklist:

### Before
before

### After
after

### DIFF
--- a/app/javascript/js/controllers/search_controller.js
+++ b/app/javascript/js/controllers/search_controller.js
@@ -208,7 +208,7 @@ export default class extends Controller {
   handleOnSelect({ item }) {
     if (this.isBelongsToSearch) {
       this.updateFieldAttribute(this.hiddenIdTarget, 'value', item._id)
-      this.updateFieldAttribute(this.buttonTarget, 'value', item._label)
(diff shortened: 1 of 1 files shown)


adrianthedev:
What if the user wants to show some HTML on those entries?

Paul-Bob:
> What if the user wants to show some HTML on those entries?

Should we keep HTML attributes and applying them on the entry?

Or specifically target to remove only `<mark>` and `</mark>` ? 

Maybe someone want to keep even the marked version.

My idea to KSS was to remove all HTML marks and get back to it if someone present a real use case

adrianthedev (approved):

PR #2386, opened 2024-01-19
Reload icon

gabrielgiroe1:
# Description

Fixes # (issue)
https://github.com/avo-hq/avo/issues/2172

# Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the [documentation](https://github.com/avo-hq/avodocs)
- [ ] I have added tests that prove my fix is effective or that my feature works

## Manual review steps

1. Step 1
1. Step 2

Manual reviewer: please leave a comment with output from the test if that's the case.

PR #2427, opened 2024-01-29
✨ feat. reload table / relationships

RocKhalil:
# Description
I've added a button to reload the content of a specific frame.
For example, I'm doing updates via console, or the user did something on his end, and I was to reload this specific relationship or table without refreshing the whole page; the button will find the nearest frame and reload it.

# Checklist:

# Notes:
I know that you guys prefer using stimulus controllers for these type of things, but I was not able to make it run with a controller looking like this:

```js
import { Controller } from '@hotwired/stimulus'
  get parentTurboFrame() {
    return this.context.scope.element.closest('turbo-frame')
  }

  refresh() {
    if (!this.parentTurboFrame.src) return
    this.parentTurboFrame.reload()
  }
}
```

### DIFF
--- a/app/components/avo/resource_component.rb
+++ b/app/components/avo/resource_component.rb
@@ -282,6 +282,18 @@ class Avo::ResourceComponent < Avo::BaseComponent
     end
   end
 
+  def render_reload_button(control)
+    # return unless can_see_the_create_button?
+
+    a_button color: :primary,
+      style: :text,
+      type: :button,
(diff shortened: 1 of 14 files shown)


adrianthedev:
It's so cool you submitted this PR.

We were thinking about the same thing but a bit more integrated.
Please follow https://github.com/avo-hq/avo/pull/2386 for updates. It's close to merge.

Will close this one.
Thank you!

PR #2431, opened 2024-01-30
Stop file field from hiding filename

iainbeeston:
# Description

On main, even if you choose to display filenames, filenames are hidden for file fields on avo 3.x. This is happening because the parent flexbox is set to flex row and truncate css class was truncating the text down to 0 width.

To fix this I've made the parent flexbox use flex-col, which puts the filename underneath the image (as it was in avo 2.x) so it doesn't get truncated and can be seen.

Fixes # (issue)

# Checklist:

## Screenshots & recording

Before it looks like this:

Screenshot 2024-01-30 at 16 37 58

On this branch it looks like this:

Screenshot 2024-01-30 at 16 39 01

## Manual review steps

1. Find a resource with a files component and set `display_filename: true` is set to `true`
2. View it in the browser and confirm that the filename isn't visible
3. Switch to this branch
4. Confirm that the filename is visible now.


adrianthedev:
Thanks for digging into this @iainbeeston

PR #2438, opened 2024-02-01
feature: added a preview when editing/creating images, video or audio

maxencehenneron:
# Description

It adds a preview when adding an image/video/audio to a "file" field.

Note: a preview already exists when the record is created, but does not automatically update when you select another file. While my code works, I don't think it's the cleanest.

Some logic is duplicated between edit_component.html.erb (file field) and grid_item_component.html.erb.
grid_item_component references file_field_controller and I don't know if it's a bad practice. I also cannot think of a better way to do this as I don't use stimulus in my projects

# Checklist:

## Screenshots & recording

https://github.com/avo-hq/avo/assets/377682/f6d3cc4c-cfac-4595-b4a8-d967b52f122f

### DIFF
--- a/app/components/avo/fields/file_field/edit_component.html.erb
+++ b/app/components/avo/fields/file_field/edit_component.html.erb
@@ -1,20 +1,33 @@
 <%= field_wrapper **field_wrapper_args do %>
-  <% if @field.value.present? %>
+  <%= content_tag :div, data: {
+    controller: "file-field",
+  } do %>
     <div class="mb-2">
-      <%= render Avo::Fields::Common::Files::ViewType::GridItemComponent.new resource: @resource, field: @field %>
+      <% if @field.value.present? %>
+        <%= render Avo::Fields::Common::Files::ViewType::GridItemComponent.new resource: @resource, field: @field %>
+      <% elsif @field.is_image %>
+        <%= image_tag "", "data-file-field-target": "preview", class: "rounded-lg max-w-full" %>
+      <% elsif @field.is_audio %>
+        <%= audio_tag("", "data-file-field-target": "preview", controls: true, preload: false, class: 'w-full') %>
+      <% elsif @field.is_video %>
+        <%= video_tag("", "data-file-field-target": "preview", controls: true, preload: false, class: 'w-full') %>
+      <% end %>
     </div>
-  <% end %>
 
-  <% if can_upload_file? %>
-    <%= @form.file_field @field.id,
-      accept: @field.accept,
-      data: @field.get_html(:data, view: view, element: :input),
-      direct_upload: @field.direct_upload,
-      disabled: disabled?,
-      style: @field.get_html(:style, view: view, element: :input),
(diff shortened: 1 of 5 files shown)


app/components/avo/fields/file_field/edit_component.html.erb
```diff
@@ -1,20 +1,33 @@
...
     <div class="mb-2">
-      <%= render Avo::Fields::Common::Files::ViewType::GridItemComponent.new resource: @resource, field: @field %>
+      <% if @field.value.present? %>
+        <%= render Avo::Fields::Common::Files::ViewType::GridItemComponent.new resource: @resource, field: @field %>
+      <% elsif @field.is_image %>
+        <%= image_tag "", "data-file-field-target": "preview", class: "rounded-lg max-w-full" %>
+      <% elsif @field.is_audio %>
+        <%= audio_tag("", "data-file-field-target": "preview", controls: true, preload: false, class: 'w-full') %>
+      <% elsif @field.is_video %>
+        <%= video_tag("", "data-file-field-target": "preview", controls: true, preload: false, class: 'w-full') %>
```
adrianthedev, line 13: I don't think it's a good idea to have this duplication here.
The first time someone updates one tag here or in `GridItemComponent`, the other one will not be updated.

Can we try to use only `GridItemComponent`?

app/components/avo/fields/file_field/edit_component.html.erb
```diff
@@ -1,20 +1,33 @@
 <%= field_wrapper **field_wrapper_args do %>
-  <% if @field.value.present? %>
+  <%= content_tag :div, data: {
+    controller: "file-field",
+  } do %>
     <div class="mb-2">
```
adrianthedev, line 5: Can we add the `mb-2` class inside that parent `div` so we won't have two nested divs?

app/javascript/js/controllers/fields/file_field_controller.js
```diff
@@ -0,0 +1,13 @@
+import { Controller } from '@hotwired/stimulus'
+
+export default class extends Controller {
+  static targets = ['preview']
+
+  update(e) {
+    const reader = new FileReader()
+    reader.onload = () => {
+      this.previewTarget.src = reader.result
```
adrianthedev, line 9: Let's do a check to see if the target is present.

adrianthedev (changes requested):
This is a great start @maxencehenneron. Thanks!

I left a comment to remove a bit of duplication.
Usually, I'm not afraid of duplication, but this is not the right place IMHO. We'll forget it's in two places.

If you can help address that, it'll be pretty good to merge

adrianthedev:
@maxencehenneron are you able to move this past the finish line?

maxencehenneron:
> @maxencehenneron are you able to move this past the finish line?

Sorry I don't have much time to complete this, I don't see any simple way to achieve this without significant refactoring as we do not have a file yet when first creating the object

PR #2449, opened 2024-02-05
🐛 Bypass protocol when using full link

RocKhalil:
# Description
Adding `protocol: :link` to fields

Fixes # (issue)
Using a field with full url will never work unless we do a custom `as_html` ; adding `protocol: :link` should fix it

### DIFF
--- a/app/components/avo/fields/text_field/index_component.html.erb
+++ b/app/components/avo/fields/text_field/index_component.html.erb
@@ -2,7 +2,11 @@
   <% if @field.as_html %>
     <%== @field.value %>
   <% elsif @field.protocol.present? %>
-    <%= link_to @field.value, "#{@field.protocol}:#{@field.value}" %>
+    <% if @field.protocol.to_sym == :link %>
+      <%= link_to @field.value, @field.value %>
+    <% else %>
+      <%= link_to @field.value, "#{@field.protocol}:#{@field.value}" %>
+    <% end %>
   <% else %>
     <%= link_to_if @field.link_to_record, @field.value, resource_view_path %>
   <% end %>
--- a/app/components/avo/fields/text_field/show_component.html.erb
+++ b/app/components/avo/fields/text_field/show_component.html.erb
@@ -2,7 +2,11 @@
   <% if @field.as_html %>
     <%== @field.value %>
   <% elsif @field.protocol.present? %>
-    <%= link_to @field.value, "#{@field.protocol}:#{@field.value}" %>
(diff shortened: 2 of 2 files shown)


Paul-Bob (commented):
Hello @RocKhalil thanks for the PR!

We already has a similar discussion here https://github.com/avo-hq/avo/pull/1869

IMO `as_html: true` is much powerful and customizable than having a flag for a link.

This PR would work great for simple links, but when you want to add some data attributes or any `html_options` to your link you would fallback to the  `as_html: true` approach.

The following example is impossible to achieve with `protocol: :link`

`link_to "Visit Avo Site", "http://www.avo.cool/", data: { confirm: "Are you sure?" }`

IMHO this approach is for a very specific and simple use case and it is not scalable.

```ruby
field :user_link, as: :text, only_on: :show, name: 'Link', protocol: :link, readonly: true do
  record.user.link
end
```
`==`
```ruby
field :user_link, as: :text, only_on: :show, name: 'Link', html: true, readonly: true do
  link_to record.user.link, record.user.link
end
```

RocKhalil:
@Paul-Bob this makes total sense btw; I didn't try to use `link_to` ( for no reason ) and just tried it. it works better than what I had in mind

adrianthedev:
I think it's time for a `link` field.
We have plenty of workarounds for adding links, but none are bulletproof.
Didn't we use to have a PR somewhere @Paul-Bob?

Paul-Bob:
Only related PR  that I remember is this one https://github.com/avo-hq/avo/pull/1869

How would be the API for example for this link?

`link_to "Visit Avo Site", "http://www.avo.cool/", data: { confirm: "Are you sure?" }`

Possible API:

`field :something, as: :link, name: "Visit Avo Site", url:  "http://www.avo.cool/", html_options: { data: { confirm: "Are you sure?" } }`

What will the `link` field fix that can't be achieved with `html: true` option?

We need to have [`link_to`](https://apidock.com/rails/ActionView/Helpers/UrlHelper/link_to) helper arguments in consideration and make possible to use all arguments on the `link` field. This is the main reason that I think `html: true` is more powerful, we don't need to add support and maintain a interface that will use the `link_to`, it can be directly used.

adrianthedev:
Maybe instead of `url` we use `href`?

adrianthedev:
> What will the link field fix that can't be achieved with html: true option?

To me, it's just the easiest way to add something that most folks need.
For those who need more... they can use the text field with `as_html` or build their own custom field.

PR #2454, opened 2024-02-06
fix: fetch cards only when necessary

Paul-Bob:
# Description

Apply same logic that we use to fetch fields but on cards. We should fetch cards only when visiting a dashboard.

Related PR https://github.com/avo-hq/avo-dashboards/pull/31


adrianthedev (approved):
LGTM!

PR #2462, opened 2024-02-07
fix: Trix field disable_attachments option unexpected behaviour

olivierbuffon:
# Description
This PR comes in order to fix a critical regression introduced by #2429 which changed 
```ruby
@attachments_disabled = args[:attachments_disabled] || false
```
by
```ruby
@attachments_disabled = args[:attachments_disabled] || true
@attachments_disabled = false unless @attachment_key.present?
```

Unfortunately this created an unexpected behaviour making `@attachments_disabled` always defaulting to `true`. Here is why:
```ruby
irb(main):001> args = {}
=> {}
irb(main):002> args[:attachments_disabled] = false
=> false
irb(main):003> @attachments_disabled = args[:attachments_disabled] || true
=> true
```
This has the side effect of breaking the attachment feature by always disabling it even when specifying an attachment key or manually set the `attachments_disabled` option to `false`.

This PR comes with a simple fix extracting the logic in a private method and validating if the boolean value is set to `true` or `false`.

# Checklist:

## Manual review steps

There's  multiple ways to reproduce it:

- Add a `trix` field with an `attachment_key` and try to add an attachment to the content
- Add a `trix` field and explicitely set the `attachments_disabled` to true, the button to add an attachment will show up anyway

### DIFF
--- a/lib/avo/fields/trix_field.rb
+++ b/lib/avo/fields/trix_field.rb
@@ -15,8 +15,6 @@ module Avo
 
         @always_show = args[:always_show] || false
         @attachment_key = args[:attachment_key]
-        # If we don't have an attachment_key, we disable attachments.  There's no point in having
-        # attachments if we can't store them.
         @attachments_disabled = disable_attachments?(args)
         @hide_attachment_filename = args[:hide_attachment_filename] || false
         @hide_attachment_filesize = args[:hide_attachment_filesize] || false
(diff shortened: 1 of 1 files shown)


adrianthedev:
🤦‍♂️ Oh yeah... I didn't see that.

Thank you for the contribution and for digging into the problem @olivierbuffon.

olivierbuffon:
@adrianthedev No worries, my pleasure! Love the product by the way! Keep up the great job 👏!

Issue #2385, opened 2024-01-19
Download action without refresh lead to unexpected behavior 


Paul-Bob:
### Describe the bug

Download action without refresh lead to unexpected behavior.

From here -> https://discord.com/channels/740892036978442260/1197846489163112449/1197912362213195847

### Steps to Reproduce
1. Go to https://main.avodemo.com/avo/resources/projects
2. Select some records
3. Click on Actions -> Export csv
4. W8 download to complete 
5. Repeat step 3
6. Notice weird behavior where URL changes and background turns white

[action_double_download.webm](https://github.com/avo-hq/avo/assets/69730720/b8341243-0837-4c7a-b7c3-a8189030dcff)

mid:
Encountered the same issue.

Also, downloading the binary files (e.g., PNG images or PDF images) generated by Avo Actions no longer works after upgrading from 3.2 to 3.3 :( The downloaded files are corrupted.

PR #2465, opened 2024-02-08
chore: upgrade to Turbo 8

### LINKED ISSUE #2182: Allow arguments on action's `link_arguments` method
`link_arguments` should recieve `arguments:` param and return encrypted arguments within path.

Ex: 
`Avo::Actions::Dummy.link_arguments(resource: resource, arguments: { something: :here })`:
`avo/resources/fish/actions?action_id=Avo::Actions::Dummy&arguments=encrypted_string`

### LINKED ISSUE #2185: Refactor action arguments
Right now we generate path / arguments "all over the code". On action class, on components etc... Let's organize it and move all path / arguments generation into the action class.

We should have 2 methods for encoding / decoding ex:

```ruby
def encode_arguments(args)
  # Base64...
end
```

Same for decoding. This way each user can override and implement a custom encoding / decoding or even disable them.

We also should create all actions links from the `link_arguments` method instead of duplicating that code in several places.

`link_arguments` should receive arguments as hash and encode them using the mentioned `encode_arguments` this way we don't "force" users to use our encoding system when generating action links

adrianthedev:
# Description

Upgrading the core Turbo library

Also fixes #2185 and fixes #2182

## Related PRs
- https://github.com/avo-hq/avo-advanced/pull/28

# Checklist:


app/views/layouts/avo/application.html.erb
```diff
@@ -47,7 +47,7 @@
         </div>
       </div>
     </div>
-    <%= turbo_frame_tag 'actions_show' %>
+    <%= turbo_frame_tag Avo::ACTIONS_TURBO_FRAME_ID %>
```
Paul-Bob: Great! We should use this on our docs on the multiple action flux https://docs.avohq.io/3.0/actions.html#:~:text=you%20may%20want%20to%20redirect%20to%20another%20action.%20

lib/avo.rb
```diff
@@ -18,6 +18,7 @@ module Avo
   IN_DEVELOPMENT = ENV["AVO_IN_DEVELOPMENT"] == "1"
   PACKED = !IN_DEVELOPMENT
   COOKIES_KEY = "avo"
+  ACTIONS_TURBO_FRAME_ID = :actions_show
```
Paul-Bob, line 21: Great! We should use this on our docs on the multiple action flux https://docs.avohq.io/3.0/actions.html#:~:text=you%20may%20want%20to%20redirect%20to%20another%20action.%20

app/controllers/avo/actions_controller.rb
```diff
@@ -73,12 +73,15 @@ def respond
...
+            render turbo_stream: turbo_stream.turbo_frame_set_src(frame_id, src)
           when :redirect
-            # Turbo redirect to the path
             render turbo_stream: turbo_stream.redirect_to(
               Avo::ExecutionContext.new(target: @response[:path]).handle,
-              nil,
-              @response[:redirect_args][:turbo_frame],
+              turbo_frame: @response[:redirect_args][:turbo_frame],
               **@response[:redirect_args].except(:turbo_frame)
             )
```
Paul-Bob, line 86: Let's also update the DOCS about how to use `navigate_to_action` on the multi action flux 
https://docs.avohq.io/3.0/actions.html#:~:text=you%20may%20want%20to%20redirect%20to%20another%20action.%20

Will this still work for people who implemented it with the redirect method or should be referred as a breaking change?

adrianthedev: No. it will not work anymore.
That's a breaking change.
Do we know who originally asked for this?

Paul-Bob: We had some discussion arround this:

Adding and maintaining a new action response ( `navigate_to_action` in this case ) VS Using the `redirect_to` method ( no maintenance need, only docs on how to use it on multi action flux )

This is the original request:
https://discord.com/channels/740892036978442260/1162459368827146292

I think more projects are using it, I remember sending the docs links on some discord threads.

app/controllers/avo/actions_controller.rb
```diff
@@ -73,12 +73,15 @@ def respond
...
+
+            render turbo_stream: turbo_stream.turbo_frame_set_src(frame_id, src)
           when :redirect
-            # Turbo redirect to the path
             render turbo_stream: turbo_stream.redirect_to(
               Avo::ExecutionContext.new(target: @response[:path]).handle,
-              nil,
-              @response[:redirect_args][:turbo_frame],
+              turbo_frame: @response[:redirect_args][:turbo_frame],
               **@response[:redirect_args].except(:turbo_frame)
```
Paul-Bob, line 85: Not sure if this modification follows the `redirect_to` arguments syntax:
https://github.com/marcoroth/turbo_power-rails/blob/80f9d1a8b8aa43200557d3210d0b2cabe0700c47/lib/turbo_power/stream_helper.rb#L239

app/controllers/avo/actions_controller.rb
Paul-Bob: @adrianthedev just tested it locally with the city update action, it works with the `redirect_to` so it's not a breaking change, it's only another way of doing it

adrianthedev: It is if you run the action multiple times. It does something weird like emptying out the page.

Paul-Bob: And that don't happen on main? There is a [similar issue with downloads](https://github.com/avo-hq/avo/issues/2385)

Paul-Bob:
@adrianthedev this one is ready, in this commit https://github.com/avo-hq/avo/commit/2a002a1bc8b522e232af9a5dd307c53bc75a30f1#diff-6cacc45576472b27cf8c23100fe9b8fd939c44c51c67dd4988f993adf91d366bR8 the `resourceName` stimulus value was registered as Boolean by mistake making all actions break on index, that's why the multi flux action was not working, it works now with both APIs

adrianthedev:
Wow. Great spotting! My bad on this one!
Yes, I would like to keep that API. I think it's really cool and will help people get used to that flow.
WDYT?

Paul-Bob:
Right now both works just fine:

```ruby
 def handle(**args)
    arguments = Base64.encode64 Avo::Services::EncryptionService.encrypt(
      message: {
        cities: args[:query].map(&:id),
        render_name: args[:fields][:name],
        render_population: args[:fields][:population]
      },
      purpose: :action_arguments
    )

    path, data = Avo::Actions::City::Update.link_arguments(resource:, arguments: )
    redirect_to path, **data
  end
```

OR

```ruby
 def handle(**args)
    arguments = Base64.encode64 Avo::Services::EncryptionService.encrypt(
      message: {
        cities: args[:query].map(&:id),
        render_name: args[:fields][:name],
        render_population: args[:fields][:population]
      },
      purpose: :action_arguments
    )

    navigate_to_action Avo::Actions::City::Update, arguments:
  end
```

IMHO if we document well the `redirect_to` we don't need to add `navigate_to_action` . We also have a `redirect_to` approach when want to navigate to an resource instead `navigate_to_resource`. But I see that `navigate_to_action` can be cleaner and a little more user friendly.

I think we should approach this issue https://github.com/avo-hq/avo/issues/2185 and make possible this API:

```ruby
 def handle(**args)
    path, data = Avo::Actions::City::Update.link_arguments(
      resource:,
      arguments: {
        cities: args[:query].map(&:id),
        render_name: args[:fields][:name],
        render_population: args[:fields][:population]
      }
    )
    redirect_to path, **data
  end
```

Same for `navigate_to_action` it makes no sense to pass the encoding logic to the user since the user have no power over the decoding logic.

I would go only for the `redirect_to` with the above API. We can keep `navigate_to_action` for user simplicity sake but we should also do the encoding part, for the same reason, decoding is done by `ActionsController`:

```ruby
 def handle(**args)
    arguments =  {
        cities: args[:query].map(&:id),
        render_name: args[:fields][:name],
        render_population: args[:fields][:population]
      }

    navigate_to_action Avo::Actions::City::Update, arguments:
  end
```

Paul-Bob:
We decided to keep the `navigate_to_action`.

Approach to end this PR:

- [ ] Create an class method `encode_arguments` used to encode action arguments
- [ ] `navigate_to_action` and `link_arguments` should both receive arguments and a hash and encode them using the `encode_arguments` from the action class
- [ ] Check where we use the  `Base64.encode64 Avo::Services::EncryptionService.encrypt`... with the `purpose: :action_arguments` and replace it by also using the action class created method `encode_arguments`

@gabrielgiroe1 can you please take this to the finish line?

adrianthedev:
We also need to add some kind of API to disable [instantclick](https://turbo.hotwired.dev/handbook/drive#instantclick). @Paul-Bob can you please add it to the requirements?

Paul-Bob:
@adrianthedev wondering if it's suitable to add an turbo configuration hash where we can add options:

```ruby
# initializers/avo.rb
config.turbo = {
  instantclick: false,
  another_turbo_option: true,
  ...
}
``

**Comment by adrianthedev** [comment]
> @adrianthedev wondering if it's suitable to add an turbo configuration hash where we can add options:
> 
> ```ruby
> # initializers/avo.rb
> config.turbo = {
>   instantclick: false,
>   another_turbo_option: true,
>   ...
> }
> ```

Hmm, yeah. Let's try that approach @Paul-Bob.

File: app/controllers/avo/actions_controller.rb [hunk 5]
```diff
@@ -119,10 +124,7 @@ def decrypted_arguments
       arguments = params[:arguments] || params.dig(:fields, :arguments)
       return if arguments.blank?
```
Paul-Bob, line 125:
```suggestion
```

app/helpers/avo/turbo_stream_actions_helper.rb
```diff
@@ -11,7 +11,9 @@ def flash_alerts
     end
 
     def close_action_modal
-      turbo_stream_action_tag :replace, target: "actions_show", html: @view_context.turbo_frame_tag("actions_show")
+      turbo_stream_action_tag :replace,
+        target: Avo::ACTIONS_TURBO_FRAME_ID,
+        html: @view_context.turbo_frame_tag(Avo::ACTIONS_TURBO_FRAME_ID)
```
adrianthedev, line 16: Instead of doing a replace and going through the `view_context` and all, wouldn't it be easier to empty it?
`turbo_power` has `inner_html = ""`

adrianthedev:
Looks good to merge. Thanks you guys for finishing this.

app/helpers/avo/turbo_stream_actions_helper.rb
Paul-Bob: https://github.com/avo-hq/avo/pull/2521#issuecomment-1966017434

PR #2489, opened 2024-02-16
Updated the default client authorization to nil

### LINKED ISSUE #2475: Add `noop` authorization driver
### Feature

If a user want to use an authorization client different from `pundit`, they will still need to install it so it won't raise an error.

CleanShot 2024-02-13 at 21 56 39@2x

gabrielgiroe1:
# Description
Updated the default client authorization to nil in avo.tt
In the Avo initializer file, modified the setting for `config.authorization_client` to be nil instead of `:pundit`.

Fixes #2475

# Checklist:

## Manual review steps

1. Step 1
1. Step 2


adrianthedev (approved):
GG!

Issue #2443, opened 2024-02-04
Association has_many not ordering correctly


rctneil:
I have a Visit model and a Cycle model.  A Visit has_many Cycles and a Cycle belongs_to a Visit.

I have these showing up as a has_many field type on my Visit Show view. I wish to add an order to them of position: :asc, but it just refuses to do so.

I see in my log:

`SELECT "cycles".* FROM "cycles" WHERE "cycles"."visit_id" = $1 ORDER BY cycles.created_at desc LIMIT $2 OFFSET $3`

My Cycle resource has:

```
  self.index_query = -> {
    if params[:via_association] == 'has_many'
        query.order(position: :asc)
    else
        query.order(date: :desc, position: :asc)
    end
  }
```

This works flawlessly on the Index view, but when being shown via an association, the order just doesn't take affect and seems to order by created_at as per the query shown.

Paul-Bob:
Having a quick look into this, the `index_query` never gets executed when the table index is rendered from an association.

https://github.com/avo-hq/avo/blob/edb3a18aac8a15fd67be51e4758db12f0335df63/app/controllers/avo/base_controller.rb#L29

We should either apply it on the above line or on the `Avo::AssociationsController` when generating the query.

rctneil:
@Paul-Bob Thanks for taking a look!

Glad I wasn't doing something wrong for a change!   Do you have an ETA on a fix for this one?

Paul-Bob:
@rctneil forgot to mention that there is an workaround:

https://docs.avohq.io/3.0/associations/has_many.html#scope

You can scope the association field

Paul-Bob:
Just opened an issue with a feature request on this one.

https://github.com/avo-hq/avo/issues/2445

We're not prioritizing it right now since there's a solid workaround. Any PR regarding this issue is welcomed!

rctneil:
@Paul-Bob thanks for the workaround! Works perfectly!

PR #2492, opened 2024-02-17
feature: Always use index_query when loading a resource

sriddbs:
# Description

Fixes # (issue)

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/app/controllers/avo/associations_controller.rb
+++ b/app/controllers/avo/associations_controller.rb
@@ -24,6 +24,7 @@ module Avo
       association_name = BaseResource.valid_association_name(@parent_record, params[:related_name])
       @query = @related_authorization.apply_policy @parent_record.send(association_name)
       @association_field = @parent_resource.get_field params[:related_name]
+      @query = Avo::ExecutionContext.new(target: @resource.class.index_query, query: @query, parent: @parent_record).handle
 
       if @association_field.present? && @association_field.scope.present?
         @query = Avo::ExecutionContext.new(target: @association_field.scope, query: @query, parent: @parent_record).handle


sriddbs:
hey @Paul-Bob let me know if this is all what is required 😄 
can you also point me where to add the specs for this please?

Paul-Bob (changes requested):
Hello @sriddbs thank you for the contribution!

There are a few points that we need to take in consideration.

1 - How to distinguish on the `index_query` block where we're on an association (with the current implementation i guess we can do `@parent.present?`)
2 - Add an example about how to distinguish when run on resource's index or association
3 - Run tests on that example
4 - Documentation (https://github.com/avo-hq/avodocs)

The example can be added anywhere that you think it fits on `spec/dummy` app.

You can add a new test file that tests both query application, on resource's index and as an association.

We have our specs on `spec/system/avo/...`  / `spec/feature/avo/...` ... We're trying to add more feature tests since the system ones are already taking much time. But if system test is necessary that's ok!

Thanks again for looking into this!

sriddbs:
Thank you @Paul-Bob for the detailed explanation 👍 

> How to distinguish on the index_query block where we're on an association (with the current implementation i guess we can do @parent.present?)

am making the changes in `app/controllers/avo/associations_controller.rb`, so do we still need the `@parent.present?` check?

> The example can be added anywhere that you think it fits on spec/dummy app.

yes, I added the `index_query` on a resource which is also loaded via association and the changes I made seem working
but as you suggested I will add the specs as well 👍

adrianthedev:
I'm going to let you guys jam on this. I have a couple of questions.

1. Is this going to be a breaking change for some existing users?
2. Do we really want to apply the `index_query` to associations too? Why?

Paul-Bob:
> am making the changes in `app/controllers/avo/associations_controller.rb`, so do we still need the `@parent.present?` check?

@sriddbs I mean on the Proc, we should be able to separate the query when it's executed as association:

```ruby
self.index_query = -> {
  if on_association? # fictive method I think we can do @parent.present?
    query.do_this...
  else
    query.do_that...
  end
}
```

Makes sense? There are cases when we want to apply different logic to the query when is loaded as association.

sriddbs:
> I mean on the Proc, we should be able to separate the query when it's executed as association

Ahh got it 👍

adrianthedev:
@sriddbs what is the use-case on your end? What would you like to apply from `index_query` to the association?

spec/features/index_query_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+      find('a[data-selected="false"][data-tabs-tab-name-param="People"]', match: :first).click
+      people = all('[data-field-id="name"]')
+
+      first_person = people[0].find("a").text
+      second_person = people[1].find("a").text
+
+      expect(first_person).to eq person_1.name
+      expect(second_person).to eq person_2.name
+    end
+  end
```
sriddbs, line 39: @Paul-Bob this spec fails because `@resouce.record` is `nil`, any idea how to fix?
I think the `record` is not set for `related_resource` [here](https://github.com/avo-hq/avo/pull/2492/files#diff-b1f64e5509fa895bb0217e2740a262487a1344069c30bef0391e92f05c760b91R21) 🤔

adrianthedev:
Closing for now as we haven't gotten the use-case for it. Please feel free to reopen.

Paul-Bob:
Linked discussions: 

- https://github.com/avo-hq/avo/issues/2443
- https://discord.com/channels/740892036978442260/1217489935007879239/1217512469170491412

Workaround: use [scope](https://docs.avohq.io/3.0/associations/has_many.html#scope), if it repeats all over the code create a helper for it

PR #2520, opened 2024-02-26
fix: remove duplicate create method

### LINKED ISSUE #1944: Weird behavior when creating a model from a has_many field
### Context

I have a model which `belongs_to` another model
```ruby
create_table :notes do |t|
  t.text :note, null: false
  t.belongs_to :provider, null: false, foreign_key: true
end
```
```ruby
class Note < ApplicationRecord
  belongs_to :provider

  validates :note, presence: true
end
```

then on the parent resource i have a `has_many` field
```ruby
field(:notes, as: :has_many)
```

When i try to create a new note the UI says the creation failed but the note is created successfully

I am not sure if i missed something and this is expected but i took a quick look at the source code and believe that the issue is [Here](https://github.com/avo-hq/avo/blob/8f780eca21e0e80595925aece6f76f703ec78a22/app/controllers/avo/base_controller.rb#L129C9-L129C9), the controller tries to save the model before setting `provider_id` which fails (This would be the error reported on the UI) but then [later on](https://github.com/avo-hq/avo/blob/8f780eca21e0e80595925aece6f76f703ec78a22/app/controllers/avo/base_controller.rb#L144) the model is saved again after setting `provider_id` and succeds.

as a work around i am overriding the `save_model_action` method on the `NotesController`
```ruby
class Avo::NotesController < Avo::ResourcesController
  def save_model_action
    true
  end
end
```

with that creating notes works as expected

### System configuration
**Avo version**: 2.36.2

**Rails version**: 7.0.7.2

**Ruby version**: 3.2.0

**License type**

 - [x] Community
 - [x] Pro

 - [x] No

https://github.com/avo-hq/avo/assets/6798292/4e760788-373e-41cb-ae6c-c3dd0c89e4d0

Paul-Bob:
# Description

Fixes #1944

Save record only once on creation after applying the associations logic.

# Checklist:

### DIFF
--- a/app/controllers/avo/base_controller.rb
+++ b/app/controllers/avo/base_controller.rb
@@ -123,10 +123,6 @@ module Avo
     end
 
     def create
-      # record gets instantiated and filled in the fill_record method
-      saved = save_record
-      @resource.hydrate(record: @record, view: :new, user: _current_user)
-
       # This means that the record has been created through another parent record and we need to attach it somehow.
       if params[:via_record_id].present? && params[:via_belongs_to_resource_class].nil?
         @reflection = @record._reflections[params[:via_relation]]
@@ -139,7 +135,6 @@ module Avo
           related_record = related_resource.find_record params[:via_record_id], params: params
 
           @record.send("#{@reflection.foreign_key}=", related_record.id)
-          @record.save
         end
 
         # For when working with has_one, has_one_through, has_many_through, has_and_belongs_to_many, polymorphic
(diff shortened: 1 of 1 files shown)


app/controllers/avo/base_controller.rb
```diff
@@ -139,7 +135,6 @@ def create
           related_record = related_resource.find_record params[:via_record_id], params: params
 
           @record.send("#{@reflection.foreign_key}=", related_record.id)
```
adrianthedev, line 137: Hmmm. Wouldn't this break if this is a new record and it's not persisted to the DB?

```ruby
user = User.new
user.post = Post.new # wouldn't this break if there were a FK constraint or validation rule?
user.save
```

Paul-Bob: Actually this is being done:

```ruby
user = User.new
user.post_id = 4 # random id
user.save
```

`related_record` is already persistent

adrianthedev: Got it! Then it's good to merge!

adrianthedev (approved):

Issue #2515, opened 2024-02-24
Confusing error message


dlupu:
### Describe the bug
Hi guys,

I'm trying to setup AvoAdvanced 3.4.1 in a client project and just run into a small bug. 

For a `User` model the `has_one :user_preference` Avo  complains that there is no Avo resource for user_preference, which is indeed the case. However the error message is wrong:

> Failed to find a resource while rendering the  :userpreference field. 
> You may generate a resource for it by running 'rails generate avo:resource userpreference'.

(notice that the message uses `userpreference` instead of  `user_preference`)

Running that command does not fix the problem, as you can imagine

> ➜  cashlee-api add-avo ✗ fxbx rails generate avo:resource userpreference
> Avo advanced 3.4.1
> Can't generate fields from model. 'userpreference.rb' not found!
>       create  app/avo/resources/userpreference.rb
>       create  app/controllers/avo/userpreferences_controller.rb

### Steps to Reproduce

=> see explanation above

### System configuration

**Avo version**: 3.4.1
**Rails version**: 6.1.6
**Ruby version**: 3.2.2
**License type**:

 - [ ] Community
 - [ ] Pro
 - [x] Advanced

**Are you using Avo monkey patches, overriding views or view components?**

 - [ ] Yes. If so, please post code samples.
 - [x] No

#### Impact

 - [ ] High impact (It makes my app un-usable.)
 - [ ] Medium impact (I'm annoyed, but I'll live.)
 - [x] Low impact (It's really a tiny thing that I could live with.)

#### Urgency

 - [ ] High urgency (I can't continue development without it.)
 - [ ] Medium urgency (I found a workaround, but I'd love to have it fixed.)
 - [x] Low urgency (It can wait. I just wanted you to know about it.)

PR #2531, opened 2024-02-28
fix: Update error message in MissingResourceError class

gabrielgiroe1:
# Description
The error message now correctly converts camel case resource names to snake case before displaying the message

Fixes [2515](https://github.com/avo-hq/avo/issues/2515)

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/lib/avo.rb
+++ b/lib/avo.rb
@@ -39,7 +39,7 @@ module Avo
     private
 
     def missing_resource_message(resource_name)
-      name = resource_name.to_s.gsub(/([a-z])([A-Z])/, '\1_\2').downcase
+      name = resource_name.to_s.underscore.downcase
 
       "Failed to find a resource while rendering the :#{name} field.\n" \
       "You may generate a resource for it by running 'rails generate avo:resource #{name.singularize}'.\n" \
--- a/lib/avo.rb
+++ b/lib/avo.rb
@@ -39,7 +39,7 @@ module Avo
     private
 
     def missing_resource_message(resource_name)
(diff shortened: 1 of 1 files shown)


lib/avo.rb
```diff
@@ -39,7 +39,7 @@ def initialize(resource_name)
     private
 
     def missing_resource_message(resource_name)
-      name = resource_name.to_s.downcase
+      name = resource_name.to_s.gsub(/([a-z])([A-Z])/, '\1_\2').downcase
```
Paul-Bob, line 42: I think regex is overkill for this use case, WDYT about using the [`underscore`](https://apidock.com/rails/String/underscore) method?

```suggestion
      name = resource_name.to_s.underscore
```

adrianthedev (approved):

lib/avo.rb
```diff
@@ -39,7 +39,7 @@ def initialize(resource_name)
     private
 
     def missing_resource_message(resource_name)
-      name = resource_name.to_s.downcase
+      name = resource_name.to_s.underscore.downcase
```
Paul-Bob, line 42:
```suggestion
      name = resource_name.to_s.underscore
```

We can remove the [`downcase`](https://apidock.com/ruby/String/downcase) since [`underscore`](https://apidock.com/rails/String/underscore) already lowercase the string

gabrielgiroe1: Ok

Paul-Bob (approved):

PR #2546, opened 2024-03-01
feature: add autofocus on first field

gabrielgiroe1:
# Description

# Checklist:

## Manual review steps

1. Step 1
1. Step 2


app/views/avo/actions/show.html.erb
```diff
@@ -32,7 +32,7 @@
             <%= render field
               .hydrate(resource: @resource, record: @resource.record, user: @resource.user, view: @view, action: @action)
               .component_for_view(@view)
-              .new(field: field, resource: @resource, index: index, form: form, compact: true)
+              .new(field: field, resource: @resource, index: index, form: form, compact: true, autofocus: index.zero?)
```
Paul-Bob, line 35: Where is this `autofocus` option being applied? Lets ensure that it applies on all edit components

app/views/avo/actions/show.html.erb
```diff
@@ -51,7 +51,8 @@
             size: :sm,
             data: {
               target: :submit_action,
-            } do %>
+              action_target: "submit"
```
Paul-Bob, line 54: Do we really need this target?

```suggestion
```

app/views/avo/actions/show.html.erb
```diff
@@ -51,7 +51,8 @@
...
+            }, autofocus: (@action.get_fields.count == 2 ? true : false)  do %>
```
Paul-Bob, line 55: `get_fields` will fetch the fields again, that's not necessary, actually we are already fetching them twice. Lets fetch them on `Avo::ActionsController` like: `@fields = @action.get_fields` and replace all `@action.get_fields` with `@fields` on the view.

Paul-Bob (changes requested):
Hey @gabrielgiroe1 we have a good start, I let some comments about the following steps

app/views/avo/actions/show.html.erb
gabrielgiroe1: I think this came from the car when I resolved the conflict

app/views/avo/actions/show.html.erb
gabrielgiroe1: Good idea. I will make the changes.

app/components/avo/fields/text_field/edit_component.html.erb
```diff
@@ -7,6 +7,7 @@
     placeholder: @field.placeholder,
     style: @field.get_html(:style, view: view, element: :input),
     multiple: multiple,
-    autocomplete: @field.autocomplete
+    autocomplete: @field.autocomplete,
+    autofocus: index.zero?
```
Paul-Bob, line 11: `autofocus` should be passed when initializing the field and not computed here.

For this feature we want to add focus when first field on an action.

```suggestion
    autofocus: @field.autofocus?
```

And lets add `autofocus` as an option for all components of fields types that can be focused.

app/controllers/avo/actions_controller.rb
```diff
@@ -48,6 +48,7 @@ def set_action
         view: :new, # force the action view to in order to render new-related fields (hidden field)
         arguments: BaseAction.decode_arguments(params[:arguments] || params.dig(:fields, :arguments)) || {}
       )
+      @fields = @action.get_fields
```
Paul-Bob, line 51: This run every time that `set_action` is called. `set_action`  is called on `:show` and `:handle` (`before_action :set_action, only: [:show, :handle]`).

Since we only want to use `@fields` on `show` lets move this line to `show` method to avoid the unnecessary call on `handle`

app/views/avo/actions/show.html.erb
```diff
@@ -26,9 +26,9 @@
       <%= form.hidden_field :avo_resource_ids, value: params[:id] || params[:resource_ids], 'data-action-target': 'resourceIds' %>
       <%= form.hidden_field :avo_selected_query, 'data-action-target': 'selectedAllQuery' %>
       <%= form.hidden_field :arguments, value: params[:arguments] %>
-      <% if @action.get_fields.present? %>
+      <% if @fields.present? %>
         <div class="mt-4 -mx-6">
-          <% @action.get_fields.each_with_index do |field, index| %>
+          <% @fields.each_with_index do |field, index| %>
             <%= render field
```
Paul-Bob, line 32: Here when initializing the component we should decide if autofocus is on or off, something like:

`.component_for_view(@view).new(..., autofocus: index == 0)`

app/views/avo/actions/show.html.erb
```diff
@@ -51,7 +51,7 @@
             size: :sm,
             data: {
               target: :submit_action,
-            } do %>
+            }, autofocus: (@fields.count == 2 ? true : false)  do %>
```
Paul-Bob, line 54:
```suggestion
            }, autofocus: @fields.empty?  do %>
```

Lets autofocus the run button when fields are empty instead when there are exactly 2 fields

Paul-Bob (changes requested):
@gabrielgiroe1 I let some comments about the next steps, where you think the suggestion totally fits as it is you can hit the `Commit suggestion` button

app/views/avo/actions/show.html.erb
gabrielgiroe1: I tried something like this, but it only works for the text field.

app/views/avo/actions/show.html.erb
gabrielgiroe1: ok, I will update it.

app/controllers/avo/actions_controller.rb
gabrielgiroe1: Got it.

gabrielgiroe1:
Closed because it is related to https://github.com/avo-hq/avo/pull/2554

PR #2549, opened 2024-03-01
fix: index header and rows visibility

### LINKED ISSUE #2555: Broken index when listing records with different field visibilities
### Describe the bug
Broken table layout on index when listing records with visibilities which differ between rows.

### Steps to Reproduce
Using this fields definition:

```rb
  def fields
    field :title, as: :text, sortable: true, link_to_record: true
    field :slug, as: :text, sortable: true, hide_on: :new
    field :type, as: :text, sortable: true
    field :designation, as: :select, enum: ::Content.designations, sortable: true, visible: ->{ resource.record&.type.nil? }
    field :free, as: :boolean, sortable: true
    field :public, as: :boolean, sortable: true
   # ...
  end
```

I followed this [documentation](https://docs.avohq.io/3.0/field-options.html#field-visibility) which does not mention that `visibility` couldn't be used in this way.

See screenshot. 
Expected: In the first row, designation should be "—", free should be "✅" and public should be "✅"
Actual: In the first row, designation has "✅" (value of the "free" column)

### System configuration
**Avo version**: 3.4.3

**Rails version**: 7.0.8.1

**Ruby version**: 3.3.0

**License type**:
 - [x] Community
 - [ ] Pro
 - [ ] Advanced

 - [x] No

Screenshot 2024-03-04 at 16 03 33

#### Impact

#### Urgency

Paul-Bob:
# Description

Fixes #2555

### Problems: 

1. Header is fetching the resource fields while hydrated with last table record. Let's say that last record don't have 2 columns visible, header will miss that columns and some rows will have values for it.

2. If with luck last record have all columns visible we still having 1 issue, some rows can have 1 or 2 or more columns not visible, right now that columns are not even taking in consideration, them are just skipped and next column value is rendered on the wrong spot.
image

### Solution:

1. Prepare components for each row.
2. Extract all distinct columns and render them as header.
4. Render `-` when a row don't have field for a specific column.

# Checklist:

### Before
image

### After
image


adrianthedev:
Please let's wait a little bit with the `@header_fields` implementation. I wanna have a closer look on this.

app/controllers/avo/base_controller.rb
```diff
@@ -68,11 +68,6 @@ def index
         @resource.dup.hydrate(record: record)
       end
 
-      # Temporary fix for visible blocks when geting fields for header
-      # Hydrating with last record so resource.record != nil
-      # This is keeping same behavior from <= 3.4.1
-      @resource.hydrate(record: @records.last)
-
       set_component_for __method__
```
Paul-Bob, line 71: Fixes https://github.com/avo-hq/avo/issues/2653#issuecomment-2044518084

Paul-Bob:
> Please let's wait a little bit with the `@header_fields` implementation. I wanna have a closer look on this.

This PR is ready to review. The main point of `@header_fields` implementation is to "prepare" each table row before rendering the header. 

If we know what columns are present on each row we can then compute the header by making an union.

This solves many issues. 

1. Resource will no longer need to be wrongly hydrated with the last table record:
  1.1. This lead to several problems like an action link generated for that resource will wrongly target the last table record. Also all blocks executed on the index have the last table record hydrated, for example the `self.index_controls` https://github.com/avo-hq/avo/issues/2653#issuecomment-2044518084
2. Header will no longer miss columns, even if only 1 row have a value for a column, that column will be rendered
3. Row values will respect the column, I.E. if a row have a certain field hidden but the column exists now an empty dash will be rendered instead printing the next field value on the wrong column

app/controllers/avo/base_controller.rb
adrianthedev: I understand this, but didn't we used to have a problem with checking the the presence of `record` in blocks, or this is a different fix?

spec/dummy/app/avo/resources/project.rb
```diff
@@ -22,7 +22,11 @@ def fields
       filterable: true,
       summarizable: true
     field :name, as: :text, required: true, sortable: true, default: "New project default name"
-    field :progress, as: :progress_bar, value_suffix: "%", display_value: true
+    field :progress,
+      as: :progress_bar,
+      value_suffix: "%",
+      display_value: true,
+      visible: -> { resource.view.form? || resource.record.progress&.positive? }
```
adrianthedev, line 29:
```suggestion
      visible: -> do 
        # conditionally hiding the fields we can text that it's not going to break the table layout
        resource.view.form? || resource.record.progress&.positive?
      end
```

app/components/avo/index/table_row_component.html.erb
```diff
@@ -21,8 +21,13 @@
...
     </td>
   <% end %>
-  <% @resource.get_fields(reflection: @reflection, only_root: true).each_with_index do |field, index| %>
-    <%= render field.component_for_view(:index).new(field: field, resource: @resource, reflection: @reflection, index: index, parent_record: @parent_record, parent_resource: @parent_resource) %>
+  <% @header_fields.each_with_index do |header_field, index| %>
+    <% field = @fields.find { |field| field.id == header_field } %>
+    <% if field.present? %>
+      <%= render field.component_for_view(:index).new(field: field, resource: @resource, reflection: @reflection, index: index, parent_record: @parent_record, parent_resource: @parent_resource) %>
+    <% else %>
+      <td class="text-center">—</td>
```
adrianthedev, line 29:
```suggestion
      <td class="text-center">—</td>
```

app/components/avo/index/resource_table_component.rb
```diff
@@ -13,6 +13,30 @@ def initialize(resources: nil, resource: nil, reflection: nil, parent_record: ni
...
+          reflection: reflection,
+          parent_record: parent_record,
+          parent_resource: parent_resource,
+          actions: actions
+        }
+      end
+
+      # Render only uniq fields on header
+      @header_fields.uniq! { |field| field.table_header_label }
+    end
```
adrianthedev, line 39: I feel like this is a bit too much.
Can we split it or refactor it to be more easily understood at first glance?

It took a while and some concentrating power to make sure I understand what it does.

app/controllers/avo/base_controller.rb
Paul-Bob: > didn't we used to have a problem with checking the the presence of `record` in blocks

## TL;DR; 

With new computed header logic this is no longer an issue and this bug can be safely removed fixing many edge cases.

## Long version

When we discovered this bug we removed hydration with the last record on the resource which resulted in `resource.record` and `record` being `nil` as expected on the Index when fetching fields for the header. This rollback of hydration was done solely to prevent breaking the provided example.

For example:

```ruby
field :name, as: :text, visible: -> { record.something? } 
```

When checking fields for header this visible block was getting executed and when we removed the last record hydration it was breaking since record was `nil`. 

With this PR we can safely remove this wrong hydration since header now is calculated having in consideration each row fields. This hydration was causing so many other issues, and many of them not "breakable" but weird edge cases.

Through the approach in this PR, we now execute `visible: -> { record.something? }` for each row, ensuring that record is hydrated, and then compute the header based on the visible fields of each row.

spec/dummy/app/avo/resources/project.rb
Paul-Bob: @adrianthedev did you mean "we can test" instead "we can text"?

app/components/avo/index/resource_table_component.rb
Paul-Bob: Totally understand, let«s refactor it

Paul-Bob: Can you please look at the last version? I tried to comment each line and split the logic

spec/dummy/app/avo/resources/project.rb
adrianthedev: yes 🤦‍♂️

adrianthedev (approved):

spec/dummy/app/avo/resources/project.rb
```diff
@@ -22,7 +22,14 @@ def fields
       filterable: true,
       summarizable: true
     field :name, as: :text, required: true, sortable: true, default: "New project default name"
-    field :progress, as: :progress_bar, value_suffix: "%", display_value: true
+    field :progress,
+      as: :progress_bar,
+      value_suffix: "%",
+      display_value: true,
+      visible: -> do
```
Paul-Bob, line 29:
```suggestion
      visible: -> do
```

xeron:
I feel like this PR might be the reason for the issue I'm seeing: https://github.com/avo-hq/avo/issues/2732

Issue #2187, opened 2023-12-18
Focus the run button on an action without fields


adrianthedev:
When an action that doesn't have any fields is opened up, we should focus the Run button so the user can quickly hit return to run it or tab to cancel.

- [ ] Soemwhere around [here](https://github.com/avo-hq/avo/blob/chore/tags-styles-improvements/app/views/avo/actions/show.html.erb#L46)

Issue #2186, opened 2023-12-18
Focust the first input when an action opens


adrianthedev:
Depends on https://github.com/avo-hq/avo/issues/2368

When an action with some fields is opened up, the first field is not focused. 
We should autofocus it so they do less clicking or tabbing around.

- [ ] Probably set the `autofocus` to true somewhere around [here](https://github.com/avo-hq/avo/blob/chore/tags-styles-improvements/app/views/avo/actions/show.html.erb#L33).

Issue #2142, opened 2023-12-18
Add field_input_attributes


adrianthedev:
_First see if it's pheasable_

Let's try to control the field attributes from the field somehow. This would be useful for plugins.

```erb
# instead of
<%= field_wrapper **field_wrapper_args do %>
  <%= form.text_field @field.id,
    class: classes("w-full"),
    data: @field.get_html(:data, view: view, element: :input),
    disabled: @field.is_readonly? || @field.is_disabled?,
    placeholder: @field.placeholder,
    style: @field.get_html(:style, view: view, element: :input),
    # value: @field.value,
    multiple: multiple
  %>
<% end %>


# we have
<%= field_wrapper **field_wrapper_args do %>
  <%= form.text_field @field.id, **field_input_args %>
<% end %>
```

Now, the `field_input_args` can be intercepted and change the args.

### Use case

Before, we had only `@field.readonly` to mark the field as disabled. Now we're introducing the `@field.disabled` option. and we have to update all field components to support the new `@field.disabled` option.

PR #2554, opened 2024-03-04
Autofocus first field

gabrielgiroe1:
# Description

Fixes [#2186](https://github.com/avo-hq/avo/issues/2186)
Fixes [#2187](https://github.com/avo-hq/avo/issues/2187)

# Checklist:

## Manual review steps

1. Step 1
1. Step 2


app/controllers/avo/actions_controller.rb
```diff
@@ -17,6 +17,7 @@ def show
       @view = Avo::ViewInquirer.new("new")
 
       @resource.hydrate(record: @record, view: @view, user: _current_user, params: params)
+      @fields = @action.get_fields.reject { |field| field.is_a?(Avo::Fields::HiddenField) }
```
Paul-Bob, line 20: We want to render all fields on show, even the hidden ones.

```suggestion
      @fields = @action.get_fields
```

app/views/avo/actions/show.html.erb
```diff
@@ -51,7 +51,7 @@
             size: :sm,
             data: {
               target: :submit_action,
-            } do %>
+            }, autofocus: @fields.empty? do %>
```
Paul-Bob, line 54: Here is where we want to ignore the hidden ones

```suggestion
            }, autofocus: @fields.reject { |field| field.is_a?(Avo::Fields::HiddenField) }.empty? do %>
```

Paul-Bob (changes requested):
Lets render all fields and check the hidden ones only for button autofocus logic

app/components/avo/fields/edit_component.rb
```diff
@@ -12,8 +12,9 @@ class Avo::Fields::EditComponent < ViewComponent::Base
   attr_reader :resource
   attr_reader :stacked
   attr_reader :view
+  attr_reader :autofocus
```
Paul-Bob, line 15: `attr_reader` is slower than directly using the `@autofocus` instance variable and since we don't want to access it outside the component we should use directly the instance variable.

```suggestion
```

app/components/avo/fields/text_field/edit_component.html.erb
```diff
@@ -7,6 +7,7 @@
     placeholder: @field.placeholder,
     style: @field.get_html(:style, view: view, element: :input),
     multiple: multiple,
-    autocomplete: @field.autocomplete
+    autocomplete: @field.autocomplete,
+    autofocus: index.zero?
```
Paul-Bob, line 11:
```suggestion
    autofocus: @autofocus
```

Paul-Bob (changes requested):

app/components/avo/fields/text_field/edit_component.html.erb
gabrielgiroe1: I forgot to push this changes.

app/components/avo/fields/edit_component.rb
gabrielgiroe1: Yes, my mistake.

app/controllers/avo/actions_controller.rb
gabrielgiroe1: Yes. I will change it.

app/views/avo/actions/show.html.erb
gabrielgiroe1: Good idea.

adrianthedev:
Closing for now. We'd like to implement it using this solution
https://github.com/avo-hq/avo/issues/2142

adrianthedev (approved):
Looks good! Haven't manually tested it.

gabrielgiroe1:
> Looks good! Haven't manually tested it.

I tested, everything seems to be working fine. No issues found.

Paul-Bob (approved):

PR #2560, opened 2024-03-05
feature: `for_attribute` option for fields

### LINKED ISSUE #2559: Implement `association` option for association fields.
### Feature

Implement `association` option that allow specifying the association target for an association filed.

```ruby
with_options as: :has_many do
  field :users, scope: -> { query.non_admins }
  field :admins, association: :users, scope: -> { query.admins }
end
```

### Additional context

https://github.com/avo-hq/avo/issues/2513

Paul-Bob:
# Description

Fixes #2559

The `for_atribute` option allows to specify the association used for a certain field. This option make possible to define same association with different scopes and different name several times on the same resource.

```ruby
field :reviews,
  as: :has_many

field :special_reviews,
  as: :has_many,
  for_atribute: :reviews,
  scope: -> { query.special_reviews }
```

# Checklist:

  - https://github.com/avo-hq/avodocs/pull/180


Paul-Bob:
TODO:

- [ ] docs
- [ ] tests
- [ ] fix select all issue where it selects all from both tables

adrianthedev:
Let's wait on this.
I have an idea of a `for_column` or `for_field` (open to better naming) which should do the same thins as `association` but for any field.

```ruby
field :full_name, as: :text, for_field: :name, format_using: -> {something}
field :super_name, as: :text, for_field: :name, format_using: -> {something_else}
```

Do you think we need something like that?

Paul-Bob:
> Let's wait on this. I have an idea of a `for_column` or `for_field` (open to better naming) which should do the same thins as `association` but for any field.
> 
> ```ruby
> field :full_name, as: :text, for_field: :name, format_using: -> {something}
> field :super_name, as: :text, for_field: :name, format_using: -> {something_else}
> ```
> 
> Do you think we need something like that?

`for_field` option aims to the model attribute? I.E. both fields on your example will fetch the value from `record.name`?

If i understand it right I think that's interesting and is like renaming and extending the `association` option to all fields :ok_hand: 

This opens another suggestion, a way to directly pass the value similar like the field block but applied to all views

```ruby
field :full_name, as: :text, value: -> { record.name }, format_using: -> {something}
field :super_name, as: :text, value: -> { record.name }, format_using: -> {something_else}
```

Lets keep an eye on this PR since some logic might be reused whatever approach we follow.

Paul-Bob:
Actually with the `value: -> { record.name }` suggestion there is no need to format, is like `for_field` + `format_using` in one option

adrianthedev:
Let's create a PoC for this and name it `for_attribute` or `as_attribute`.

adrianthedev (changes requested):
Letf a comment.

PR #2573, opened 2024-03-08
Use human model name when creating new polymorphic association

iainbeeston:
# Description

Use the human model name in the link for creating a new polymorphic association, instead of the class name

Fixes # (issue)

Right now, when you have a polymorphic belongs to field and you specify the types argument (so the field shows a useful select menu allowing you to choose from the different polymorphic types) it has a link below the select menu saying "Create new <classname>". However, the name it uses for the resource is the class name used in the code, which isn't very user-friendly.

# Checklist:

## Screenshots & recording

This is what it looks like before (note the class name with modules in the link):

Screenshot 2024-03-08 at 12 36 13

This is what it looks like after:

Screenshot 2024-03-08 at 12 35 36

## Manual review steps

1. Find a polymorphic belongs to field that uses types, eg. https://main.avodemo.com/avo/resources/comments/549/edit
1. Switch to this branch
1. Confirm that the "create new <classname>" hasn't changed
2. Change the human name for the polymorphic class in the rails translations file
3. Reload avo and confirm that the name has updated to use the new translation

### DIFF
--- a/app/components/avo/fields/belongs_to_field/edit_component.html.erb
+++ b/app/components/avo/fields/belongs_to_field/edit_component.html.erb
@@ -70,7 +70,7 @@
           <% if field.can_create? %>
             <% create_href = create_path(Avo.resource_manager.get_resource_by_model_class(type.to_s)) %>
             <% if !disabled && create_href.present? %>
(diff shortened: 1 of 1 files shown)


Paul-Bob (approved):
Thanks for this fix @iainbeeston !

Issue #2283, opened 2023-12-21
Search bar still visible even if not authorized


Paul-Bob:
### Describe the bug

If `search?` policy return false or isn't defined the search bar still visible returning empty results

This can be confusing to debug, if an user is not authorized to search maybe the search bar shouldn't be visible at all.

Mentioned:
 - https://discord.com/channels/740892036978442260/1187426840709693470
 - https://github.com/avo-hq/avo/issues/1750
 - https://github.com/avo-hq/avo/issues/1778

adrienpoly:
This bites me a few time in times in the past too 

Maybe a way to solve it would be to be able to set a default policy for each resource. This is what I am doing but I need to manually add this policy to all my new resources and sometimes I forget. 

If either the generator could add in comment 

```ruby
# self.authorization_policy = Admin::BasePolicy
```

or 

in avo.rb we could define a base policy I think it would help

adrianthedev:
I'm curious on how you have your policies in the project. I noticed the `Admin` namespace.

I'm saying that because policies are automatically added to resources if they are not namespace.

Example:
`UserPolicy` -> `Avo::Resources::User`

adrianthedev:
> in avo.rb we could define a base policy I think it would help

I always thought that users will default to the `ApplicationPolicy` that pundit sets up for you.
But I see what you mean...

1. You generate a resource
2. gotta make sure the policy is applied
3. then add rules

We'll evaluate this approach. I'm just a bit afraid of all the `if/else`s we should add and the mental model that developers have right now.

adrienpoly:
We found that using the default resource policy does not work for us. As those policies are tied to the application they are lots of things that can be prevented for a user. A solution would be to always authorize if teh user is an admin but then it becomes very complicated to maintain.
Therefore we have created an `admin` namespace with an `Admin::ApplicationPolicy` and then resources policies inheriting from this base application policy.

so very often in a resource I will either add 
```ruby
self.authorization_policy = Admin::ApplicationPolicy
```

or create a new policy file and add it to my resource

PR #2585, opened 2024-03-11
Hide search input when avo_search? policy is not defined

gabrielgiroe1:
# Description
Modified the authorized_to_search? method in the ResourceIndexComponent to hide the search input if the avo_search? policy method is not defined for the resource.

Fixes [#2283](https://github.com/avo-hq/avo/issues/2283)

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/app/components/avo/views/resource_index_component.rb
+++ b/app/components/avo/views/resource_index_component.rb
@@ -135,10 +135,8 @@ class Avo::Views::ResourceIndexComponent < Avo::ResourceComponent
 
   def authorized_to_search?
     # Hide the search if the authorization prevents it
-    # return false unless resource.authorization.respond_to?(:avo_search?)
-    return false unless resource.authorization.respond_to?(:has_action_method?)
+    return true unless resource.authorization.respond_to?(:has_action_method?)
     return false unless resource.authorization.has_action_method?("search")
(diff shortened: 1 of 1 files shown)


app/components/avo/views/resource_index_component.rb
```diff
@@ -135,8 +135,10 @@ def show_search_input
 
   def authorized_to_search?
     # Hide the search if the authorization prevents it
-    return true unless resource.authorization.respond_to?(:has_action_method?)
-    return true unless resource.authorization.has_action_method?("search")
+    # return false unless resource.authorization.respond_to?(:avo_search?)
+    return false unless resource.authorization.respond_to?(:has_action_method?)
+    return false unless resource.authorization.has_action_method?("search")
+    return false unless resource.authorization.has_action_method?("avo_search?")
```
Paul-Bob, line 141:
```suggestion
    return true unless resource.authorization.respond_to?(:has_action_method?)
    return false unless resource.authorization.has_action_method?("search")
```

Let's just return false if the authorization is setted but `search` method is not defined.

gabrielgiroe1: Ok, thanks.

PR #2596, opened 2024-03-14
feature: configure custom save label per resource using localization

### LINKED ISSUE #2594: Customize the save button label for each resource
### Feature

Each resource should have the save button label customizable using I18n localization.

Ex:
```yml
---
en:
  avo:
    resource_translations:
      user:
        save: "Save user"
```

image

### Approach
Inside `lib/avo/resources/controls/save_button.rb` we should first try to fetch the label from `avo.resource_translations.#{resource.name}.save` and if not found fallback to `avo.save`

We can use the `default` option:

`I18n.t("avo.resource_translations.#{resource.name}.save", default: I18n.t("avo.save"))`

Paul-Bob:
# Description

Fixes #2594

Make possible to configure the save button label per each resource using localization

# Checklist:

## Screenshots & recording
```yml
---
en:
  avo:
    resource_translations:
      product:
        save: "Save the product!"
```

image

### DIFF
--- a/lib/avo/base_resource.rb
+++ b/lib/avo/base_resource.rb
@@ -177,22 +177,27 @@ module Avo
       end
 
       def name
-        default = class_name.underscore.humanize
-
-        if translation_key
-          t(translation_key, count: 1, default: default).humanize
-        else
-          default
-        end
+        name_from_translation_key(count: 1, default: class_name.underscore.humanize)
       end
       alias_method :singular_name, :name
 
       def plural_name
-        default = name.pluralize
+        name_from_translation_key(count: 2, default: name.pluralize)
+      end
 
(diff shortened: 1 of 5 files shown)


spec/rails_helper.rb
```diff
@@ -187,7 +187,7 @@ def headless_download_setup(driver)
   # config.filter_gems_from_backtrace("gem name")
 
   # https://medium.com/@velciov.vlad/retrying-flaky-tests-fae14de26c1b
-  config.default_retry_count = 2
+  config.default_retry_count = 3
```
Paul-Bob, line 190: Some time related specs fail only on CI tests, increased retry to 3 times until we fix that flaky ones

PR #2610, opened 2024-03-18
fix: collect errors from record without exception

### LINKED ISSUE #2602: Handle when there is error but no exception
### Describe the bug

It is probably wrong to do this, but it was in our app and then I was facing this error:

```ruby
NoMethodError in Avo::RoomsController#update
undefined method `include?' for nil:NilClass
Extracted source (around line #248):

      # Add the errors from the record
      @errors = @record.errors.full_messages.reject { |error| exception_message.include? error }.unshift exception_message

      succeeded
    end

avo (3.5.3) app/controllers/avo/base_controller.rb:248:in `block in perform_action_and_record_errors'
```

It is caused by a model which does something like calling this on `before_update`:

```ruby
errors.add(:field, "Error message")
```

### Steps to Reproduce
1. Add before_update method which adds an error like this.
   ```ruby
   errors.add(:field, "Error message")
   ```
2. Try to create a new record of that resource in Avo

Actually I don't know, in reality it saves the record, so maybe just show the error but not raise it?

### System configuration
**Avo version**: 3.5.3

**Rails version**: 6.1.7.7

**Ruby version**: 3.1.4

**License type**:

 - [x] Community
 - [ ] Pro
 - [ ] Advanced

 - [x] No

#### Impact

 - [x] Low impact (It's really a tiny thing that I could live with.)

#### Urgency

Paul-Bob:
# Description

Fixes #2602 

Collect errors from record even if nothing is raised. Errors can be added on callbacks without raising exception. Action fail if any error is present

# Checklist:

### DIFF
--- a/app/controllers/avo/base_controller.rb
+++ b/app/controllers/avo/base_controller.rb
@@ -245,9 +245,14 @@ module Avo
       end
 
       # Add the errors from the record
-      @errors = @record.errors.full_messages.reject { |error| exception_message.include? error }.unshift exception_message
+      @errors = @record.errors.full_messages
 
-      succeeded
+      # Remove duplicated errors
+      if exception_message.present?
+        @errors.reject { |error| exception_message.include? error }.unshift exception_message
+      end
+
+      @errors.any? ? false : succeeded
     end
 
     def model_params
--- a/app/controllers/avo/base_controller.rb
+++ b/app/controllers/avo/base_controller.rb
@@ -249,7 +249,7 @@ module Avo
 
       # Remove duplicated errors
       if exception_message.present?
-        @errors.reject { |error| exception_message.include? error }.unshift exception_message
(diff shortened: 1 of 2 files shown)


adrianthedev:
I wonder what side effects this might have for existing users.
Maybe some existing users have some errors in their models that fail silently. This PR will trigger errors for them, is that correct?

Paul-Bob:
> Maybe some existing users have some errors in their models that fail silently. This PR will trigger errors for them, is that correct?

If there are errors on the record but no exception is raised it does not fail right now, the record is saved and errors are swallowed.

If someone is in this scenario yes, same workflow will now break with this approach and show the errors, IMO that's good, if errors are present the create / update action should fail.

adrianthedev:
Got it!
Sounds good!
Let's ship it and make a note in the upgrade guide.

adrianthedev (approved):

PR #2611, opened 2024-03-18
fix: keep original view after run update or create actions

### LINKED ISSUE #2605: View `edit` vs. `update`
### Discussed in https://github.com/avo-hq/avo/discussions/2601

<div type='discussions-op-text'>

<sup>Originally posted by **krystof-k** March 16, 2024</sup>
I had a condition like this:

```ruby
field :name,
  as: :text,
  disabled: -> { view == :edit },
```

which results on disabled field on edit view, but when validation fails, it stops being disabled because the view changes from `edit` to `update`.

Is this intentional? Is there any actual difference between `edit` and `update`? If so it would be nice it to have a mention in the docs about it (where I haven't found any).</div>

Paul-Bob:
# Description

Fixes #2605

Keep original `new` / `edit` views when `create` / `update` actions are executed.

# Checklist:

- [ ] ~I have added tests that prove my fix is effective or that my feature works~

### DIFF
--- a/app/controllers/avo/application_controller.rb
+++ b/app/controllers/avo/application_controller.rb
@@ -161,13 +161,22 @@ module Avo
       @resource.class.find_scope
     end
 
+    # Force actions to have specific view
+    VIEW_ACTION_MAPPING = {
+      update: :edit,
+      create: :new
+    } unless defined? VIEW_ACTION_MAPPING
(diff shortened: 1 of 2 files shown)


adrianthedev (approved):
Looks good!

PR #2613, opened 2024-03-19
chore: update tags and fix styles

Paul-Bob:
# Description

Fixes some tags issues by updating tagify and adjust some styles.

# Checklist:

- [ ] ~I have added tests that prove my fix is effective or that my feature works~

### DIFF
--- a/app/assets/stylesheets/css/fields/tags.css
+++ b/app/assets/stylesheets/css/fields/tags.css
@@ -10,7 +10,7 @@ tags.tagify {
 }
 
 tags.tagify {
-  @apply !p-1 !gap-1 !border-gray-200;
+  @apply !ml-0 !gap-1 !border-gray-200;
 
   &.tagify--focus {
     @apply !border-gray-600;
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/css/fields/tags.css
```diff
@@ -22,10 +22,5 @@ tags.tagify {
 }
 
 tag.tagify__tag {
-  @apply text-sm !m-0;
-}
-
-/* When the tags field is used to select just one value there's a weird ZeroWidthSpace (\u200B) character that breaks the vertical spacing. */
```
adrianthedev, line 28: Why did you remove this bit @Paul-Bob?
Was it causing issues?

app/assets/stylesheets/css/fields/tags.css
```diff
@@ -10,7 +10,7 @@ tags.tagify {
 }
 
 tags.tagify {
-  @apply !p-1 !gap-1 !border-gray-200;
```
adrianthedev, line 13: The only thing that I would add here is `!px-2`. This makes the horizontal padding the same as the vertical one.

### Before

CleanShot 2024-03-19 at 07 32 12@2x

### After
CleanShot 2024-03-19 at 07 31 58@2x

app/assets/stylesheets/css/fields/tags.css
Paul-Bob: Yes, when using in select mode that was a weird behavior, it's now on the input as `min-h-10`

adrianthedev: Ok. Thanks!

PR #2623, opened 2024-03-21
fix: date_time fields timezone

### LINKED ISSUE #2510: Timezone discrepancy on datetime fields
### Describe the bug
I noticed that without setting the timezone, it displayed a different time than saved in database (supposedly using browser settings by default)
So I've set up the fields like:
 ```
    field :start_time, as: :date_time, timezone: "UTC"
    field :end_time, as: :date_time, timezone: "UTC"
```
And in the show and index views it was correct

But when entering edit, the field value is different again (1 hour difference, CEST sits in the input value and UTC is supposed to be displayed it is in data attributes)
It seems that the timezone setting is not applied to the input field

Checking the field value in the hidden input it is in CEST, but regular JS conversion to UTC does return the correct value, but the field does not

### Steps to Reproduce
1. Create date_time
2. Add timezone UTC
3. Create a record (this happens both when record is created from avo and from the regular app)
4. Compare the hours in show/index view to the edit field value. There should be a difference 

The values should always be the same, timezone setting should be applied always. It seems it is not applied in the edit

### Models and resource files
```ruby
    field :start_time, as: :date_time, timezone: "UTC"
    field :end_time, as: :date_time, timezone: "UTC"
```
The timezone is not set in avo config file. That is actually a different issue, cause when I've set up UTC there it simply did not take effect and always displayed CEST.

### System configuration
**Avo version**:
3.3.3

**Rails version**:
7.1.3

**Ruby version**:
3.2.2

**License type**:

 - [x] Community
 - [ ] Pro
 - [ ] Advanced

 - [x] No

image
image

#### Impact
This is a very important field where hours and minutes matter a lot, but for now it can be explained to the admins that they need to accounts for the time zone. Also I can just create a custom field that replaces this one for the time being.

 - [] High impact (It makes my app un-usable.)

#### Urgency

I can create a workaround in the form of a custom field.

Paul-Bob:
# Description

Fixes #2510 

Flatpickr is always using the local timezone when editing. If `timezone` option is present it will ignore it. With this PR we change the timezone of the selected input keeping the date and time values before converting it to UTC.

Force timezone option when `UTC` to `Etc/UTC` on edit due this issue https://github.com/moment/luxon/issues/1358

# Checklist:

### DIFF
--- a/app/javascript/js/controllers/fields/date_field_controller.js
+++ b/app/javascript/js/controllers/fields/date_field_controller.js
@@ -204,18 +204,22 @@ export default class extends Controller {
       return
     }
 
+    let timezonedDate = DateTime.fromISO(selectedDates[0].toISOString())
+      .setZone(this.displayTimezone, { keepLocalTime: true })
+      .setZone('UTC', { keepLocalTime: !this.relativeValue })
+
     let value
     switch (this.fieldTypeValue) {
       case 'time':
         // For time values, we should maintain the real value and format it to a time-friendly format.
-        value = DateTime.fromISO(selectedDates[0].toISOString()).setZone('UTC', { keepLocalTime: !this.relativeValue }).toFormat(RAW_TIME_FORMAT)
+        value = timezonedDate.toFormat(RAW_TIME_FORMAT)
         break
       case 'date':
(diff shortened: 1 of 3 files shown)


Paul-Bob:
Check issue when using `timezone: "UTC"` and edit use local timezone...
For some reason `UTC` date string end in `Z` and with `Etc/UTC` end in `+00:00`

Issue #2624, opened 2024-03-21
Money field (with decimals and currency)


Bergrebell:
### Feature
There is currently no comfortable way to store money in Avo. This feature should help to store money in a more robust way using the battle-tested "money-rails"-gem, while providing a good UX for the users and DX for the developers. 

**The issue with money:**
There are some issues with storing money-amounts as `float` or `numeric` that can lead to rounding errors ([link](https://blog.appsignal.com/2023/11/29/keep-your-ruby-code-maintainable-with-money-rails.html)). One workaround is to store money-amounts in their smallest unit (eg. for Dollar this is cents) instead of using floats. A popular project that uses this stratey is RubyMoney and their [money-rails](https://github.com/RubyMoney/money-rails)-gem. 

Money-rails stores money as a combination of `amount_cents` and `amount_currency`. This allows for easy conversion between currencies and the gem further helps with the formatting according to different standards. 

### Current workarounds
But using the money-rails gem in avo is currently not possible with good UX and DX. 

One option is to use a text-field to display the amount and an additional dropdown for the currency. But this leads to a whole lot of strange behavior on saving and changing the currency. 
```
field :amount, as: :text,
          format_using: -> {
            value.format
          }, update_using: -> do
        value
      end
    field :amount_currency, as: :select, options: %w[CHF EUR USD], only_on: [:edit]
```

Another option is using a number-field, but this prevents us from storing decimals. 

### Screenshots
The field could look something like that, if currencies are provided (as an option)
<img width="1132" alt="image" src="https://github.com/avo-hq/avo/assets/6041919/566a2a05-9f44-4300-9e3a-c242fd37053e">

or just do the correct formatting if no currencies are provided.

<img width="629" alt="image" src="https://github.com/avo-hq/avo/assets/6041919/c75c4bc5-692d-4731-ba82-3deb4b811158">

Bergrebell:
@adrianthedev - for the input formatting, we could use [cleavejs](https://nosir.github.io/cleave.js/) as described [here](https://dev.to/bilalbudhani/currency-input-masking-with-cleave-js-and-money-gem-in-ruby-on-rails-490l). 

but i'm not really sure if this is really necessary. the gem already does a nice job in formatting and [provides a good api](https://github.com/RubyMoney/money?tab=readme-ov-file#usage) to convert money from full-amounts or from cents into money-objects.

adrianthedev:
[...]

PR #2642, opened 2024-03-27
feature: money field

adrianthedev:
# Description

Fixes https://github.com/avo-hq/avo/issues/2624

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -163,3 +163,5 @@ gem "prefixed_ids"
 gem "mapkick-rb", "~> 0.1.4"
 
 gem "pluggy", path: "./pluggy"
+
+gem "money-rails"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -297,6 +297,15 @@ GEM
     mini_mime (1.1.5)
     mini_portile2 (2.8.5)
     minitest (5.22.2)
+    monetize (1.13.0)
+      money (~> 6.12)
+    money (6.19.0)
+      i18n (>= 0.6.4, <= 2)
+    money-rails (1.15.0)
+      activesupport (>= 3.0)
+      monetize (~> 1.9)
+      money (~> 6.13)
+      railties (>= 3.0)
     msgpack (1.7.2)
     multi_xml (0.6.0)
     mutex_m (0.2.0)
@@ -584,6 +593,7 @@ DEPENDENCIES
   listen (>= 3.5.1)
   mapkick-rb (~> 0.1.4)
   meta-tags
+  money-rails
   net-smtp
   pg (>= 0.18, < 2.0)
   pluggy!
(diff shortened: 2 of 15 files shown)


adrianthedev:
@Bergrebell this is the PoC version of it.
Try https://rubygems.org/gems/avo/versions/3.5.6.beta1. I haven't pushed beta releases in a while so if it's broken somehow, sorry. Show me and I'll fix it.

BTW, is there a way to set the "default" currencies with `money-rails`? The way it works now, is that you have to declare them on the field everytime. We'd love to pluck them from somewhere where you set them as "default" for your app.

Bergrebell:
@adrianthedev - awesome! this looks really good! i will test it this afternoon and add some feedback here! :) 

the default currency can be set in the [money.rb initializer](https://github.com/avo-hq/avo/blob/7fa7d4336767cb74c5cd06c3e593a5d993c4d765/config/initializers/money.rb)

Bergrebell:
@adrianthedev - sorry for the delay! it looks pretty good already! :) awesome that you could implement it already! there are just some minor bugs i found: 

1. the biggest one i guess is that when saving the record and changing the currency (eg. from EUR to CHF) it complains about the format being used. Not sure how to fix that though. i had the same issues when setting up the field myself. 

2. there are some visual bugs when using it in combination with a sidebar (or when there is not enough horizontal space) 

image

besides that everything looks fine so far! :) let me know if i can assist in testing again.

adrianthedev:
@Bergrebell, I fixed the layout issue. Good catch!

I can't seem to reproduce the currency issue.
Here's a recording of what I'm doing.

Here's the field declaration:

```ruby
field :price, as: :money, currencies: %w[CHF EUR USD RON SVC PEN]
```

CleanShot 2024-03-29 at 22 03 09

adrianthedev:
Hey @Bergrebell. I took a different approach for this field and bundled it separately.

Please add it to the `Gemfile` with the latest Avo version and let me know if you still see those issues.

https://docs.avohq.io/3.0/fields/money.html

Bergrebell:
@adrianthedev - in which version of avo is it currenlty available? i've tried with 3.6 but it's not there yet.

adrianthedev:
I should have been more specific. 

This new implementation we moved the field into a new gem. 
Please follow the docs from the link and install the new gem.

Bergrebell:
sorry! my bad! sometimes i am a little dyslexic 🙈 

i works super nice now! also i like the idea with packing special fields into their own gems.

Bergrebell:
by the way that's the issue i have when switching currencies. not sure if it's something on my side though (it seems to work in your case right). 

avo money_field

adrianthedev:
Maybe it's the separator?
I'll try to reproduce. 
I wonder if it's the separator, how is that solved in an non-Avo context?

Bergrebell:
okay, i got it. the issue was that i had set 

```ruby
config.locale_backend = :currency
```

in the money-initializer.

Bergrebell:
no, i think its alright. also might be somehow on purpose from money-rails. thanks adrian - the field looks great from my side! i am happy to implement it :)

adrianthedev:
Amazing @Bergrebell. Thanks!
I'm going to close this one.

PR #2645, opened 2024-03-28
Support composite primary keys on show, edit, update and destroy actions

filipegiusti:
Impressively, this seems to be all that's required to support composite primary keys in Rails 7.1

# Description
Support composite primary keys on show, edit, update and destroy actions on Rails 7.1.

`params.extract_values` was [introduced in Rails 7.1.0.beta1](https://github.com/rails/rails/commit/da7a6da4e7f5caf227420c550a8a13550dd9c682).

# Checklist:

## Screenshots & recording

Not really.

## Manual review steps

We do have a model that uses 3 enums as primary keys and I was able to view it's show page, edit and update it after the change. To implement the change in our codebase we have overwriting `set_record` on each resource that has a composite primary key.

### DIFF
--- a/app/controllers/avo/application_controller.rb
+++ b/app/controllers/avo/application_controller.rb
@@ -146,7 +146,8 @@ module Avo
     end
 
     def set_record
-      @record = @resource.find_record(params[:id], query: model_scope, params: params)
+      id = @resource.model_class.primary_key.is_a?(Array) ? params.extract_value(:id) : params[:id]
+      @record = @resource.find_record(id, query: model_scope, params:)
       @resource.hydrate(record: @record)
     end
 
--- a/app/controllers/avo/application_controller.rb
+++ b/app/controllers/avo/application_controller.rb
@@ -146,7 +146,12 @@ module Avo
     end
 
     def set_record
-      id = @resource.model_class.primary_key.is_a?(Array) ? params.extract_value(:id) : params[:id]
+      id = if @resource.model_class.primary_key.is_a?(Array) && params.respond_to?(:extract_value)
+        params.extract_value(:id)
+      else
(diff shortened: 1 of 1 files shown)


filipegiusti:
This PR probably needs a test and a check to only use `params.extract_values` on Rails 7.1 and above. I decided to create it right away in case others want to simple copy and paste the method in their resource's controller.

app/controllers/avo/application_controller.rb
```diff
@@ -142,7 +142,8 @@ def set_related_resource
     end
 
     def set_record
-      @record = @resource.find_record(params[:id], query: model_scope, params: params)
+      id = @resource.model_class.primary_key.is_a?(Array) ? params.extract_value(:id) : params[:id]
+      @record = @resource.find_record(id, query: model_scope, params:)
```
Paul-Bob, line 146: Let's keep the ` params[:id]` call for versions that doesn't have the `extract_value` method
```suggestion
id = if @resource.model_class.primary_key.is_a?(Array) && params.respond_to?(:extract_value)
  params.extract_value(:id)
else
  params[:id]
end

@record = @resource.find_record(id, query: model_scope, params:)
```

Paul-Bob (changes requested):
Thanks for the contribution @filipegiusti 

I let a comment with a change suggestion to avoid breaks on Rails versions that doesn't have the `extract_value` method.

app/controllers/avo/application_controller.rb
filipegiusti: When I opened the PR I was thinking in checking the Rails version, your suggestion is much better. Thanks.

Paul-Bob (approved):

PR #2646, opened 2024-03-29
[WIP & Experimental] fix: prevent memory bloat in associations when searchable is set to false

### LINKED ISSUE #2406: Prevent memory bloat in associations when searchable is set to false
### Feature

By default an association field has a searchable: false attribute

When attaching a record. A modal loads with a select field and the associated records. This works perfectly when the list of associated records is small. But when the collection is large it is either very slow or worse can create a significant memory bloat and crash the app. 

This is the typical example where locally it will run fine as you might not have lots of users in the development db but will crash in prod when displaying the association modal.
```ruby
field :users, as: :has_many 
``` 

### Current workarounds

After creating 2 crashes in prod, I  currently ensure that all association fields have searchable True with a custom Cop. 

I like the simplicity of a simple `select` when the collection has a reasonable size but when this size grows it should not create a memory bloat

Several solution were discussed in: https://discord.com/channels/740892036978442260/1199641108880298127

I don't have a preferred solution and understand that this could be confusing. But something should be done to avoid bloat in prod as this is a critical issue.

https://github.com/avo-hq/avo/issues/2085 will help as we will be able to set a default value for searchabe. But it will still require for the user to set it and in the mean time could experience a bloat

Paul-Bob:
# Description

This is just a POC of `hotwire_combobox` implementation to replace the attach modal select.

Can be intersting to implement `hotwire_combobox` on belongs to field and replace every select / searchable interface with `hotwire_combobox`.

As a con `hotwire_combobox` have dependency on rails 7 which breaks avo support for rails lower than 7.

Fixes #2406 

https://github.com/avo-hq/avo/assets/69730720/5dcffcb3-7fe4-4027-a92c-a0c74cf1fb05

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/app/controllers/avo/associations_controller.rb
+++ b/app/controllers/avo/associations_controller.rb
@@ -60,7 +60,7 @@ module Avo
       render turbo_stream: helpers.async_combobox_options(
         @options,
         next_page: @pagy.next,
-        display: Proc.new { |record| @attachment_resource.new(record: record).record_title }
+        display: proc { |record| @attachment_resource.new(record: record).record_title }
       )
     end
 
--- a/app/controllers/avo/associations_controller.rb
+++ b/app/controllers/avo/associations_controller.rb
@@ -2,18 +2,20 @@ require_dependency "avo/base_controller"
 
 module Avo
   class AssociationsController < BaseController
-    before_action :set_record, only: [:show, :index, :new, :create, :destroy]
+    include Pagy::Backend
+
+    before_action :set_record, only: [:show, :index, :new, :create, :destroy, :options]
     before_action :set_related_resource_name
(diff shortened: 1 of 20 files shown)


Paul-Bob:
Commented tests need refactor to test scenarios with `hotwire_combobox`

app/controllers/avo/associations_controller.rb
```diff
@@ -42,19 +44,24 @@ def show
 
     def new
       @resource.hydrate(record: @record)
+    end
 
+    def options
```
Paul-Bob, line 49: To increase performance we can serialize and encrypt the query and pass from new to options through params avoiding the query compute on each options request.

adrianthedev:
Closing to keep the PR list clean.

Issue #1555, opened 2023-01-27
Clicking "go back" from a resource always goes to the first page of the collection


davidcelis:
### Describe the bug

When using Avo's various "go back" or "cancel" buttons/links when viewing or editing a resource, the default behavior is that you get directed back to the first page of the resource's collection, regardless of how far you had to page through the collection to get to the resource you'd been viewing or editing.

### Steps to Reproduce

1. Have or create an Avo resource that has many records, at least more than one page
2. View a resource on page 2 or later
3. Click the "⬅ Go back" link
4. You are now back on page 1 and must manually page through the collection again

### Expected behavior & Actual behavior

You should be taken back to the page you had been viewing when you clicked through to the resource

### System configuration

**Avo version**: 2.24.0

**Rails version**: 7.0.4.1

**Ruby version**: 3.2.0

**License type**:

 - [x] Community
 - [ ] Pro

**Are you using Avo monkey patches or overriding views or view components?**

 - [ ] Yes. If so, please post code samples.
 - [x] No

### Screenshots or screen recordings

https://user-images.githubusercontent.com/36873/215209486-a7b9c47c-6022-4654-a420-30fec7925695.mov

#### Impact

 - [ ] High impact
 - [x] Medium impact
 - [ ] Low impact

#### Urgency

 - [ ] High urgency
 - [ ] Medium urgency
 - [x] low urgency

adrianthedev:
Hmm, yes. We might be able to get off easy by adding a `via_page` param which we pick up from the `params` object.
I'll add it to the roadmap.

Paul-Bob:
@Nevelito, thanks for opening a PR for this issue! Honestly, I think this is already working as expected in the latest Avo version, so we might not need any changes here. 

@davidcelis, could you please double-check that it's working as expected in the latest Avo version `3.14.0`?

Paul-Bob:
Closing since this was already implemented

PR #2652, opened 2024-04-01
fix: preserve page when go back from resource show page (#1555)

dmitrytrager:
# Description

Preserve page number when going back from resource show page.
Page number is fetched from referral url by parsing its query params.

Another approach can be using pagination data from index page as a parameter for generated links and then just checking `params` for `back_path`. The problem with this option is not that we will have to pass `@pagy` down through components tree, but that page number will be present in show links as well, which seems redundant.

Fixes [#1555](https://github.com/orgs/avo-hq/projects/8?pane=issue&itemId=47768610)

# Checklist:

## Manual review steps

1. Run local dummy app
2. Open project list and go to page 2
3. Select some project and open show page
4. Click 'Go back'
5. Make sure you are on page 2 

### DIFF
--- a/app/components/avo/resource_component.rb
+++ b/app/components/avo/resource_component.rb
@@ -117,8 +117,8 @@ class Avo::ResourceComponent < Avo::BaseComponent
     (params[:via_resource_class].present? || params[:via_relation_class].present?) && params[:via_record_id].present?
   end
 
-  def via_page?
-    referrer_params["page"].present?
+  def keep_referrer_params
+    { page: referrer_params["page"] }.compact
   end
 
   def render_back_button(control)
--- a/app/components/avo/resource_component.rb
+++ b/app/components/avo/resource_component.rb
@@ -118,7 +118,7 @@ class Avo::ResourceComponent < Avo::BaseComponent
   end
 
   def keep_referrer_params
-    { page: referrer_params["page"] }.compact
+    {page: referrer_params["page"]}.compact
   end
 
   def render_back_button(control)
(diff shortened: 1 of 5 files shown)


dmitrytrager:
Updated code style here a bit

Paul-Bob (changes requested):
Thanks for the initiative @dmitrytrager is a great beginning!

Let's refactor it a little bit and apply the same logic when editing.

I was thinking about a private method inside `Avo::ResourceComponent` class:
```ruby
  def referrer_params
    {
      page: referrer_params["page"]
    }.compact
  end
```

This way we can apply it to all needed paths like `**referrer_params`. Since we compact the `referrer_params` that means `nil` values will be erased from the hash and we do no longer need the `via_page?` method and verification.

We can now apply this strategy:
```ruby
# app/components/avo/views/resource_edit_component.rb
# Line 26
return helpers.resource_path(record: @resource.record, resource: @resource, **referer_params)

# Line 33
helpers.resources_path(resource: @resource, **referer_params)
```

Inside `app/components/avo/views/resource_show_component.rb` file we can replace 

```ruby
via_page? ?
        helpers.resources_path(resource: @resource, page: referrer_params["page"]) :
        helpers.resources_path(resource: @resource)
```

By:

```ruby
helpers.resources_path(resource: @resource, **referer_params)
```

Thanks again for looking into it!

Paul-Bob (approved):
Thank you @dmitrytrager !

app/components/avo/resource_component.rb
```diff
@@ -116,6 +117,10 @@ def via_resource?
     (params[:via_resource_class].present? || params[:via_relation_class].present?) && params[:via_record_id].present?
   end
 
+  def keep_referrer_params
+    { page: referrer_params["page"] }.compact
```
adrianthedev, line 121:
```suggestion
    {page: referrer_params["page"]}.compact
```

adrianthedev (approved):

PR #2656, opened 2024-04-03
fix: resource search

gabrielgiroe1:
# Description
Implement resource search functionality in cases where there are no existing policies defined for the resource.

Fixes https://discord.com/channels/740892036978442260/1223913103998193664

# Checklist:

### DIFF
--- a/app/components/avo/views/resource_index_component.rb
+++ b/app/components/avo/views/resource_index_component.rb
@@ -135,12 +135,9 @@ class Avo::Views::ResourceIndexComponent < Avo::ResourceComponent
 
   def authorized_to_search?
     # Hide the search if the authorization prevents it
-    if @resource.class.authorization_policy.present?
-      return true unless resource.authorization.respond_to?(:has_action_method?)
-      return false unless resource.authorization.has_action_method?("search")
-    else
-      return true
-    end
+    return true if @resource.class.authorization_policy.blank?
+    return true unless resource.authorization.respond_to?(:has_action_method?)
+    return false unless resource.authorization.has_action_method?("search")
 
     resource.authorization.authorize_action("search", raise_exception: false)
   end
--- a/app/components/avo/views/resource_index_component.rb
+++ b/app/components/avo/views/resource_index_component.rb
@@ -135,9 +135,12 @@ class Avo::Views::ResourceIndexComponent < Avo::ResourceComponent
 
   def authorized_to_search?
     # Hide the search if the authorization prevents it
-    return true unless resource.authorization.respond_to?(:has_action_method?)
-    return true unless resource.authorization.has_action_method?("search")
-    return false unless resource.authorization.has_action_method?("avo_search?")
+    if @resource.class.authorization_policy.present?
+      return true unless resource.authorization.respond_to?(:has_action_method?)
+      return false unless resource.authorization.has_action_method?("search")
+    else
+      return true
+    end
 
     resource.authorization.authorize_action("search", raise_exception: false)
   end
--- a/app/components/avo/views/resource_index_component.rb
+++ b/app/components/avo/views/resource_index_component.rb
@@ -135,10 +135,6 @@ class Avo::Views::ResourceIndexComponent < Avo::ResourceComponent
 
   def authorized_to_search?
     # Hide the search if the authorization prevents it
-    return true if @resource.class.authorization_policy.blank?
-    return true unless resource.authorization.respond_to?(:has_action_method?)
-    return false unless resource.authorization.has_action_method?("search")
-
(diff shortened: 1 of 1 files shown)


app/components/avo/views/resource_index_component.rb
```diff
@@ -136,7 +136,8 @@ def show_search_input
   def authorized_to_search?
     # Hide the search if the authorization prevents it
     return true unless resource.authorization.respond_to?(:has_action_method?)
-    return false unless resource.authorization.has_action_method?("search")
+    return true unless resource.authorization.has_action_method?("search")
+    return false unless resource.authorization.has_action_method?("avo_search?")
```
Paul-Bob, line 140: `has_action_method?(argument)` will pick the `argument` and use it as key to access the `Avo.configuration.authorization_methods` hash that looks like:

```ruby
      @authorization_methods = {
        index: "index?",
        show: "show?",
        edit: "edit?",
        new: "new?",
        update: "update?",
        create: "create?",
        destroy: "destroy?"
      }
```

`avo_search?` is something specific that the user can set like:

```ruby
Avo.configure do |config|
  # ...

  ## == Authorization ==
  config.authorization_methods = {
    # ...
    search: "avo_search?" # override this method
  }

  # ...
end
```

It can be even `can_search_or_not?` if the user configure it with:

```ruby
Avo.configure do |config|
  # ...

  ## == Authorization ==
  config.authorization_methods = {
    # ...
    search: "can_search_or_not?" # override this method
  }

  # ...
end
```

We should always access the `search` key to verify if the method exists.

app/components/avo/views/resource_index_component.rb
```diff
@@ -135,8 +135,12 @@ def show_search_input
...
   def authorized_to_search?
     # Hide the search if the authorization prevents it
-    return true unless resource.authorization.respond_to?(:has_action_method?)
-    return false unless resource.authorization.has_action_method?("search")
+    if @resource.class.authorization_policy.present?
+      return true unless resource.authorization.respond_to?(:has_action_method?)
+      return false unless resource.authorization.has_action_method?("search")
+    else
+      return true
+    end
```
Paul-Bob, line 143:
```suggestion
    return true if @resource.class.authorization_policy.blank?
    return true unless resource.authorization.respond_to?(:has_action_method?)
    return false unless resource.authorization.has_action_method?("search")
```

Let's favor the use of guard clauses over  "if-else" statements. Guard clauses enhance code readability and maintainability.

Paul-Bob (changes requested):
Hello @gabrielgiroe1 I let a comment about readability, otherwise the logic seems correct.

app/components/avo/views/resource_index_component.rb
gabrielgiroe1: Thanks. I will update it.

Paul-Bob:
@gabrielgiroe1 lets try this approach, I think this will have the desired output:

```ruby
  def authorized_to_search?
    resource.authorization.authorize_action("search", raise_exception: false)
  end
```

Can you please check this scenarios? 

1. [ ] On `avo`, without avo-pro installed resource search should be working.
2. [ ] On `avo-pro` with `config.authorization_client = nil` should be working.
3. [ ] On `avo-pro` with `pundit` as client but without policy configured for the resource should be working.
4. [ ] On `avo-pro` with `pundit` as client and without the search method defined should search should not work.
5. [ ] On `avo-pro` with `pundit` as client and with the `search` method defined returning false should search should not work.
6. [ ] On `avo-pro` with `pundit` as client and with the `search` method defined returning true search should work.

Paul-Bob (changes requested):
@gabrielgiroe1 just noticed that we can simplify the method.

Lets try this approach, I think this will have the desired output:

```ruby
  def authorized_to_search?
    resource.authorization.authorize_action("search", raise_exception: false)
  end
```

Can you please check this scenarios? 

1. [x] On `avo`, without avo-pro installed resource search should be working.
2. [x] On `avo-pro` with `config.authorization_client = nil` should be working.
3. [x] On `avo-pro` with `pundit` as client but without policy configured for the resource should be working.
4. [x] On `avo-pro` with `pundit` as client and without the search method defined should search should not work.
5. [x] On `avo-pro` with `pundit` as client and with the `search` method defined returning false should search should not work.
6. [x] On `avo-pro` with `pundit` as client and with the `search` method defined returning true search should work.

gabrielgiroe1:
Yes, you are right. I tested it with these changes and it work as expected.

Paul-Bob (approved):
Thanks @gabrielgiroe1 it's looking great!

PR #2682, opened 2024-04-12
chore: performance tweak on boolean group field and select field

Paul-Bob:
# Description

Using instance variable instead the reader method slightly increase performance. Reader methods that are not used outside the class can be removed.


adrianthedev (approved):
LGTM

Issue #1328, opened 2022-10-18
Audit logging


adrianthedev:
> [!NOTE]  
> Enterprise feature - https://github.com/avo-hq/avo-audit_logging

### Feature

Know who did what in Avo.

### Approach

This feature should:

 - [x] record which user edited a record and which fields he updated
 - [x] record when a user runs an action on which record and with which field values
 - [x] display these changes in a separate view (maybe as associations?) (maybe a new timeline view?)
 - [x] enable the user to rollback the record changes (not action changes)
 - [ ] maybe custom implementation (last resort)
 - [ ] maybe custom audited model

```[tasklist]
### Related
- [ ] https://github.com/avo-hq/avo/issues/1322
```

### Gems to test out:

- https://github.com/collectiveidea/audited (preferred)
- https://github.com/paper-trail-gem/paper_trail

### Related

We could use the same thing in https://github.com/avo-hq/avo/issues/1322

"We’ve seen some particularities that make Audited mostly used to “Active Record User Stamping”, (checking the actions taken by a particular user), while Paper Trail is referenced as being in the “Active Record Versioning Area” (getting a timeline of changes in a record due to the larger set of options that Paper Trail offers in terms of functions and methods)."

https://revs.runtime-revolution.com/the-battle-for-auditing-and-versioning-in-rails-audited-vs-paper-trail-17ad0011ecd9

![CleanShot 2024-04-16 at 16 25 31@2x](https://github.com/avo-hq/avo/assets/1334409/a0350dc3-010a-4ee5-b473-cf4dd10eaf1a)

segiddins:
I did some audit log work for the RubyGems.org admin site in https://github.com/rubygems/rubygems.org/pull/3426, if that helps at all with inspiration!

adrianthedev:
That's interesting work you're doing there with the `ActionHandler` @segiddins. That will definitely come in handy when we'll implement our own logging feature. 
Thanks for sharing!

Was it difficult to navigate the codebase and understand how actions work?

segiddins:
Not too difficult, though I did spend a bunch of time reading through the code (instead of being able to use the docs 100%). Overall I'd say it was easy to figure out what I needed in order to do my hackery :)

adrianthedev:
Thanks for sharing!

MrJoy:
[...]

PR #2703, opened 2024-04-18
feature: audit logs

Paul-Bob:
# Description

Related to #1328 

Audit logging calls.

# Checklist:


lib/avo/configuration.rb
```diff
@@ -160,7 +162,9 @@ def license=(value)
     def license
       gems = Gem::Specification.map {|gem| gem.name}
 
-      @license ||= if gems.include?("avo-advanced")
+      @license ||= if gems.include?("avo-audits")
+        "enterprise"
```
yorsant, line 166: Why? 😢

yorsant: Can you consider moving the audit feature to the advanced plan?

app/components/avo/sidebar_component.html.erb
```diff
@@ -54,6 +54,7 @@
           <% end %>
         <% end %>
 
+
```
Paul-Bob, line 57:
```suggestion
```

lib/avo/configuration.rb
```diff
@@ -221,6 +225,22 @@ def default_logger
...
 
+    def database_exists?
+      ActiveRecord::Base.connection
+    rescue ActiveRecord::NoDatabaseError
+      false
+    else
+      true
+    end
+
+
```
Paul-Bob, line 236:
```suggestion

```

lib/avo/configuration.rb
```diff
@@ -221,6 +225,20 @@ def default_logger
...
     end
 
+    def database_exists?
+      ActiveRecord::Base.connection
+    rescue ActiveRecord::NoDatabaseError
+      false
+    else
+      true
+    end
+    def audit?
```
Paul-Bob, line 235:
```suggestion
    end

    def audit?
```

lib/avo/configuration.rb
adrianthedev: Hey @yorsant.
Apologies for the late reply.

The audit logging feature is, generally speaking, an enterprise feature.

As much as we want to give features away or make them cheaper, we can't do that. It takes a lot of time and effort to build Avo.
We love the self-serve product (Pro, Advanced) but the reality is that we can't survive and improve our business without having some big customers. We've heard from customers that they saved tens and even hundreds of thousands of dollars by using Avo and that they would like to pay more.
If we don't have any Enterprise features, we can't sell Enterprise subscriptions.

This all comes back to us being able to build a robust business and being able to ship more value through our products.

Thanks for asking the question.

xeron: FWIW you can build audit log yourself using papertrail and avo resources.

app/controllers/avo/application_controller.rb
```diff
@@ -25,6 +25,7 @@ class ApplicationController < ::ActionController::Base
     before_action :set_view
     before_action :set_sidebar_open
     before_action :set_stylesheet_assets_path
+    before_action :set_paper_trail_whodunnit, if: -> { defined? PaperTrail }
```
adrianthedev, line 28: Can't we add this ourselves in the audit logging gem?

lib/generators/avo/templates/initializer/avo.tt
```diff
@@ -100,6 +100,9 @@ Avo.configure do |config|
   #   file_logger
   # }
 
+  ## == Audit ==
+  # config.audit = false
```
adrianthedev, line 104: Let's think about the configuration of this gem a bit more. true/false is a bit terse and we'll for sure have more configs on that gem.
How about moving this config as `config.enabled = true|false` on that gem's configuration?

lib/avo/configuration.rb
```diff
@@ -231,6 +235,21 @@ def default_logger
...
 
+    def database_exists?
+      ActiveRecord::Base.connection
+    rescue ActiveRecord::NoDatabaseError
+      false
+    else
+      true
+    end
+
+    def audit?
```
adrianthedev, line 246: I don't see where we use this method, probably in the gem itself?
If so, how about moving it there so there's little noise about it here?

lib/avo/configuration.rb
```diff
@@ -172,7 +174,9 @@ def license=(value)
     def license
       gems = Gem::Specification.map {|gem| gem.name}
 
-      @license ||= if gems.include?("avo-advanced")
+      @license ||= if gems.include?("avo-audit_logging")
+        "enterprise"
```
adrianthedev, line 178: This feature will not be bundled in a gem, so let's not force the license to enterprise.

app/controllers/avo/base_controller.rb
```diff
@@ -570,10 +582,6 @@ def set_pagy_locale
       @pagy_locale = locale.to_s || Avo.configuration.default_locale || "en"
     end
 
-    def safe_call(method)
```
adrianthedev, line 573: Don't we need this? And shouldn't it take more args?

adrianthedev (changes requested):

lib/generators/avo/templates/initializer/avo.tt
Paul-Bob: >How about moving this config as `config.enabled = true|false` on that gem's configuration?

Well, that's a really good idea to have a configuration file for each gem!

lib/avo/configuration.rb
Paul-Bob: Yes, I'm moving the audit configuration to the gem and will revert this change, thanks for pointing it

app/controllers/avo/base_controller.rb
Paul-Bob: I guess you missed this change https://github.com/avo-hq/avo/pull/2703/files#diff-27ec89c6677b158fe3a800b268c9c196b5409f439315240504626f4c8bf81bc7R140

The method was extracted to be accessible in more places

app/controllers/avo/application_controller.rb
Paul-Bob: Makes sense, done

lib/tasks/avo_tasks.rake
```diff
@@ -81,7 +81,7 @@ task "avo:sym_link" do
 
   gem_paths = `bundle list --paths 2>/dev/null`.split("\n")
 
-  ["avo-advanced", "avo-pro", "avo-dynamic_filters", "avo-dashboards", "avo-menu", "avo-kanban"].each do |gem|
+  ["avo-advanced", "avo-pro", "avo-dynamic_filters", "avo-dashboards", "avo-menu", "avo-kanban", "avo-audit_logging"].each do |gem|
```
Paul-Bob, line 84:
```suggestion
  ["avo-advanced", "avo-pro", "avo-dynamic_filters", "avo-dashboards", "avo-menu", "avo-kanban"].each do |gem|
```

PR #2722, opened 2024-04-25
feature: profile and cover photos

adrianthedev:
# Description

Adds a cover and profile photo to the show view of the record.

CleanShot 2024-04-25 at 11 46 47@2x

# Todo:

- [ ] Add cover images to the index/edit page
- [ ] add support for plain paths (interpolated with the record data) (Paul's suggestion)

# Checklist:

## Manual review steps

1. Step 1
1. Step 2


Paul-Bob (changes requested):
Looking great!

I have one concern about the way to declare the cover/photo components. IMO it should not rely on existing fields, and pass the photo URL directly on the API like grid view cover.

Pros: 
- Any image can be used even an external URL
- Custom path helpers can be used to get the image URL
- No need of field declaration if user don't want to have that fields.

```ruby
self.cover_photo = -> do
  {
    # size: :sm,
    source: main_app.url_for(record.cover_photo.url)
  }
end
```

adrianthedev:
yup! makes sense!
Thanks for the review!

adrianthedev:
Let's look into memoization of these obejcts (`profile_photo` & `cover_photo`) from the concern.

adrianthedev:
Implements RSpec tests for Avo's event page cover and profile photos components, ensuring proper visibility under various conditions.

- **Adds component tests** for both `Avo::CoverPhotoComponent` and `Avo::ProfilePhotoComponent` to verify rendering logic based on the presence and visibility of photos. These tests cover scenarios where photos are present, absent, not visible in the current view, and dynamically sourced.
- **Updates feature tests** in `spec/features/avo/cover_profile_photos_spec.rb` to include scenarios for an event page without cover and profile photos, with only a cover photo, with only a profile photo, and with both photos. This ensures the application behaves as expected under each condition.


---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/avo-hq/avo/pull/2722?shareId=76ee8294-b3dd-4b08-b5a6-3fb721ad01af).

Paul-Bob (approved):

PR #2741, opened 2024-05-03
fix: check if ActionText::RichText is defined

### LINKED ISSUE #2738: uninitialized constant Avo::Fields::TrixField::ActionText in avo (3.6.4)
### Describe the bug

When editing a resource that has had a working Trix editor with version 3.6.4, we encounter an undefined constant referring to ActionText:

 lib/avo/fields/trix_field.rb:28:in `is_action_text?'

### Steps to Reproduce
1. Set up project without Activetext, latest avo
2. Create a trix editor field
3. Open the edit form for a resource with text already set via trix editor

Actiontext installation should be optional or there should be clear documentation requiring it. Installing the actiontext gem and enabling the engine in config/application.rb is a sufficient workaround to allow the system to work again.

### System configuration

**Avo version**:
3.6.4

**Rails version**:
(7.1.3.2, notably without actiontext as it was upgraded from an earlier Rails app)

**Ruby version**:
3.3.1

**License type**:

 - [ ] Community
 - [ ] Pro
 - [X] Advanced

 - [X] No
 
#### Impact

#### Urgency

 - [X] Low urgency (It can wait. I just wanted you to know about it.)

Paul-Bob:
# Description

Check if `ActionText::RichText` is defined before use it.

Fixes #2738

# Checklist:

### DIFF
--- a/lib/avo/fields/trix_field.rb
+++ b/lib/avo/fields/trix_field.rb
@@ -23,7 +23,7 @@ module Avo
 
       # Identify if field is bonded to a rich text model attribute
       def is_action_text?
-        return false if record.nil? || !record.respond_to?(id)
+        return false if !defined?(ActionText::RichText) || record.nil? || !record.respond_to?(id)
 
         record.send(id).is_a?(ActionText::RichText)
       end


adrianthedev (commented):
This will fail silently, right? Maybe we should not add this to the long list of things that fail silently?

Paul-Bob:
> This will fail silently, right? Maybe we should not add this to the long list of things that fail silently?

What fail, can you explain please?

adrianthedev:
Fail silently, meaning that it won't work and the user will get no feedback about it.
Then they will be frustrated on why that's happening.

Paul-Bob:
> Fail silently, meaning that it won't work and the user will get no feedback about it. Then they will be frustrated on why that's happening.

Got it. This doesn't belong to that scenarios. This method (`is_action_text?`) identify if field is bonded to a rich text model attribute and return `true` or `false` whenever the field attribute is using `ActionText` or not respectively. This PR prevents breaking the app if  `ActionText::RichText` is not defined when checking `record.send(id).is_a?(ActionText::RichText)`

If `ActionText::RichText` is not defined we return false early because 100% there is 0 attributes using `ActionText`

adrianthedev:
Ok. THanks for the explanation!

PR #2754, opened 2024-05-08
feature: JS for dashboard wide filters

gabrielgiroe1:
# Description
Add JavaScript for: [dashboard-wide filters](https://github.com/avo-hq/avo-dashboards/pull/36)

# Checklist:

## Manual review steps

1. Step 1
1. Step 2


adrianthedev (changes requested):
Let's move this into the existing dashboads cards controller

adrianthedev (changes requested):
I made a mistake with the last decision to move it into the dashboard_card controller.

The mistake is that we are attaching this controller to elements that don't need to be refreshed (the previous behavior).

So, please let's move this into a `card_filters_controller`.

Sorry for the back and forth

app/javascript/js/controllers/dashboard_card_controller.js
```diff
@@ -1,9 +1,24 @@
...
+
+function updateQueryParam(url, key, value) {
+  const uri = new URI(url)
+  uri.setQuery(key, value)
+
+  return uri.toString()
+}
+
+function getDateParam(url) {
+  const queryString = url.split('?')[1]
```
adrianthedev, line 12: Why are we splitting this as a string?
Can't we use `URI` as well to pick up that param?

Please let's do that.

gabrielgiroe1: Yes, sure! I will update it using URI.

adrianthedev (approved):

Issue #1929, opened 2023-08-31
Incorrect label for belongs_to selects


xeron:
### Describe the bug

When editing or creating new records with `belongs_to` fields label's `for` value doesn't match the input's `id` value.

In Chome inspector Issues tab the following messages appear:

> The label's for attribute doesn't match any element id. This might prevent the browser from correctly autofilling the form and accessibility tools from working correctly.
> To fix this issue, make sure the label's for attribute references the correct id of a form field.

![Screenshot 2023-08-31 at 4 44 11 AM](https://github.com/avo-hq/avo/assets/110569/da54f3ec-a26e-41d7-a82d-9f9304002094)

### Steps to Reproduce

1. Create 2 models, with `has_many` / `belongs_to` relationship
2. Create avo resource with `belongs_to` field
3. Go to new or edit view of the object which uses `belongs_to`
4. Try clicking a label for `belongs_to` field
5. Correct select input won't be focused

### Expected behavior & Actual behavior

This config generates the following HTML:

```html
…
<label for="result_event">Event</label>
…
<select … name="result[event_id]" id="result_event_id"><option value="">Choose an option</option>
<option value="…">Event 1</option>
…
```

You can see label's `for` is `result_event` but input's `id` is `result_event_id`.

Expected to see the same id in label's `for` and in input's `id`.

### Models and resource files

```ruby
class Event < ApplicationRecord
  has_many :results, dependent: :destroy
end
```

```ruby
class Result < ApplicationRecord
  belongs_to :event, touch: true
end
```

```ruby
class ResultResource < Avo::BaseResource
…
  field :event, as: :belongs_to
…
end
```

### System configuration
**Avo version**: 2.40.0
**Rails version**: 7.0.7.2
**Ruby version**: 3.2.2

**License type**:

 - [x] Community
 - [ ] Pro

**Are you using Avo monkey patches, overriding views or view components?**

 - [ ] Yes. If so, please post code samples.
 - [x] No

### Screenshots or screen recordings

N/A

### Additional context

N/A

#### Impact

 - [ ] High impact (It makes my app un-usable.)
 - [ ] Medium impact (I'm annoyed, but I'll live.)
 - [x] Low impact (It's really a tiny thing that I could live with.)

#### Urgency

 - [ ] High urgency (I can't continue development without it.)
 - [ ] Medium urgency (I found a workaround, but I'd love to have it fixed.)
 - [x] Low urgency (It can wait. I just wanted you to know about it.)

[...]

PR #2758, opened 2024-05-09
fix: wrong label on `belongs_to`field

adrianthedev:
# Description

Fixes https://github.com/avo-hq/avo/issues/1929

# Checklist:

### DIFF
--- a/lib/avo/fields/belongs_to_field.rb
+++ b/lib/avo/fields/belongs_to_field.rb
@@ -265,6 +265,10 @@ module Avo
         @can_create
       end
 
+      def form_field_label
+        id
+      end
+
       private
 
       def get_model_class(record)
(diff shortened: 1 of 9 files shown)


xeron:
@adrianthedev this doesn't fix the issue. ID is still different:

```
…
<label for="result_driver">Driver</label>
…
<select … id="result_driver_id"><option value="">Choose an option</option></select>
…
```

lib/avo/fields/belongs_to_field.rb
```diff
@@ -265,6 +265,10 @@ def can_create?
         @can_create
       end
 
+      def form_field_label
+        id
```
xeron, line 269: Should have been `"#{id}_id"` probably?

adrianthedev:
You're right @xeron.
I refactored the code but didn't apply the fix.

I took this oportunity to also improve & slim down the JS controller.

adrianthedev:
It'll be fixed in the next release.

Issue #2707, opened 2024-04-18
Unfriendly behavior when Action status message is very long.


MrJoy:
### Describe the bug

I just spent about 20 minutes trying to figure out why I couldn't get a toast with the result of an Action.  It turns out the action was doing this:

```ruby
error("Unknown error: #{e.message}")
```

And the resulting error message length was causing the session cookie to exceed 4KB.  I noticed that via a warning in the browser console.

I assume this is due to how Rails handles flash messages.

### Steps to Reproduce
1. Create an action, and expose it on the resource.  The action should produce a response that is very long.  (In my case, `e.message.length` was 1,644.)
2. Run the action.

### Expected behavior & Actual behavior

It would be nice if Avo would report that something went wrong and the result couldn't be presented.  Or, perhaps, if it truncated the message before tossing it in the session hash (although finding an acceptable limit for that could be challenging).

It would be even nicer if the response to the action wasn't conveyed via the session object.  At least, not when the response is a normal Turbo response and not a redirect or download.

### System configuration
**Avo version**: 3.6.1

**Rails version**: 7.1.3.2

**Ruby version**: 3.2.2

**License type**:

 - [ ] Community
 - [ ] Pro
 - [x] Advanced

**Are you using Avo monkey patches, overriding views or view components?**

 - [ ] Yes. If so, please post code samples.
 - [x] No

#### Impact

 - [ ] High impact (It makes my app un-usable.)
 - [x] Medium impact (I'm annoyed, but I'll live.)
 - [ ] Low impact (It's really a tiny thing that I could live with.)

#### Urgency

 - [ ] High urgency (I can't continue development without it.)
 - [ ] Medium urgency (I found a workaround, but I'd love to have it fixed.)
 - [x] Low urgency (It can wait. I just wanted you to know about it.)

adrianthedev:
Hey @MrJoy.
I know it's frustrating when things fail silently, but it's incredibly hard for us to detect these kinds of issues in advance.
I mean... I didn't really know about that limit, and nowhere in the Rails docs is that mentioned as an issue that we should take into account.

> It would be even nicer if the response to the action wasn't conveyed via the session object. At least, not when the response is a normal Turbo response and not a redirect or download.

The response here is pretty much "default", wouldn't you say?

```ruby
def update_fail_action
  respond_to do |format|
    flash.now[:error] = update_fail_message
    format.html { render :edit, status: :unprocessable_entity }
  end
end
```

This is pretty much what we do.

[...]

PR #2773, opened 2024-05-16
fix: truncate long action messages to avoid overflow

gabrielgiroe1:
# Description
Implement message truncation to 320 characters on action.

Fixes https://github.com/avo-hq/avo/issues/2707

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/lib/avo/base_action.rb
+++ b/lib/avo/base_action.rb
@@ -292,6 +292,7 @@ module Avo
     private
 
     def add_message(body, type = :info)
+      body = body.truncate(320)
       response[:messages] << {
         type: type,
         body: body
--- a/lib/avo/base_action.rb
+++ b/lib/avo/base_action.rb
@@ -294,7 +294,7 @@ module Avo
     def add_message(body, type = :info)
       response[:messages] << {
         type: type,
-        body: body&.truncate(321)
+        body: body&.truncate(320)
       }
     end
   end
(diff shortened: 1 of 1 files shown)


Paul-Bob (approved):
I'm approving the PR since it's respecting the request, let's just directly truncate and avoid unnecessary `body` re-attribution as a performance tweak.

# More complex review
This solution may work sometimes but do not fix the `CookieOverflow` problem. We're storing the messages on flash hash and the default `cookie_store` has a limit of 4kb. When that limit is exceeded `CookieOverflow` is raised.

```ruby
  def handle(**args)
    succeed "🥑"*1600
    warn "🥑"*1600
    inform "🥑"*1600
    error "🥑"*1600
    return
  end
```

With this truncate solution the above code snippet still breaking because the sum of the 4 truncated messages exceed the 4kb limit.

Sources: 
- https://stackoverflow.com/questions/9473808/cookie-overflow-in-rails-application
- https://naoj-gior.medium.com/flash-messages-and-cookie-overflow-c1886c9c3054

adrianthedev (approved):

JeremasPosta:
Hey @adrianthedev, it is possible to make this feature optional?

It is breaking our implementation because of:
 - We don't always send `strings` to `error()`, so `truncate` fails. (Easy to fix on our end with just a `.to_s`)
 - We _need_ to display long messages sometimes, we even ejected the `alert_component` partial and we make it vanish after a custom length-based time.

PR #2789, opened 2024-05-22
Rails 8 support

peterberkenbosch:
`._reflections` returns a Hash with symbols as keys since https://github.com/rails/rails/commit/d9a54e41ab147a7e98559ffddda8ffc0fc059daa, and using the `to_s` is causing the `belongs_to` field to error with 

```undefined method `klass' for nil``` 

when trying to create a new record.

# Description

Fixes # (issue)

# Checklist:

## Manual review steps

1. `rails new demo --main`
2. `cd demo`
3. `bin/rails app:template LOCATION='https://avohq.io/app-template'`
4. `bin/rails g model Category name`
5. `bin/rails g model Book name category:references`
6. add category field to the book resource with `belongs_to`
7. try adding a new book. This will error without this change.
8. 


adrianthedev:
I love that change!
Would you be able to point out the PR in the Rails repo where that change has been made so we can track it as well?

Also, should we use some indifferent access or something so we can support all pre-rails 8 apps too?

peterberkenbosch:
@adrianthedev the change is introduced here: https://github.com/rails/rails/commit/d9a54e41ab147a7e98559ffddda8ffc0fc059daa

We could do `with_indifferent_access` for sure.

peterberkenbosch:
There might be some more spots that will be an issue, searching the codebase where the reflections hash is used and is getting passed a String as the key.

adrianthedev:
This will be helpful to have these tests pass https://github.com/avo-hq/avo/pull/2792

peterberkenbosch:
@adrianthedev I will be able to spend some time on that coming evenings / weekend.

Issue #2159, opened 2023-12-18
Redesign the actions dropdown


adrianthedev:
Use this as an inspiration.

https://twitter.com/jamesm/status/1626189716935823360
https://twitter.com/jamesm/status/1617577097006022656
https://twitter.com/jamesm/status/1622702890912456704
https://twitter.com/jamesm/status/1623643221665411074

https://dribbble.com/shots/20481335--CelestiaScan

![Image](https://github.com/avo-hq/avo/assets/1334409/3c99862a-393b-48ae-a739-72bb43c76875)

adrianthedev:
Fixed in Avo 4.

PR #2806, opened 2024-05-23
ensure action icon does not shrink

adrienpoly:
when the names of the actions are very long the icon shrinks (yeah I know they are very long 😄 )

CleanShot 2024-05-24 at 01 46 58@2x

This PR prevents that


adrianthedev:
Thank you @adrienpoly!
We do want to [redesign](https://github.com/avo-hq/avo/issues/2159) those bulky dropdowns at some point.

PR #2813, opened 2024-05-27
fix: add model-unrelated errors to alerts

adrianthedev:
# Description

The create failing action already has a `turbo_stream` response.
This PR adds it to the update action as well.

We also add any exception error to the response.

# Checklist:


app/controllers/avo/base_controller.rb
```diff
@@ -231,7 +231,17 @@ def perform_action_and_record_errors(&block)
...
       if exception_message.present?
-        @errors = @errors.reject { |error| exception_message.include? error }.unshift exception_message
+        @errors = @errors.reject { |error| exception_message.include? error }
+      end
+
+      # Figure out if we have to output the exception_message
+      # Usually it means that it's not a validation error but something else
+      if exception_message.present?
+        exception_is_validation = @errors.select { |error| exception_message.include? error }.present?
+      end
```
Paul-Bob, line 241: Since this code `@errors = @errors.reject { |error| exception_message.include? error }` runs before this one `exception_is_validation = @errors.select { |error| exception_message.include? error }.present?` will the second one ever be true?

Issue #2672, opened 2024-04-11
Action list dividers with labels


adrianthedev:
### Feature

Dividers are great but they would be better if they had a label on it so folks can group the actions better.

### Screenshots or screen recordings

Card dividers
![CleanShot 2024-04-11 at 17 50 38@2x](https://github.com/avo-hq/avo/assets/1334409/03d87400-a6a7-4e20-aaf3-4a7739de69b8)

PR #2821, opened 2024-05-29
Action list dividers with labels

gabrielgiroe1:
# Description

Fixes https://github.com/avo-hq/avo/issues/2672
[Test for divider in actions_list in show_controls](https://github.com/avo-hq/avo-advanced/pull/35)
Screenshot 2024-05-29 at 22 07 49

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/app/components/avo/actions_component.rb
+++ b/app/components/avo/actions_component.rb
@@ -88,14 +88,18 @@ class Avo::ActionsComponent < ViewComponent::Base
   def render_item(action)
     case action
     when Avo::Divider
-      label = action.label.is_a?(Hash) ? action.label[:label] : nil
-      render Avo::DividerComponent.new(label)
+      render_divider(action)
     else
       render_action_link(action)
     end
   end
 
   private
+
+  def render_divider(action)
+    label = action.label.is_a?(Hash) ? action.label[:label] : nil
+    render Avo::DividerComponent.new(label)
+  end
 
   def render_action_link(action)
     link_to action_path(action),
--- a/app/components/avo/actions_component.rb
+++ b/app/components/avo/actions_component.rb
@@ -86,7 +86,8 @@ class Avo::ActionsComponent < ViewComponent::Base
   end
 
   def render_item(action)
-    if action.is_a?(Avo::Divider)
+    case action
+    when Avo::Divider
       label = action.label.is_a?(Hash) ? action.label[:label] : nil
       render Avo::DividerComponent.new(label)
     else
--- a/app/components/avo/actions_component.rb
+++ b/app/components/avo/actions_component.rb
@@ -89,7 +89,7 @@ class Avo::ActionsComponent < ViewComponent::Base
     case action
     when Avo::Divider
       render_divider(action)
-    else
+    when Avo::BaseAction
       render_action_link(action)
     end
   end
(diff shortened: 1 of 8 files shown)


app/components/avo/actions_component.rb
```diff
@@ -86,8 +86,9 @@ def icon(action)
   end
 
   def render_item(action)
-    if action.is_a?(Avo::DividerComponent)
-      render Avo::DividerComponent.new
+    if action.is_a?(Avo::Divider)
+      label = action.label.is_a?(Hash) ? action.label[:label] : nil
+      render Avo::DividerComponent.new(label)
     else
       render_action_link(action)
```
adrianthedev, line 93: Let's use a `case` statement and instead of `else`, check for an actual action class.

spec/dummy/app/avo/resources/user.rb
```diff
@@ -67,7 +67,7 @@ def fields
   def actions
     action Avo::Actions::ToggleInactive, icon: "heroicons/outline/globe"
     action Avo::Actions::ToggleAdmin
-    divider
+    divider label: "test if it work"
```
adrianthedev, line 70:
```suggestion
    divider label: "Other actions"
```

adrianthedev (changes requested):
Good job to get it to this stage.

Let's add tests to check the following:

- divider works in the `def actions` method
- divider works in `actions_list` in `show_controls`
- test for a divider with and without label

app/components/avo/actions_component.rb
gabrielgiroe1: Ok

spec/features/avo/divider_spec.rb
```diff
@@ -0,0 +1,24 @@
+require "rails_helper"
+
+RSpec.feature "Divider", type: :feature do
+  let!(:user) { create :user }
+
+  before do
+    visit "/admin/resources/users"
+    click_on "Actions"
+    @parsed_body = Nokogiri::HTML(page.body)
+    @dividers = @parsed_body.css(".relative.col-span-full.border-t")
```
adrianthedev, line 10: We don't need to parse the response like so with Nokogiri. capybara/rspec should have something already.

```ruby
expect(page).to have SOMETHING

within(find("SOME_SELECTOR")) do
  find("SOMETHING_ELSE")
end
```

Please check other feature tests.

Also, let's ensure the dividers and actions are in order. So we're not checking just for presence on the page. 

[This](https://www.betterspecs.org/) is a good resource aboout how to write better specs. It doesn't necesarily have guides on selectors, but you can see some best practices.

app/components/avo/actions_component.rb
```diff
@@ -86,15 +86,21 @@ def icon(action)
...
+      render_divider(action)
+    when Avo::BaseAction
       render_action_link(action)
     end
   end
 
   private
 
+  def render_divider(action)
+    label = action.label.is_a?(Hash) ? action.label[:label] : nil
```
adrianthedev, line 100: Why would the `label` be a hash?

adrianthedev (changes requested):
It's getting close to merging.

I left a question and some feedback about changing the specs.

Issue #2799, opened 2024-05-23
`self.authorization_policy` doesn't apply to left side menu `Policy#index?` action.


mtomov:
### Steps to Reproduce

1. Have an `application_policy`

```ruby
class ApplicationPolicy
  attr_reader :user, :record

  def initialize(user, record)
    @user = user
    @record = record
  end

  def index?
    false # this is important
  end
end
```

2. Create a custom policy

```ruby
class AvoPolicy < ApplicationPolicy
  class Scope < Scope
    def resolve
      scope.all
    end
  end

  def index?
    true # note this
  end
end
```

3. Assign the custom policy to a resource

```ruby
class Avo::Resources::Recording < Avo::BaseResource
  self.authorization_policy = AvoPolicy
end
```

4. Go to admin UI, and check the menu on the left - the resource won't be visible. 

5. Change

```ruby
  def index?
    false # this is important
  end
```
to `true`, -> and the resource will be visible in the left menu.

### System configuration
**Avo version**:

**Rails version**:

**Ruby version**:

**License type**:

 - [ ] Community
 - [ ] Pro
 - [x] Advanced
#### Urgency

 - [ ] High urgency (I can't continue development without it.)
 - [ ] Medium urgency (I found a workaround, but I'd love to have it fixed.)
 - [x] Low urgency (It can wait. I just wanted you to know about it.)

adrianthedev:
Hey @mtomov can I please ask you to create a repo with those instructions so we can quickly clone and run it?
Thank you!

PR #2827, opened 2024-06-01
Use correct policy class in search controller

martinemde:
# Description

The authorization_policy on the resource should be used when set in the search controller.

Fixes something similar to #2799 (related to #2805)

#2805 fixed most of the test failures for us, but this PR is still necessary to get them all passing.

Here's the [test run](https://github.com/rubygems/rubygems.org/actions/runs/9325667018/job/25673035487?pr=4745) that failed.
This is the commit without the fix applied that fails: https://github.com/rubygems/rubygems.org/pull/4745/commits/4726858ea8a102e3ecf3bec3c43a4c5d412048b0

Specifically, the search controller test failure is caused in [test/integration/avo/gem_name_reservations_controller_test.rb:15](https://github.com/rubygems/rubygems.org/blob/8c6080fbc313712c134b3f433297695640209f2b/test/integration/avo/gem_name_reservations_controller_test.rb#L15)

# Checklist:

## Manual review steps

1. This change was necessary to get rubygems.org tests working with a custom authorization policy.

### DIFF
--- a/app/controllers/avo/search_controller.rb
+++ b/app/controllers/avo/search_controller.rb
@@ -20,7 +20,11 @@ module Avo
       resources
         .map do |resource|
           # Apply authorization
-          next unless @authorization.set_record(resource.model_class).authorize_action(:search, raise_exception: false)
+          next unless @authorization.set_record(resource.model_class).authorize_action(
+            :search,
+            policy_class: resource.authorization_policy,
+            raise_exception: false
+          )
           # Filter out the models without a search_query
           next if resource.search_query.nil?
 


martinemde:
I'm concerned that all the places where `authorize_action` is called might need to have this added, which would imply a refactor is needed. I see that there aren't any tests that use a resource with a custom policy so I couldn't easy add something for this.

adrianthedev:
Thank you @martinemde
We'll add some tests for this use-case in https://github.com/avo-hq/avo/issues/2828

AFAIK you are still using Avo 2 on rubygems.org
This PR will be merged in `main` which will be released for Avo 3.x

We'll need to backport it to https://github.com/avo-hq/avo/tree/2.x

martinemde:
Thanks @adrianthedev! I know we've looked at the Avo 3 upgrade. I'm not sure what is blocking it at the moment. I'll check again when I have a moment.

PR #2831, opened 2024-06-03
Add grid_view? helper method to support grid view detection

gabrielgiroe1:
# Description

Fixes # (issue)

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/app/helpers/avo/application_helper.rb
+++ b/app/helpers/avo/application_helper.rb
@@ -142,7 +142,7 @@ module Avo
     end
 
     def view_type(params, resource)
-      params[:view_type]&.to_sym || resource.available_view_types.first || resource.default_view_type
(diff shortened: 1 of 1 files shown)


adrianthedev (changes requested):
I'm not sure this is the right approach.
Please search `@view_type` and `available_view_types` and see how they work together.

If we add this change that you're making, we're back at two places where we do the `params[:view_type] == "grid"` check.
Let's bring it all in one place/helper/something.

Check these places too:

- https://github.com/avo-hq/avo/blob/main/app/components/avo/views/resource_index_component.rb#L53
- https://github.com/avo-hq/avo/blob/main/app/controllers/avo/base_controller.rb#L315

PR #2861, opened 2024-06-14
fix: generators can run without prior migrations

adrianthedev:
# Description

Fixes an issue where the generators would fail if there were pending migrations.
This is common if you install rails, run the devise generator and then Avo's.

This PR also brings a new command `all_resources` which will generate all resources.

### Problem

We're now stuck at generating all the resources. The `all_resources` generator will not continue past the first resource it finds.

CleanShot 2024-06-14 at 16 43 02@2x

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/lib/generators/avo/install_generator.rb
+++ b/lib/generators/avo/install_generator.rb
@@ -13,12 +13,15 @@ module Generators
         route "mount Avo::Engine, at: Avo.configuration.root_path"
 
         template "initializer/avo.tt", "config/initializers/avo.rb"
+
         create_resources
       end
 
-      def create_resources
-        if defined?(User).present?
-          Rails::Generators.invoke("avo:resource", ["user", "-q"], {destination_root: Rails.root })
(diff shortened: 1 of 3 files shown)


lib/generators/avo/install_generator.rb
```diff
@@ -13,12 +13,15 @@ def create_initializer_file
...
         create_resources
       end
 
-      def create_resources
-        if defined?(User).present?
-          Rails::Generators.invoke("avo:resource", ["user", "-q"], {destination_root: Rails.root })
+      no_tasks do
+        def create_resources
+          if defined?(User).present?
+            Rails::Generators.invoke("avo:resource", ["user", "-q"], {destination_root: Rails.root })
```
Paul-Bob, line 23:
```suggestion
            Rails::Generators.invoke("avo:resource", ["user", "-q"], {destination_root: Rails.root})
```

PR #2869, opened 2024-06-18
Make database seed work if Unsplash down

krystof-k:
Well, Unsplash is [down](https://source.unsplash.com) so just a little rescue in the seeds script…

### DIFF
--- a/spec/dummy/db/seeds.rb
+++ b/spec/dummy/db/seeds.rb
@@ -169,7 +169,7 @@ puts "Creating posts"
     width = [1000, 1100, 1200, 1300].sample
     height = [1000, 1100, 1200, 1300].sample
     url = "https://source.unsplash.com/random/#{width}x#{height}/?all"
-    io = URI.open(url)
+    io = URI.open(url) # standard:disable Security/Open
   rescue
     puts "Failed to fetch cover photo from Unsplash"
   end
--- a/spec/dummy/db/seeds.rb
+++ b/spec/dummy/db/seeds.rb
@@ -173,10 +173,7 @@ puts "Creating posts"
   rescue
     puts "Failed to fetch cover photo from Unsplash"
   end
-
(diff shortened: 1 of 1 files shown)


spec/dummy/db/seeds.rb
```diff
@@ -165,7 +165,15 @@
...
-  post.cover_photo.attach(io: URI.open("https://source.unsplash.com/random/#{[1000, 1100, 1200, 1300].sample}x#{[1000, 1100, 1200, 1300].sample}/?all"), filename: "cover.jpg")
+  begin
+    width = [1000, 1100, 1200, 1300].sample
+    height = [1000, 1100, 1200, 1300].sample
+    url = "https://source.unsplash.com/random/#{width}x#{height}/?all"
+    io = URI.open(url) # standard:disable Security/Open
+  rescue
+    puts "Failed to fetch cover photo from Unsplash"
+  end
+  post.cover_photo.attach(io: io, filename: "cover.jpg") if io
```
adrianthedev, line 176:
```suggestion
  begin
    width = [1000, 1100, 1200, 1300].sample
    height = [1000, 1100, 1200, 1300].sample
    url = "https://source.unsplash.com/random/#{width}x#{height}/?all"
    io = URI.open(url) # standard:disable Security/Open

    post.cover_photo.attach(io: io, filename: "cover.jpg")
  rescue
    puts "Failed to fetch cover photo from Unsplash"
  end
```

adrianthedev: Could this work as well?

krystof-k: Yeah but in case just the attach fails the error message isn't true :)

adrianthedev: Got it! makes sense!
Ok. let's go with the original version.

PR #2893, opened 2024-06-25
[WIP] Introduce Literal properties

joeldrapper:
# Description

[WIP] This PR introduces Literal properties.

Note: this should follow #2892. I couldn’t stack it on that branch because the branch is on a fork.

# Checklist:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,6 +11,9 @@ gemspec
 # Git. Remember to move these dependencies to your gemspec before releasing
 # your gem to rubygems.org.
 
+# TODO: Move this into the Gemspec
+gem "literal", path: "../literal"
+
 # To use a debugger
 # gem 'byebug', group: [:development, :test]
 
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,6 @@
-source "https://rubygems.org"
+# frozen_string_literal: true
+
+source 'https://rubygems.org'
 git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
 # Declare your gem's dependencies in avo.gemspec.
@@ -11,9 +13,6 @@ gemspec
 # Git. Remember to move these dependencies to your gemspec before releasing
 # your gem to rubygems.org.
 
-# TODO: Move this into the Gemspec
-gem "literal", path: "../literal"
-
 # To use a debugger
 # gem 'byebug', group: [:development, :test]
 
@@ -25,16 +24,16 @@ gem 'cssbundling-rails'
 #
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
 # gem "rails", "~> 7.1.0"
-gem "rails", github: "rails/rails", branch: "main"
+gem 'rails', github: 'rails/rails', branch: 'main'
 
 # Avo file filed requires this gem
 # gem "activestorage"
-gem "activestorage", github: "rails/rails", branch: "main"
+gem 'activestorage', github: 'rails/rails', branch: 'main'
 
 # Use postgresql as the database for Active Record
-gem "pg", ">= 0.18", "< 2.0"
+gem 'pg', '>= 0.18', '< 2.0'
 # Use Puma as the app server
-gem "puma", "~> 6.4"
+gem 'puma', '~> 6.4'
 # Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
 # gem "jbuilder", "~> 2.7"
 # Use Redis adapter to run Action Cable in production
@@ -43,31 +42,31 @@ gem 'redis', '~> 5.0'
 # gem 'bcrypt', '~> 3.1.7'
 
 # Reduces boot times through caching; required in config/boot.rb
-gem "bootsnap", ">= 1.4.2", require: false
+gem 'bootsnap', '>= 1.4.2', require: false
 # Call 'byebug' anywhere in the code to stop execution and get a debugger console
-gem "byebug", platforms: [:mri, :mingw, :x64_mingw]
-gem "dotenv-rails"
+gem 'byebug', platforms: %i[mri mingw x64_mingw]
+gem 'dotenv-rails'
 # Access an interactive console on exception pages or by calling 'console' anywhere in the code.
-gem "web-console", ">= 3.3.0"
-gem "listen", ">= 3.5.1"
+gem 'web-console', '>= 3.3.0'
+gem 'listen', '>= 3.5.1'
 # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
-gem "spring"
+gem 'spring'
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
-gem "tzinfo-data", platforms: [:mingw, :mswin, :x64_mingw, :jruby]
+gem 'tzinfo-data', platforms: %i[mingw mswin x64_mingw jruby]
 
-gem "devise"
+gem 'devise'
 #
 # END Dependencies for dummy_app
 #
 
 group :development do
-  gem "standard"
+  gem 'standard'
 
   # Release helper
-  gem "bump", require: false
-  gem "gem-release", require: false
+  gem 'bump', require: false
+  gem 'gem-release', require: false
 
-  gem "annotate"
+  gem 'annotate'
 
   # gem 'rack-mini-profiler'
   # gem 'memory_profiler'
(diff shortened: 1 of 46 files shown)


Gemfile
```diff
@@ -11,6 +11,9 @@ gemspec
 # Git. Remember to move these dependencies to your gemspec before releasing
 # your gem to rubygems.org.
 
+# TODO: Move this into the Gemspec
+gem "literal", path: "../literal"
```
joeldrapper, line 15: I will cut a release on RubyGems before this PR is ready to merge and move this up to the Gemspec.

app/components/avo/actions_component.rb
```diff
@@ -1,34 +1,37 @@
...
+  prop :exclude, Array, default: -> { [] } do |value|
+    Array(value)
+  end
+  prop :include, Array, default: -> { [] }
+  prop :style, Symbol, default: :outline
+  prop :color, Symbol, default: :primary
+  prop :label, _Nilable(String), reader: :public do |value|
+    value || I18n.t("avo.actions")
+  end
+  prop :size, Symbol, default: :md, reader: :public
```
joeldrapper, line 18: I’m not sure what the union of sizes is for this component. I couldn't see that this option is even being used at the moment.

app/components/avo/actions_component.rb
```diff
@@ -1,34 +1,37 @@
...
+  prop :view, _Nilable(String)
+  prop :exclude, Array, default: -> { [] } do |value|
+    Array(value)
+  end
+  prop :include, Array, default: -> { [] }
+  prop :style, Symbol, default: :outline
+  prop :color, Symbol, default: :primary
+  prop :label, _Nilable(String), reader: :public do |value|
+    value || I18n.t("avo.actions")
+  end
```
joeldrapper, line 17: If `I18n.t("avo.actions")` will never be nil, we can actually make the type of this `String` instead of `_Nilable(String)`.

app/components/avo/sidebar_profile_component.rb
```diff
@@ -1,11 +1,7 @@
 # frozen_string_literal: true
 
-class Avo::SidebarProfileComponent < ViewComponent::Base
-  attr_reader :user
-
-  def initialize(user:)
-    @user = user
-  end
+class Avo::SidebarProfileComponent < Avo::BaseComponent
+  prop :user, _Nilable(_Any), reader: :public
```
joeldrapper, line 4: I couldn’t figure out what to make this type. It seems to be called with `nil` in some of the tests. Otherwise it’s a User record, but it doesn't seem to be specifically an "Avo-User". So maybe it should be an `ActiveRecord::Base`? Can we be more specific? `Avo::BaseResource` or something, maybe?

app/components/avo/divider_component.rb
```diff
@@ -1,9 +1,6 @@
...
 
-class Avo::DividerComponent < ViewComponent::Base
-  attr_reader :label
-
-  def initialize(label = nil, **args)
-    @label = label
-  end
+class Avo::DividerComponent < Avo::BaseComponent
+  prop :label, _Nilable(String), :positional, reader: :public
+  prop :args, Hash, :**
```
joeldrapper, line 5: I don’t think this is needed, but it was allowed before so didn’t want to break anything.

app/components/avo/actions_component.rb
```diff
@@ -1,34 +1,37 @@
...
-  def actions
+  def filter_actions
     if @exclude.present?
-      @actions.reject { |action| action.class.in?(@exclude) }
+      @actions.reject! { |action| action.class.in?(@exclude) }
     elsif @include.present?
-      @actions.select { |action| action.class.in?(@include) }
-    else
-      @actions
+      @actions.select! { |action| action.class.in?(@include) }
```
joeldrapper, line 34: This seemed to be doing quite a bit of work each time it was called so I thought it might make sense to do it once and store it back to the `@actions` ivar.

app/components/avo/field_wrapper_component.rb
```diff
@@ -1,50 +1,29 @@
...
+  prop :help, _Nilable(String) # do we really need it?
+  prop :field, Avo::Fields::BaseField, reader: :public
+  prop :form, _Nilable(ActionView::Helpers::FormBuilder), reader: :public
+  prop :full_width, _Boolean, default: false, reader: :public
+  prop :label, _Nilable(String) # do we really need it?
+  prop :resource, _Nilable(Avo::BaseResource), reader: :public
+  prop :short, _Boolean, default: false
+  prop :stacked, _Boolean, default: false do |value|
+    !!value
+  end
```
joeldrapper, line 18: This seemed to be used as a boolean, but was `true | nil` instead of `true | false`. For now, I’ve made it coerce input into a boolean, but it might be better to fix the callsites to send `true | false` instead of `true | nil`.

app/components/avo/field_wrapper_component.rb
```diff
@@ -99,14 +78,7 @@ def data
...
   def stacked?
-    # Override on the declaration level
-    return @stacked unless @stacked.nil?
-
-    # Fetch it from the field
-    return field.stacked unless field.stacked.nil?
-
-    # Fallback to defaults
-    Avo.configuration.field_wrapper_layout == :stacked
+    @stacked || field.stacked || Avo.configuration.field_wrapper_layout == :stacked
```
joeldrapper, line 81: `nil` and `false` are both falsey, so I think we can just make this a simple OR fallback.

app/components/avo/field_wrapper_component.rb
```diff
@@ -1,50 +1,29 @@
...
+  prop :args, Hash, :**
+
+  def after_initialize
     @action = field.action
-    @short = short
-    @stacked = stacked
-    @style = style
-    @view = view
-    @label_for = label_for
+    @classes = @args[:class].present? ? @args[:class] : ""
```
joeldrapper, line 26: Is there a specific reason why use `@classes` instead of `@class` here?

app/components/avo/field_wrapper_component.rb
```diff
@@ -1,50 +1,29 @@
...
+  prop :stacked, _Boolean, default: false do |value|
+    !!value
+  end
+  prop :style, String, default: ""
+  prop :view, String, default: "show", reader: :public
+  prop :label_for, _Nilable(String)
+  prop :args, Hash, :**
+
+  def after_initialize
     @action = field.action
```
joeldrapper, line 25: Unless `field.action` is slow, there's probably no need to assign an instance variable here. This might be better as a method or delegation, e.g.

```ruby
delegate :action, to: :@field
```

app/components/avo/field_wrapper_component.rb
```diff
@@ -1,50 +1,29 @@
...
-    @resource = resource
+  prop :dash_if_blank, _Boolean, default: true, reader: :public
+  prop :data, Hash, default: -> { {} }
+  prop :compact, _Boolean, default: false, reader: :public
+  prop :help, _Nilable(String) # do we really need it?
+  prop :field, Avo::Fields::BaseField, reader: :public
+  prop :form, _Nilable(ActionView::Helpers::FormBuilder), reader: :public
+  prop :full_width, _Boolean, default: false, reader: :public
+  prop :label, _Nilable(String) # do we really need it?
+  prop :resource, _Nilable(Avo::BaseResource), reader: :public
```
joeldrapper, line 14: I’m not convinced all these should be nilable, but I needed to make these ones nilable to get the tests to pass.

joeldrapper, line 14: There’s a lot of `reader: :public` here, but we can probably just use instance variable references for most of these. I doubt many are actually being accessed externally.

app/components/avo/index/grid_item_component.rb
```diff
@@ -4,16 +4,15 @@ class Avo::Index::GridItemComponent < Avo::BaseComponent
...
-    @parent_record = parent_record
-    @parent_resource = parent_resource
-    @actions = actions
-    @card = Avo::ExecutionContext.new(target: resource.grid_view[:card], resource: resource, record: resource.record).handle
-    @whole_html = Avo::ExecutionContext.new(target: resource.grid_view[:html], resource: resource, record: resource.record).handle
+  prop :resource, Avo::BaseResource
+  prop :reflection, _Never
+  prop :parent_record, _Void
+  prop :parent_resource, _Void, reader: :public
+  prop :actions, _Void, reader: :public
```
joeldrapper, line 11: Need to figure out what types these are. `_Void` here is always true, allows anything through.

app/components/avo/items/panel_component.rb
```diff
@@ -3,19 +3,17 @@
...
+  prop :item, _Any
+  prop :is_main_panel, _Boolean
+  prop :resource, Avo::BaseResource
+  prop :view, Symbol, &:to_sym
+  prop :actions, _Nilable(_Array(Avo::BaseAction))
+  prop :index, _Nilable(Integer)
+  prop :parent_component, _Void
+  prop :parent_record, _Nilable(ActiveRecord::Base)
+  prop :parent_resource, _Nilable(Avo::BaseResource)
+  prop :reflection, _Void
```
joeldrapper, line 16: What type is a reflection?

app/components/avo/modal_component.rb
```diff
@@ -1,22 +1,19 @@
...
+  renders_one :controls
 
   def width_classes
-    case width.to_sym
-    when :md
+    case @width
+    in :md
       "w-11/12 lg:w-1/2 sm:max-w-168"
-    when :xl
+    in :xl
```
joeldrapper, line 16: I switched this to case-in, since it’s more exhaustive. It has a default else-raise.

app/components/avo/index/grid_item_component.rb
```diff
@@ -4,16 +4,15 @@ class Avo::Index::GridItemComponent < Avo::BaseComponent
...
-  def initialize(resource: nil, reflection: nil, parent_record: nil, parent_resource: nil, actions: nil)
-    @resource = resource
-    @reflection = reflection
-    @parent_record = parent_record
-    @parent_resource = parent_resource
-    @actions = actions
-    @card = Avo::ExecutionContext.new(target: resource.grid_view[:card], resource: resource, record: resource.record).handle
-    @whole_html = Avo::ExecutionContext.new(target: resource.grid_view[:html], resource: resource, record: resource.record).handle
+  prop :resource, Avo::BaseResource
+  prop :reflection, _Never
```
joeldrapper, line 8: Set this to `_Never` for now to try to find the type of object this is called with, but none of the tests ever pass a reflection so couldn’t find it. I’ll keep looking.

app/components/avo/index/grid_item_component.rb
joeldrapper: I’m pretty sure this one is `_Nilable(_Array(Avo::BaseAction))`

adrianthedev:
Hey @joeldrapper. Is this "ready" to review? Or should we have another look over it?
I don't want to rush you, but it touches on many files and we could end up with conflicts pretty quick.

adrianthedev:
I forgot that today is Brighton Ruby day...

joeldrapper:
@adrianthedev I think this PR is too big. I’m going to open some smaller ones instead.

PR #2896, opened 2024-06-26
fix: go back button after changing tabs

Paul-Bob:
# Description

Fixes an issue where after changing tab and clicking create / show / edit inside that tab the browser back button would not work anymore.

Found the solution on this thread

- https://github.com/hotwired/turbo/issues/792

## First approach solution (obsolete) 

Not sure whats the exact cause of the issue but I found this 2 related threads where @lorint points to this solution.

- https://stackoverflow.com/questions/73991459/browser-back-button-is-not-working-for-some-feature-in-stimulus-js
- https://github.com/hotwired/stimulus/issues/589

# Checklist:


Paul-Bob:
Let's not merge this one yet

PR #2919, opened 2024-06-30
Use literal properties for `ActionsComponent`

joeldrapper:
I’ve update the ActionsComponent to use literal properties.

- Switched from using accessor methods to using private instance variable references
- Added unions for button styles and sizes
- Now filtering actions ahead-of-time so the filtering doesn’t happen multiple times


app/components/avo/actions_component.rb
```diff
@@ -2,40 +2,43 @@
...
-    @exclude = Array(exclude)
-    @include = include
-    @color = color
-    @style = style
-    @label = label || I18n.t("avo.actions")
-    @size = size
-    @icon = icon
-    @as_row_control = as_row_control
+
+  prop :actions, _Array(Avo::BaseAction), default: -> { [] }
```
joeldrapper, line 8: Is this really an optional field? Seems like maybe it should be required, but it was optional before so I maintained that.

app/components/avo/actions_component.rb
```diff
@@ -2,40 +2,43 @@
...
+  prop :resource, _Nilable(Avo::BaseResource)
+  prop :view, _Nilable(Symbol), &:to_sym
```
joeldrapper, line 10: Should these really be nilable or are they required?

app/components/avo/actions_component.rb
```diff
@@ -2,40 +2,47 @@
...
-    @icon = icon
-    @as_row_control = as_row_control
+
+  ACTION_FILTER = _Set(_Class(Avo::BaseAction))
+
+  prop :actions, _Array(Avo::BaseAction), default: -> { [] }
+  prop :resource, _Nilable(Avo::BaseResource)
+  prop :view, _Nilable(Symbol), &:to_sym
+  prop :exclude, _Nilable(ACTION_FILTER), &:to_set
+  prop :include, _Nilable(ACTION_FILTER), &:to_set
```
joeldrapper, line 12: Couple of points on these: I updated them to be Sets since we do repeated _inclusion_ lookups. Arrays will still be accepted because of the `&:to_set` coercion.

One thing I haven’t carried over, which you may need me to bring back is the `Array()` coercion that used to be on the exclude list but not on the include list. This would have allowed you to pass a single non-enumerable object and have it wrapped in an Array. If we bring that back, we should probably add it to the include list as well.

app/components/avo/actions_component.rb
```diff
@@ -2,40 +2,47 @@
...
-    @exclude = Array(exclude)
-    @include = include
-    @color = color
-    @style = style
-    @label = label || I18n.t("avo.actions")
-    @size = size
-    @icon = icon
-    @as_row_control = as_row_control
+
+  ACTION_FILTER = _Set(_Class(Avo::BaseAction))
```
joeldrapper, line 6: I don’t know if this type is too precise for your taste. It says it must be a Set of classes that inherit from `Avo::BaseAction`. I find the documentation value of spelling this out to be helpful, but we could also just use `Set` as the type.

app/components/avo/actions_component.rb
```diff
@@ -2,40 +2,47 @@
...
+  prop :actions, _Array(Avo::BaseAction), default: -> { [] }
+  prop :resource, _Nilable(Avo::BaseResource)
+  prop :view, _Nilable(Symbol), &:to_sym
+  prop :exclude, _Nilable(ACTION_FILTER), &:to_set
+  prop :include, _Nilable(ACTION_FILTER), &:to_set
+  prop :style, Avo::ButtonComponent::STYLE, default: :outline
+  prop :color, Symbol, default: :primary
+  prop :label, String do |value|
+    value || I18n.t("avo.actions")
+  end
```
joeldrapper, line 17: As far as I can tell, this string should be non-nilable because we’ll always get a string back from the I18n lookup.

app/components/avo/actions_component.rb
```diff
@@ -2,40 +2,47 @@
...
+    @actions = @actions.dup
+
+    if @exclude.any?
+      @actions.reject! { |action| @exclude.include?(action.class) }
+    end
+
+    if @include.any?
+      @actions.select! { |action| @include.include?(action.class) }
     end
   end
```
joeldrapper, line 40: I’ve changed action filtering to happen ahead of time. I’ve also updated the conditional so that you can filter with an include and exclude list. Previously, if there was an exclude filter, the include filters would be ignored.

The dup on line 31 is so we can use `reject!` and `select!` without concerns for mutating an array that might be used for something else after being passed to this component.

The filters uses `Set#include?` now, which will be really fast in the case that there are many actions to filter.

I’m not sure if any tests cover this functionality so may be one to look out for.

app/components/avo/button_component.rb
```diff
@@ -1,9 +1,9 @@
 # frozen_string_literal: true
 
-# A button/link can have the following settings:
-# style: primary/outline/text
-# size: :xs :sm, :md, :lg
 class Avo::ButtonComponent < Avo::BaseComponent
+  SIZE = _Union(:xs, :sm, :md, :lg, :xl)
+  STYLE = _Union(:primary, :outline, :text, :icon)
```
joeldrapper, line 5: This seemed like the right place to define these unions, even though I’m not updating the `ButtonComponent` just yet.

I tried to find a limited set of colors, but it looks like these are just interpolated directly into the class list. That brings up an interesting concern: how does the Tailwind JIT know which colors to build? Have we listed them somewhere else so the JIT knows?

joeldrapper:
It doesn't seem related, but I’m getting consistent failures of the following system tests:

```
rspec ./spec/system/avo/tags_spec.rb:135 # Tags ajax request fetches the users
rspec ./spec/system/avo/test_helpers_spec.rb:192 # TestHelpers isolated tests run_action runs the panel action in show
```

They pass in `main` so I must have messed something up. I’ll look into it.

app/components/avo/actions_component.rb
```diff
@@ -2,40 +2,47 @@
 
 class Avo::ActionsComponent < Avo::BaseComponent
   include Avo::ApplicationHelper
-  attr_reader :label, :size, :as_row_control
```
joeldrapper, line 5: As far as I can tell, we don’t actually need these public accessors.

app/components/avo/actions_component.rb
adrianthedev: This is a good enforcement! We have that assumption that we expect BaseAction descendents.

app/components/avo/actions_component.rb
adrianthedev: I think it's best to keep it optional. We might have weird filters applied to the list and we don't want that to crash the app.

adrianthedev:
Is this good to review @joeldrapper?

joeldrapper:
Yep, good for an initial review. I still need to track down why these two tests are failing.

Paul-Bob:
> Yep, good for an initial review. I still need to track down why these two tests are failing.

I applied most of the changes from this PR [here](https://github.com/avo-hq/avo/pull/3080)

Had same 2 tests failing, fixed by checking the `on_record_page?` against symbols instead strings ([check change here](https://github.com/avo-hq/avo/pull/3080/files#diff-1c4a426d3fb41e4419fbf6bf9d1b5a3c520da819b300d886d1d9b6a7393642f1R60)).

Main is merged and conflicts are solved

app/components/avo/actions_component.rb
Paul-Bob: This is great! Thanks!

Is it possible to access other props in one prop initializer?

Something like:
```ruby
  prop :actions, _Array(Avo::BaseAction), default: [].freeze do |actions|
    # Apply all the include / exclude logic here
    if @exclude.any?
      actions.reject! { |action| @exclude.include?(action.class) }
    end

    if @include.any?
      actions.select! { |action| @include.include?(action.class) }
    end

    actions
  end
```

Paul-Bob (approved):
Thanks for the contribution @joeldrapper! It's looking good! I've applied most of the changes from this PR [here](https://github.com/avo-hq/avo/pull/3080) without knowing about the conflicts with this one. Already fixed the conflicts.

Paul-Bob (commented):
Linting

app/components/avo/actions_component.rb
```diff
@@ -2,35 +2,39 @@
...
+  prop :style, Avo::ButtonComponent::STYLE, default: :outline
   prop :actions, _Array(Avo::BaseAction), default: [].freeze
-  prop :exclude, Array, default: [].freeze do |exclude|
-    Array(exclude)
-  end
-  prop :resource, _Nilable(Avo::Resources::Base)
+  prop :exclude, _Nilable(ACTION_FILTER), &:to_set
+  prop :resource, _Nilable(Avo::BaseResource)
   prop :view, _Nilable(Symbol), &:to_sym
+  def after_initialize
```
Paul-Bob, line 21:
```suggestion

  def after_initialize
```

app/components/avo/actions_component.rb
```diff
@@ -2,35 +2,40 @@
 
 class Avo::ActionsComponent < Avo::BaseComponent
   include Avo::ApplicationHelper
+  ACTION_FILTER = _Set(_Class(Avo::BaseAction))
```
Paul-Bob, line 5:
```suggestion

  ACTION_FILTER = _Set(_Class(Avo::BaseAction))
```

app/components/avo/actions_component.rb
Paul-Bob: Both should allow a single non-enumerable object or an array of objects

app/components/avo/actions_component.rb
Paul-Bob: yes, can't we use the default option with the i18n string?

app/components/avo/button_component.rb
Paul-Bob: > how does the Tailwind JIT know which colors to build? Have we listed them somewhere else so the JIT knows?

There are some standard ones, but it's also possible to do a [tailwind integration](https://docs.avohq.io/3.0/tailwindcss-integration.html) and use custom colors. Let's not define any restriction on colors

PR #2927, opened 2024-07-01
refactor: Prefer Model.reflections over private Model#_reflections.

simi:
:construction: This targets 2.x branch.

- `_reflections` returns different key type per Rails version

:information_source: Model.reflections seems supported for 6.0+ (I have not checked earlier versions since those seems unsupported).

:bulb: This fixes issues at RubyGems.org with Rails 7.2 beta compatibility.

### DIFF
--- a/lib/avo/base_resource.rb
+++ b/lib/avo/base_resource.rb
@@ -114,7 +114,7 @@ module Avo
       end
 
       def get_record_associations(record)
-        record._reflections
+        record.class.reflections
       end
 
       def valid_association_name(record, association_name)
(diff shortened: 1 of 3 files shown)


adrianthedev:
@Paul-Bob please review this and it works, let's ship it!

simi:
I can update also to `reflect_on_association`, which seems better IMHO. Same should be applied to 3.x.

adrianthedev:
OK. let's retry that one then. Maybe it works on your end!

Paul-Bob:
> I can update also to `reflect_on_association`, which seems better IMHO. Same should be applied to 3.x.

Made [this PR for version 3](https://github.com/avo-hq/avo/pull/3069) that replace `_reflections` usage with `reflect_on_association`

simi:
@Paul-Bob updated, I tried to keep changes minimal. Specs seems happy.

adrianthedev (approved):
Thanks @simi!

Let's ship this.

PR #2934, opened 2024-07-03
feature: apply target resource policy on belongs_to field create new link

### LINKED ISSUE #2723: Apply policy on the `belongs_to` field create new link
### Feature

`field :user, as: :belongs_to` should render the "Create new user" link only if the association policy allows it.

image

### Current workarounds

Custom logic inside `can_create` block

Paul-Bob:
# Description

Fixes #2723

Apply the target resource policy on the belongs_to field create new link.

# Checklist:

  - https://github.com/avo-hq/docs.avohq.io/pull/232

### DIFF
--- a/app/components/avo/fields/belongs_to_field/edit_component.html.erb
+++ b/app/components/avo/fields/belongs_to_field/edit_component.html.erb
@@ -66,8 +66,9 @@
                 <%= @form.hidden_field @field.id_input_foreign_key %>
               <% end %>
             <% end %>
-            <% if field.can_create? %>
-              <% create_href = create_path(Avo.resource_manager.get_resource_by_model_class(type.to_s)) %>
+            <% target_resource_for_type = Avo.resource_manager.get_resource_by_model_class(type.to_s) %>
+            <% if field.can_create?(target_resource_for_type) %>
+              <% create_href = create_path(target_resource_for_type) %>
               <% if !disabled && create_href.present? %>
                 <%= link_to t("avo.create_new_item", item: type.model_name.human.downcase),
                       create_href,
--- a/lib/avo/fields/belongs_to_field.rb
+++ b/lib/avo/fields/belongs_to_field.rb
@@ -281,8 +281,12 @@ module Avo
         end
       end
 
+      # field :user, as: :belongs_to, can_create: true
+      # Only can create when:
+      #   - `can_create: true` option is present
+      #   - target resource's policy allow creation (UserPolicy in this example)
       def can_create?
-        @can_create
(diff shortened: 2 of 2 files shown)


adrianthedev (approved):
Looks good!

app/components/avo/fields/belongs_to_field/edit_component.html.erb
```diff
@@ -66,8 +66,9 @@
                 <%= @form.hidden_field @field.id_input_foreign_key %>
               <% end %>
             <% end %>
-            <% if field.can_create? %>
-              <% create_href = create_path(Avo.resource_manager.get_resource_by_model_class(type.to_s)) %>
+            <% target_resource_for_type = Avo.resource_manager.get_resource_by_model_class(type.to_s) %>
+            <% if field.can_create?(target_resource_for_type) %>
```
davegson, line 70: @Paul-Bob a conversation starter:

I have a case where I would love to show this link to the new form, but it is hidden since it fails the create? check.

It breaks down to a simple example like:

```Ruby
class CommentPolicy
  alias_method :comment, :record

  def new
    Current.user.present?
  end

  def create?
    Current.user.present? && comment.user_id == Current.user.id
  end
end
```

In this example I want users to be able to create new comments, but at the same time I want to assure they can only create comments themselves, not for other users - hence the `create?` method has extra logic (which hides the link)

Wouldn't it make sense to change the logic of whether to show the new_item_link or not in accordance to the policy's #new? method instead of the #create? method? (which is currently defined in the `BelongsToField#can_create?`

I am unsure if this is opening a can of words where you would have to separate logic in more places than here where in Avo new basically equals create - or if it would be possible to do this. I can work around this by overwriting this `edit_component.html.erb` - but thought it might be interesting to think about.

Also not shying away from suggesting a PR - just want to check out whether this would make sense or would quickly explode in scope

Paul-Bob: To remain consistent with the policy methods, I think your suggestion makes perfect sense: new? is checked to display the link, and create? is verified in the controller upon creation.

davegson: yeah cool - happy to hear you agree! Do you want me to create a PR or is this something you'll adapt in your workflow?

Paul-Bob: If you could create a PR, I would appreciate it. The authorization part is in the `avo-pro` repo, but you can make the relevant changes in the `avo` repo. I'll take it from there, tweak it and add tests. Thank you!

PR #2942, opened 2024-07-04
chore: drop httparty dependency

### LINKED ISSUE #2899: Remove Httparty in favor of native ruby solution
We only use Httparty on the `HQ` class and we want to slim down on dependencies we don't necessarily need.

Let's see if Chat GPT can help with this transition.

Paul-Bob:
# Description

Fixes #2899 

# Checklist:

### DIFF
--- a/lib/avo/licensing/request.rb
+++ b/lib/avo/licensing/request.rb
@@ -1,0 +1,18 @@
+module Avo
+  module Licensing
+    class Request
+      class << self
+        def post(endpoint, body:, timeout:)
+          uri = URI.parse(endpoint)
+          http = Net::HTTP.new(uri.host, uri.port)
+          http.use_ssl = (uri.scheme == "https")
+          http.read_timeout = timeout
+          http.open_timeout = timeout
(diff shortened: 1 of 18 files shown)


adrianthedev (approved):
Looks good!
Have you by any chance tested some of those exceptions by hand to see if they will be caught properly?

Paul-Bob:
Tested `SocketError`, `Errno::ECONNREFUSED`, and `Net::ReadTimeout`

Paul-Bob:
TODO: test `OpenSSL::SSL::SSLError`

lib/avo/licensing/request.rb
```diff
@@ -0,0 +1,18 @@
+module Avo
```
jpawlyn, line 1: I am seeing an `uninitialized constant Net::HTTP` when I start a rails server with rails 3.3.4 and avo 3.10.6. Do we need to add the following ?
 ```ruby
require "net/http"
```

When I add this to the top of `avo.rb` initialiser the problem is resolved.

adrianthedev: Yeah, I've been seeing that as well. We'll add it to the initial file as well.

PR #2956, opened 2024-07-08
feature: add attachments to record

adrianthedev:
# Description

This PR adds the `attachments` option which will add a new way to eager load attachments on a list of records.

This also adds `single_includes` and `single_attachments` that are being applied to those times a single record is being teched from the DB.

Another way to do this would be to optionally turn `includes` and `attachments` into callables and have an `if/else` statement to check if the view is `index` or `show/forms`.
By default they would be applied only to `index`, but in the block can be added to `show/forms` as well.

Implements changes from https://github.com/minthesize/avo/pull/3/

# Checklist:

## Manual review steps

1. Go to the products index page
1. observe that comments have been eager loaded
2. go to the product show page
3. observe that the reviews have been eager loaded

### DIFF
--- a/app/controllers/avo/base_controller.rb
+++ b/app/controllers/avo/base_controller.rb
@@ -34,6 +34,13 @@ module Avo
         @query = @query.includes(*@resource.includes)
       end
 
+      # Eager load attachments
+      if @resource.attachments.present?
+        @resource.attachments.each do |attachment|
+          @query = @query.send(:"with_attached_#{attachment}")
+        end
+      end
+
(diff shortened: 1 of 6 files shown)


Paul-Bob (approved):
IMO feels more natural to do something like:

```ruby
self.includes =  -> {
  if view.index?
    [:user, :comments]
  else
    [:user, :reviews]
  end
}

self.attachments = [:cover_photo, :audio, :attachments]
```

Where the blocks are executed on all views instead of having different DSLs for different views.

Issue #2843, opened 2024-06-06
Confirmation message before save


filipegiusti:
### Feature
When opening the edit page or when saving a resource, users should see a confirmation modal.

### Current workarounds
I guess it could be achieve by manually replicating the default actions list and overriding the save button.

### Screenshots or screen recordings
N/A.

### Additional context
We would like the ability to ask for confirmation on saves depending on resource state because some of our resources have states that are low stakes (unpublished) and ones that are high stakes (published) and we would like to avoid human error on high stakes states.

Initially discussed [this discord thread](https://discord.com/channels/740892036978442260/740892037640880180/1248008946363531285).

G-Simpson:
@filipegiusti Could you please post the context from the discord thread? I don't have access. Thanks

adrianthedev:
Hey @G-Simpson. In case you want to join, use this link [https://avo.cool/chat](https://avo.cool/chat)

adrianthedev:
This would be a cool feature to have!

### Approach

- add the `confirm_before_create|update|save` option to `BaseResource`
- create a Stimulus controller which gets attached in case the option is on
- the stimulus controller will first show a modal
- let's re-use the `custom-confirm.js` behavior to trigger the same modal. Some work will be required here to extract that functionality
- the modal should trigger a proper submit on the form

BTW, you can check out that thread [here](https://www.linen.dev/s/Avo/t/18883100/ux-improvements#ce095635-7eda-40ed-8340-734897765366) on the public chat archives.

G-Simpson:
I'm happy to have a look at this

PR #2965, opened 2024-07-11
Feature: Add optional confirmation on save

G-Simpson:
I have added a conditional turbo-confirm message to the save operation on the edit page. The spinner activated by the stimulus controller needed to be updated to work without a page refresh. I have added appropriate system tests. Please let me know what you think!

Fixes https://github.com/avo-hq/avo/issues/2843

# Checklist:

## Screenshots & recording
confirm_on_save

## Manual review steps

1. Step 1
1. Step 2


binarygit:
Hey @G-Simpson Thanks for working on this!
I took a look at your changes then went source diving into the code. There might be a less verbose way of doing this, and we also need to expand this to work with (and be configurable) create|update|save actions.

There's already an if clause in `loading_button` which checks if the user needs to confirm the action or not. We should be able to pass in the appropriate value (confirmMessageValue) and have this functionality working for us.
```js
  attemptSubmit(e) {
    // If the user has to confirm the action
    if (this.confirmationMessageValue) {
      this.confirmAndApply(e)
    } else {
      this.applyLoader()
    }

    return null
  }
```
To use this, there's already code in `/avo/button_component.rb`  which is looking at the args that are getting passed and is creating the button:
```ruby
  def args
    if @args[:loading]
      @args[:"data-controller"] = "loading-button"
      @args[:"data-loading-button-confirmed-value"] = false
      @args[:"data-action"] = "click->loading-button#attemptSubmit"

      if @args[:confirm]
        @args[:"data-loading-button-confirmation-message-value"] = @args.delete(:confirm)
      end
    end

    @args[:class] = button_classes
    @args[:aria] = @aria

    @args
  end
```
Could you look at how we can pass the `args[:confirm]` arg into our SaveButton component?

G-Simpson:
@binarygit thanks for getting back to me with a detailed response.

I've done a bit of digging and it seems that the `this.confirmAndApply(e)` is redundant. When I specify a turbo_confirm [here](https://github.com/avo-hq/avo/blob/main/app/components/avo/resource_component.rb#L185), in a way that's similar to the delete button, we don't trigger the condition you mentioned above [here](https://github.com/avo-hq/avo/blob/main/app/components/avo/button_component.rb#L33).  If I specifically add a `confirm` option, then we do trigger it, although it is the standard js confirm, not the one that has been [hijacked by turbo](https://github.com/avo-hq/avo/blob/main/app/javascript/js/custom-confirm.js) which I assume is the one we want going forward.

An example:
js-confirm

I would like to remove that redundant code but I'm not sure how the maintainers of this project feel about whether that should be part of this PR or another. I tilt towards another but am happy either way.

I agree that where I defined the original turbo_confirm on the form wasn't the best place for it, so I've moved it to match the pattern of the other button.

app/javascript/js/controllers/loading_button_controller.js
```diff
@@ -11,6 +11,16 @@ export default class extends Controller {
...
+  connect() {
+    this.button.setAttribute('data-original-content', this.button.innerHTML)
+    this.dialog = document.getElementById('turbo-confirm')
+    this.dialogCloseHandler = this.handleDialogClose.bind(this)
+
+    this.dialog.addEventListener('close', this.dialogCloseHandler)
+  }
+  disconnect() {
+    this.dialog.removeEventListener('close', this.dialogCloseHandler)
+  }
```
Paul-Bob, line 23:
```suggestion

  connect() {
    this.button.setAttribute('data-original-content', this.button.innerHTML)
    this.dialog = document.getElementById('turbo-confirm')
    this.dialogCloseHandler = this.handleDialogClose.bind(this)

    this.dialog.addEventListener('close', this.dialogCloseHandler)
  }

  disconnect() {
    this.dialog.removeEventListener('close', this.dialogCloseHandler)
  }
```

Paul-Bob (commented):
Hei @G-Simpson thanks for digging into this.

I let some lint suggestions

I also checked the code regrading the `confirm` and `attemptSubmit` discussion with @binarygit . It seems that `attemptSubmit` is indeed redundant and not used anymore

We can follow 2 approaches

1 - remove the redundant code from the button component and from the JS controller 
2 - keep that code and make it work with the custom dialog

For both approaches we need to ensure that the correct behavior will be applied for all buttons that have a combination of `loading: true` and a `confirm` message

G-Simpson:
@Paul-Bob I'll remove it this PR, thanks

G-Simpson:
The only button with `loading: true` was that button, I've removed the redundant code. all forms of button confirmation now work the same way (`turbo_confirm`)

Paul-Bob:
> The only button with `loading: true` was that button, I've removed the redundant code. all forms of button confirmation now work the same way (`turbo_confirm`)

I've also checked our private repos for `loading: true`; it was indeed used only on the save button.

Paul-Bob:
@G-Simpson I just merged a PR that changes all the code from `Avo::BaseResource` to `Avo::Resources::Base` Can you please solve the conflicts?

Paul-Bob (approved):
Thank you for this contribution @G-Simpson! It's looking good and working as expected.

PR #2966, opened 2024-07-11
Run i18n_specs in a separate job

### LINKED ISSUE #2957: Extract the i18n tests from system to their own category
### Context

The System tests take quite a while to run and theyr feedback is very valuable.
It's not cool to have them fail just because the i18n tasks failed.

Let's extract them to a separate category and have separate CI tasks for those (without ruby/rails matrix). I would assume they're fast!

binarygit:
# Description

Creates a separate job to run i18n specs.

Fixes #2957 

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/.github/workflows/i18n-tests.yml
+++ b/.github/workflows/i18n-tests.yml
@@ -1,4 +1,4 @@
-name: System Tests
+name: i18n Tests
 
 on:
   pull_request:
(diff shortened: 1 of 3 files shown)


Paul-Bob (changes requested):
Right now any error in the i18n test will cancel the system tests

Let's do a separate workflow file for the i18n.

binarygit:
Hey @Paul-Bob, error in either the system tests or the i18n one wouldn't have affected either jobs.
image

But I extracted out the test to a separate workflow because there was no advantage in keeping it as a job compared to a workflow. I also think it's easier to read the file now that it's a workflow. Thanks for your suggestion! :smile:

Paul-Bob (approved):

PR #2968, opened 2024-07-11
Render action controls without adding to actions method

### LINKED ISSUE #2917: Decouple actions in the custom controls from the actions in the `def actions` method
### Context

Right now there's a limitation of not being able to use an action in the custom controls area (show, index, edit, row) unless it's added first in the `def actions` method of the resource.

This can lead to [issues](https://github.com/avo-hq/avo/issues/2908) with displaying the actions in different contexts and doing weird gymnastics around it to support advanced use-cases.

### Proposal

Let's decouple those two and fix those issues once and for all.

binarygit:
# Description

Fixes #2917 

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/app/components/avo/resource_component.rb
+++ b/app/components/avo/resource_component.rb
@@ -50,8 +50,6 @@ class Avo::ResourceComponent < Avo::BaseComponent
   end
 
   def can_see_the_actions_button?
-    return false if @actions.blank?
-
     return authorize_association_for(:act_on) if @reflection.present?
 
     @resource.authorization.authorize_action(:act_on, raise_exception: false) && !has_reflection_and_is_read_only


binarygit:
@adrianthedev The tests pass for me locally. But I can't re trigger the job!

adrianthedev:
Let's add some tests to this as well

Paul-Bob (approved):
This is one of those issues that requires a whole adventure to find the fix. :D

I tested it locally, and it works as expected.

Thanks for digging into this @binarygit

adrianthedev (changes requested):
Great job!
Let's add some tests to future-proof this feature.

Issue #2971, opened 2024-07-11
Add the `visible` option to Panels


icaroryan:
It'd be nice if we had the `visible` option available for Panels so that we can control the visibility of all contained fields at once, without having to set them individually.

By looking at `lib/avo/concerns/visible_items.rb`, I imagine this was done by design. However, since `has_one` associations, for example, have dedicated panels and we can control whether to show all fields with `visible`, it would be great to have that much control with customized panels too.

#### Example
```rb
panel 'Foo', visible: -> { context[:user].is_admin? } do
  field 'Bar', as: :text
  field 'Baz', as: :date_time
end
```

### Current workarounds
```rb
panel 'Foo' do
  field 'Bar', as: :text, visible: -> { context[:user].is_admin? }
  field 'Baz', as: :date_time, visible: -> { context[:user].is_admin? }
  field 'Qux', as: :text, visible: -> { context[:user].is_admin? }
  field 'Fred', as: :text, visible: -> { context[:user].is_admin? }
end
```

icaroryan:
I'm not sure if this is right, but would it be just a matter of adding `include Avo::Concerns::IsVisible` to `panel.rb`? I'm curious how I can contribute to the project.

I'm a bit unsure how to change and test AVO code on my local. But once I get that setup, I'd be happy to take on some issues

Paul-Bob:
Hello @icaroryan thanks for the initiative.

There is this [contribution guide](https://github.com/avo-hq/avo/blob/main/CONTRIBUTING.MD) that will help you with the setup. 

Let us know if you find any obstacles

icaroryan:
Thanks @Paul-Bob!

icaroryan:
Hi @Paul-Bob, I managed to set up the project locally and I successfully implemented the feature 🚀
I do have some questions about testing, but I'll save them for when I open the PR, which I should be able to do as soon as I get the rights to do so. Can you help me with that?

Paul-Bob:
Great news!

> which I should be able to do as soon as I get the rights to do so. Can you help me with that?

You should be able to create a PR from a forced branch

binarygit:
Hey @icaroryan You don't need any special rights to open a PR.

If you've forked this repo you'll be able to send in a PR. You can follow the steps outlined [here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request) and feel free to comment here if you get stuck :smile:

[...]

PR #2989, opened 2024-07-15
feature: add visible option to panels

icaroryan:
# Description
This PR adds a `visible` option to Panels, which should hide all the contained fields as well as the panel.

This new option shouldn't affect `main_panel` only `panel`

Fixes [#https://github.com/avo-hq/avo/issues/2971](https://github.com/avo-hq/avo/issues/2971)

# Checklist:

## Screenshots & recording

https://github.com/user-attachments/assets/0bfa2d76-6ff3-4f31-80dd-3ab6453ca63c

## Manual review steps

1. Add `visible: -> { false }` to a panel
2. Make sure it hides all the panel header and all fields contained
3. Move set it back to true and they should reappear
4. Ensure it doesn't work when changing the visibility of `main_panel`
5. Ensure it doesn't hide anything when visible isn't set


icaroryan:
Ayy, my first collaboration!
I was wondering where I'd add automated tests for this feature. I couldn't find the proper file to write them

Paul-Bob:
Hello @icaroryan you made it! 

Thanks for the contribution, it's looking good.

Most Avo test files are under the `spec/features/avo/**` and `spec/system/avo/**` paths.

We have [this test](https://github.com/avo-hq/avo/blob/469636c685bc55e44e7cede6061b637e2e174d0b/spec/features/avo/tabs_panels_and_sidebar_visibility_spec.rb) that already checks the visibility of tabs, panels, and the sidebar. The purpose of this test is to ensure that none of these elements are rendered if their respective fields are hidden. You can add a test case for panel visibility block inside it.

icaroryan:
Thank you @Paul-Bob! Just pushed the new tests

icaroryan:
With that being said, I think it’s worth having the same option for tab. If you give me the go-ahead, I can push a PR for it. It would be basically the same thing we did with panels.

Paul-Bob:
> Thank you @Paul-Bob! Just pushed the new tests. I tried to follow the same structure of the existing ones

Looking great, Thanks!

DOCS PR (just to remain linked): https://github.com/avo-hq/docs.avohq.io/pull/248

> With that being said, I think it’s worth having the same option for tab. If you give me the go-ahead, I can push a PR for it. It would be basically the same thing we did with panels.

I would be happy to merge it, thanks for the initiative, we appreciate it!

PR #2991, opened 2024-07-16
Write tests for boolean group broken

binarygit:
# Description

This PR adds tests for the broken boolean group fix PR. 

It also makes sure such boolean group's check boxes are correctly checked in the edit view.

annotely_image

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/spec/dummy/app/models/user.rb
+++ b/spec/dummy/app/models/user.rb
@@ -48,6 +48,8 @@ class User < ApplicationRecord
   scope :admins, -> { where "(roles->>'admin')::boolean is true" }
   scope :non_admins, -> { where "(roles->>'admin')::boolean != true" }
 
+  attr_writer :permissions
+
   def is_admin?
     roles.present? && roles["admin"].present?
   end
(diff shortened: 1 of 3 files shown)


spec/dummy/app/models/user.rb
```diff
@@ -97,8 +99,4 @@ def permissions
       delete: true,
     }
   end
-
-  def permissions=(value)
-    # assert here the value is a hash and has the right values sent in by the user
-  end
```
binarygit, line 103: @adrianthedev I thought it'd be nicer to not have a body for this at all and just have this method created via the setter. WDYT?

Paul-Bob:
@binarygit, Let's keep the old tests for roles to ensure nothing breaks. Let's instead add some new test cases for permissions without modifying the roles tests.

binarygit:
All the old tests are there. I just refactored them a little bit and added the new tests but the diff is showing that I changed the whole thing for some reason.

spec/dummy/app/models/user.rb
adrianthedev: Oh, so basically set the value on the model instance class. 
Yeah, that could work too. 
Please as a comment on the writer why it's there.

adrianthedev (approved):
Looks good, thank you!

PR #2997, opened 2024-07-16
[#2996] - Change boolean field to show dash when null

### LINKED ISSUE #2996: Boolean field always treats null as falsey
### Describe the bug

When a boolean field is nullable, it shows false as a red X instead of a dash.

### Steps to Reproduce
Create a boolean field that has a false value.

The field should display on index and show as a dash "-".

### System configuration
**Avo version**:

**Rails version**:

**Ruby version**:

**License type**:

 - [ ] Community
 - [x] Pro
 - [ ] Advanced

 - [x] No

#### Impact

#### Urgency

serene:
# Description
When a boolean field is nullable, it shows false as a red X instead of a dash.
Fixes #2996.

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/app/components/avo/fields/boolean_field/index_component.html.erb
+++ b/app/components/avo/fields/boolean_field/index_component.html.erb
@@ -1,3 +1,3 @@
-<%= index_field_wrapper **field_wrapper_args, dash_if_blank: false, flush: true do %>
+<%= index_field_wrapper **field_wrapper_args, dash_if_blank: true, flush: true do %>
   <%= render Avo::Fields::Common::BooleanCheckComponent.new checked: @field.value %>
 <% end %>
--- a/app/components/avo/fields/boolean_field/index_component.html.erb
+++ b/app/components/avo/fields/boolean_field/index_component.html.erb
@@ -1,3 +1,3 @@
-<%= index_field_wrapper **field_wrapper_args, dash_if_blank: true, flush: true do %>
+<%= index_field_wrapper **field_wrapper_args, flush: true do %>
   <%= render Avo::Fields::Common::BooleanCheckComponent.new checked: @field.value %>
 <% end %>
--- a/app/components/avo/fields/boolean_field/show_component.html.erb
+++ b/app/components/avo/fields/boolean_field/show_component.html.erb
@@ -1,3 +1,3 @@
-<%= field_wrapper **field_wrapper_args, dash_if_blank: true do %>
+<%= field_wrapper **field_wrapper_args do %>
   <%= render Avo::Fields::Common::BooleanCheckComponent.new checked: @field.value %>
 <% end %>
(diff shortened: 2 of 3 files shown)


app/components/avo/fields/boolean_field/index_component.html.erb
```diff
@@ -1,3 +1,3 @@
-<%= index_field_wrapper **field_wrapper_args, dash_if_blank: false, flush: true do %>
+<%= index_field_wrapper **field_wrapper_args, dash_if_blank: true, flush: true do %>
```
Paul-Bob, line 1:
```suggestion
<%= index_field_wrapper **field_wrapper_args, flush: true do %>
```

app/components/avo/fields/boolean_field/show_component.html.erb
```diff
@@ -1,3 +1,3 @@
-<%= field_wrapper **field_wrapper_args, dash_if_blank: false do %>
+<%= field_wrapper **field_wrapper_args, dash_if_blank: true do %>
```
Paul-Bob, line 1:
```suggestion
<%= field_wrapper **field_wrapper_args do %>
```

Paul-Bob (changes requested):
Hei @serene thanks for looking into this! It's looking good

Since the default value for `dash_if_blank` is `true` ([source code](https://github.com/avo-hq/avo/blob/fix/policies/app/components/avo/field_wrapper_component.rb#L15))let's remove it from the arguments. 

Otherwise, it's good to go!

Paul-Bob (approved):
Since the changes from the last review were minor, I went ahead and applied them to merge this PR

Thank you @serene!

donnyli95:
Hey @Paul-Bob, is this going to be released in avo 3.10.7? When will that be released? I've tried pointing to the main avo branch for now to be able to use the nullable dash but `bundle exec rake avo:sym_link` doesn't seem to work since it also depends on the avo-pro gem. 

My gemfile setup looks like:
```
gem 'avo', github: 'avo-hq/avo', branch: 'main'
gem "avo-pro", ">= 3.10.6", source: "https://packager.dev/avo-hq/"
```

and `bundle exec rake avo:sym_link` results in a `[Avo->] Failed to find avo.base.css.` error

Paul-Bob:
Hello, yes, it will be available on the next release, 3.10.7, can't say when since we do not have exact dates for patch releases, only for minor ones that are always on the first Tuesday of the month.

We will investigate the error that you're experiencing.

I normally don't encourage monkey patching, but in this case, since it's a minor change, you can eject (https://docs.avohq.io/3.0/eject-views.html#field_components) the affected components and apply the changes on th3 ejected ones. If you do so, make sure to remove the monkey patch after the official release.

donnyli95:
@Paul-Bob I've just pushed my app with the latest avo update 3.10.7 and noticed that the false values for the boolean fields are also coming out as dashes. 

For example if i write: 
```
field 'cheese', as: :boolean, do
  false
end
```

this comes out as a dash instead of the original red X. Just wanted to confirm if you've also noticed this issue or not?

Paul-Bob:
You're right @donnyli95 I'll have a look.

Paul-Bob:
@donnyli95 just letting you know that it has been fixed since version 3.10.8

PR #3004, opened 2024-07-17
Feature: Run callbacks on through model when detaching association

### LINKED ISSUE #2512: Detaching a record via an association calls delete_all, skipping ActiveRecord callbacks, instead of destroy.
### Describe the bug
Detaching an association record calls `delete_all` instead of `destroy` or `destroy_all`. This skips any ActiveRecord callbacks that may exist on the association join record.

### Steps to Reproduce
1. Setup some basic rails models like the example below for a top level model called `Facility`.
2. Generate the Avo resources and add the has_many associations so that you can attach/detach `facilities` on the `user` resource.
3. Detach a `facility` from a `user`. The callback `before_destroy` is not hit within `FacilityUser`. Rails console logs also show _FacilityUser Delete All_ instead of something like _FacilityUser Destroy_.

Although `delete_all` is more performant, we would expect `destroy` or `destroy_all` to be called upon detaching so that any ActiveRecord callbacks are hit. Is there a reason that deletion was chosen?

### Models and resource files

```rb
# Facility
has_many :facility_users, dependent: :destroy
has_many :users, through: :facility_users

# User
has_many :facility_users, dependent: :destroy
has_many :facilities, through: :facility_users

# FacilityUser
before_destroy :do_something

belongs_to :facility
belongs_to :user

def do_something
 ...
end
```

### System configuration
**Avo version**: 3.4.1
**Rails version**: 7.0.8
**Ruby version**: 3.2.2

**License type**:
 - [ ] Community
 - [ ] Pro
 - [x] Advanced

 - [x] No

#### Impact

#### Urgency

binarygit:
# Description

Fixes #2512

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/spec/dummy/app/models/user.rb
+++ b/spec/dummy/app/models/user.rb
@@ -36,7 +36,6 @@ class User < ApplicationRecord
   has_many :people
   has_many :spouses
   has_many :comments
-  has_many :team_memberships
   has_and_belongs_to_many :projects, inverse_of: :users
   has_many :team_memberships
   has_many :teams, through: :team_memberships
--- a/spec/dummy/app/models/user.rb
+++ b/spec/dummy/app/models/user.rb
@@ -38,7 +38,9 @@ class User < ApplicationRecord
   has_many :comments
   has_many :team_memberships
   has_and_belongs_to_many :projects, inverse_of: :users
-  has_and_belongs_to_many :teams, join_table: :team_memberships, inverse_of: :admin
+  has_many :team_memberships
+  has_many :teams, through: :team_memberships
+
 
   has_one_attached :cv
 
--- a/spec/features/avo/has_many_field_spec.rb
+++ b/spec/features/avo/has_many_field_spec.rb
@@ -162,7 +162,6 @@ RSpec.feature "HasManyField", type: :feature do
     it "triggers callbacks on through model" do
       team.team_members << user
       expect(team.team_members.count).to eq 1
-      team_membership = team.memberships.first
 
       visit "/admin/resources/teams/#{team.id}/team_members?view=show&turbo_frame=has_many_field_show_team_members"
 
@@ -172,7 +171,6 @@ RSpec.feature "HasManyField", type: :feature do
     it "triggers callbacks when called from the other model too" do
       user.teams << team
       expect(user.teams.count).to eq 1
-      team_membership = user.team_memberships.first
 
       visit "/admin/resources/users/#{user.to_param}/teams?view=show&turbo_frame=has_and_belongs_to_many_field_show_teams"
 
--- a/spec/features/avo/has_many_field_spec.rb
+++ b/spec/features/avo/has_many_field_spec.rb
@@ -154,4 +154,29 @@ RSpec.feature "HasManyField", type: :feature do
       }.not_to raise_error
     end
   end
+
+  describe "through relationship" do
+    let!(:team) { create :team }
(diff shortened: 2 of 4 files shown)


binarygit:
@Paul-Bob 
```
# team.rb
  has_many :team_members, through: :memberships, class_name: "User", source: :user, inverse_of: :teams

# user.rb
  has_and_belongs_to_many :teams, join_table: :team_memberships, inverse_of: :admin
```

Here team defines it's team_members assoc using a through table however, User defines the same association as a HABTM.

Detaching a user from the team show page will call `team_membership.destroy` but because user has setup this assoc as HABTM I don't think when detaching a team from the user's show page should call `team_membership.destroy`. Am I correct?

Paul-Bob:
> Update: Adrian and I talked about this. user.rb should be declaring the relationship as a through relationship. It's unusual to declare a relationship as a through at one end and a HABTM on the other end.

Agree, I also noticed that particularity on our dummy recently. 

Does this PR work well from both sides if the associations are correctly defined?

spec/features/avo/has_many_field_spec.rb
```diff
@@ -154,4 +154,29 @@
...
+      team.team_members << user
+      expect(team.team_members.count).to eq 1
+      team_membership = team.memberships.first
+
+      visit "/admin/resources/teams/#{team.id}/team_members?view=show&turbo_frame=has_many_field_show_team_members"
+
+      expect { find("tr[data-resource-id='#{user.to_param}'] [data-control='detach']").click }.to raise_error("Callback Called")
+    end
+
+    it "triggers callbacks when called from the other model too" do
```
binarygit, line 172: Yes it works both ways of the association @Paul-Bob. This test verifies that :sweat_smile:

spec/dummy/app/models/user.rb
```diff
@@ -38,7 +38,9 @@ class User < ApplicationRecord
   has_many :comments
   has_many :team_memberships
   has_and_belongs_to_many :projects, inverse_of: :users
-  has_and_belongs_to_many :teams, join_table: :team_memberships, inverse_of: :admin
+  has_many :team_memberships
```
Paul-Bob, line 41:
```suggestion
```

Same association already defined (L39)

Paul-Bob (changes requested):
Looking good! Thanks @binarygit, let's please remove the duplicated association and also solve the conflicts.

binarygit:
Wait. I’m working on displaying the messages when associations are misconfigured

spec/dummy/app/models/user.rb
binarygit: Thanks. I didn't notice this!

binarygit:
@Paul-Bob I couldn't find a straightforward way to check if associations are properly configured. Sorry for holding this up, maybe merge this now and I'll pick this up again later.

binarygit:
I actually talked about this here: https://rubyonrails-link.slack.com/archives/C05054QPL/p1721928807967489 and I don't think verifying associations are setup correctly should be our job. 

I did try and it's too much work and also feels like it should be upto the user (like the devs point out in the link) to set it up correctly.

Paul-Bob:
It's ok without the association check @binarygit thanks for looking into it

adrianthedev (approved):
Looks good.

Paul-Bob (approved):
It's looking good, thanks @binarygit!

PR #3015, opened 2024-07-18
Fix: Always display persisted image in the edit view

### LINKED ISSUE #3008: Handle case when record is invalid and new image is attached
### Describe the bug

Avo breaks when the user adds a new image, makes an error in other form field, and tries to save/update the record.

### Steps to Reproduce

I was able to reproduce the issue using the [main.avodemo.com](https://main.avodemo.com) app:

1. [Open Posts](https://main.avodemo.com/avo/resources/posts/)
2. Edit existing post:
- remove `name` (which is required)
- attach new `cover photo`
- click on save

_Expected behavior:_
The form is displayed again with errors.

_Actual behavior:_
App crashes with `ArgumentError: "Cannot get a signed_id for a new record"`. 
The problem is with `app/components/avo/fields/common/files/view_type/grid_item_component.html.erb`, line: 5. `helpers.main_app.url_for(file)` does not work. 

### System configuration

Avo version: 3.10.6

Rails version: 7.2 beta

**License type**:

 - [ ] Community
 - [ ] Pro
 - [x] Advanced

#### Impact

#### Urgency

binarygit:
# Description

Fixes #3008

# Checklist:

## Manual review steps

1. Navigated to /resources/post/<id>/edit
2. Deleted text from name field
3. Clicked on Browse and attached a new image
4. Hit save
5. Ensured that error message is rendered on screen and app doesn't crash.


Paul-Bob (changes requested):
Noticed that a dash (`–`) remains while reproducing the bug, let's try to remove the dash in that scenario.

# Expected (trying other flux just to show the UI expectations):

[expected.webm](https://github.com/user-attachments/assets/4cb300f5-cfcd-423f-8a57-3fec1e45c46b)


# Actual: 

[actual.webm](https://github.com/user-attachments/assets/787ecd79-0a9e-4760-a2d4-2a042f7ebedc)

adrianthedev:
Yes, we need to take into account how it behaves when updating.

binarygit:
https://github.com/user-attachments/assets/0929f848-d897-4453-944d-87cae067248c

@Paul-Bob Pushed a commit which fixes this.

Paul-Bob:
@binarygit still the dash appear on show view when there is no file attached?

binarygit:
@Paul-Bob yes, sir :saluting_face: . It does

image

Paul-Bob:
I noticed a strange behavior, here is a video that explains it:

[strange_behaviour.webm](https://github.com/user-attachments/assets/79482ec1-40aa-4e33-9c2d-fb4cc48f382f)

lib/avo/fields/base_field.rb
```diff
@@ -162,6 +162,12 @@ def value(property = nil)
           final_value = computed_default_value
         end
 
+        # On edit view always show the persisted image. Related: issue#3008
+        if final_value.instance_of?(ActiveStorage::Attached::One) && final_value.image? && view == "edit"
+          persisted_record = record.class.find_by(id: record.id)
+          final_value = persisted_record.send(property)
+        end
```
adrianthedev, line 169: THis should go to the file field or somewhere else. THis doesn't pertaing to the base field

adrianthedev (changes requested):
!

app/components/avo/fields/common/files/view_type/grid_item_component.html.erb
```diff
@@ -1,5 +1,9 @@
 <div id="<%= dom_id file %>" class="relative min-h-full max-w-full flex-1 flex flex-col justify-between space-y-2">
-  <% if file.present? %>
+  <%# When editing a resource, if you attach a file and also edit another field which causes a validation error.
+      In such a situation, file is an object that hasn't been saved to the database (i.e id == nil) because
+      the resource update failed and so the attachment wasn't persisted.
+      In such situations url_for(file) will complain that it can't find the url for an non-existent file. %>
+  <% if file.persisted? %>
```
Paul-Bob, line 6: Is this change still necessary if we always display the persisted image?

binarygit: I think it's better than `file.present?` because it's more clear that the code inside the block needs to execute only if we have a persisted file.

lib/avo/fields/base_field.rb
```diff
@@ -162,6 +162,12 @@ def value(property = nil)
...
+
```
Paul-Bob, line 170: Let's remove this since it's already on `file_field.rb`

lib/avo/fields/file_field.rb
```diff
@@ -24,6 +24,17 @@ def initialize(id, **args, &block)
       def path
         rails_blob_url value
       end
+
+      def value
+        final_value = super
+        # On edit view always show the persisted image. Related: issue#3008
+        if final_value.instance_of?(ActiveStorage::Attached::One) && final_value.image? && view == "edit"
```
Paul-Bob, line 31: > `final_value.image?`

Should we apply this only to images? I think we should remove that guard and apply this behavior to all file types, WDYT?

Paul-Bob (changes requested):
Let's remove the duplicated code and verify if we should keep same approach for all file types

lib/avo/fields/file_field.rb
binarygit: Agreed. I'll look into it

app/components/avo/fields/common/files/view_type/grid_item_component.html.erb
Paul-Bob: Not saying `file.present?` is better, just wondering if this check is necessary or redundant since we already guarantee to show only persisted files

binarygit: True. I removed the conditional. There's a failing test in CI but it passes locally for me.

app/components/avo/fields/common/files/view_type/grid_item_component.html.erb
```diff
@@ -1,27 +1,23 @@
 <div id="<%= dom_id file %>" class="relative min-h-full max-w-full flex-1 flex flex-col justify-between space-y-2">
-  <% if file.present? %>
```
adrianthedev, line 2: The field wrapper will intercept and not render this component. That's why we don't need the `file.present?` line.

adrianthedev (approved):
Looks good. haven't manually tested the scenario.

lib/avo/fields/file_field.rb
```diff
@@ -24,6 +24,16 @@ def initialize(id, **args, &block)
       def path
         rails_blob_url value
       end
+
+      def value
+        final_value = super
+        # On edit view always show the persisted image. Related: issue#3008
+        if final_value.instance_of?(ActiveStorage::Attached::One) && view == "edit"
+          persisted_record = record.class.find_by(id: record.id)
+          final_value = persisted_record.send(@for_attribute || id)
```
Paul-Bob, line 33: @binarygit do you think it's possible to use some active record helper to reload `@for_attribute || id` from the `@record` while keeping other attributes dirty? This way we can avoid the manual `find_by` query...

binarygit: I’ll look into it paul

binarygit: @Paul-Bob I looked but there's no straight forward way to do this. I tried to dup the record and clear specific changes but the changes propagated to the original record too which we shouldn't let happen.

PR #3017, opened 2024-07-18
feature: add visible option to tab

icaroryan:
# Description
This PR adds a `visible` option to Tab, which should hide all the contained fields as well as the tab itself.

This new option shouldn't affect `tabs` only `tab`

# Checklist:

## Screenshots & recording

https://github.com/user-attachments/assets/92d3e602-de77-42b8-83c8-6956ff38a2a3

## Manual review steps

1. Add `visible: -> { false }` to a `tab`
2. Make sure it hides all the tab header and all fields contained
3. Move set it back to true and they should reappear
5. Ensure it doesn't hide anything when visible isn't set

### DIFF
--- a/lib/avo/resources/items/tab.rb
+++ b/lib/avo/resources/items/tab.rb
@@ -4,6 +4,7 @@ class Avo::Resources::Items::Tab
   include Avo::Concerns::HasItems
   include Avo::Concerns::HasItemType
   include Avo::Concerns::VisibleItems
+  include Avo::Concerns::IsVisible
   include Avo::Concerns::VisibleInDifferentViews
 
   delegate :items, :add_item, to: :items_holder
@@ -16,6 +17,7 @@ class Avo::Resources::Items::Tab
     @items_holder = Avo::Resources::Items::Holder.new
     @view = Avo::ViewInquirer.new view
     @args = args
+    @visible = args[:visible]
 
     post_initialize if respond_to?(:post_initialize)
   end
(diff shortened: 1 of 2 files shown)


Paul-Bob (approved):
Thanks for the contribution @icaroryan!

Could you also make a PR with the docs, similar to what you did with the panel visible option?

BTW, would it be difficult to add the visible options to the `tabs` DSL too?

icaroryan:
Hi @Paul-Bob! I'll add documentation for it as soon as I can.

As for the `tabs,` DSL, the biggest struggle is just how args are being passed into the builder.
`add_item Avo::Resources::Items::TabGroup::Builder.parse_block(parent: @parent, id: id, name: name, **kwargs, &block) `

icaroryan:
I think I was rubber-ducking in the message above, I'm sorry lol, the solution was just simply accepting the **args in the argument list for the tabs constructor

https://github.com/avo-hq/avo/pull/3031/files

Paul-Bob:
> I think I was rubber-ducking in the message above, I'm sorry lol, the solution was just simply accepting the **args in the argument list for the tabs constructor
> 
> https://github.com/avo-hq/avo/pull/3031/files

No need to apologize! It looks like this was not such a challenge after all... :D 

Thank you!

icaroryan:
@Paul-Bob Yea, after looking a bit closer at the code I noticed it was just a naming difference. Thank you for the explanation <3

PR #2998, opened 2024-07-16
refactor: Restructure Avo::BaseResource for user customization

Yash-Singh-Pathania:
# Description

Fixes #2959 

# Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the [documentation](https://github.com/avo-hq/docs.avohq.io/pull/247)
- [ ] I have added tests that prove my fix is effective or that my feature works

## Manual review steps

1. Ran local test cases 
2. Did dev testing via rails console.

PR #3018, opened 2024-07-18
feature: add `default_sort_column` configuration option

### LINKED ISSUE #2982: Allow to customize the default sorting attribute
By default, records are sorted by the `created_at` attribute (if it exists)

Let's make it possible to globally customize the sorting attribute.

## Approach

- Add a configuration option (name to be decided)
- Apply that option if configured around [here](https://github.com/avo-hq/avo/blob/main/app/controllers/avo/base_controller.rb#L308)

icaroryan:
# Description

This PR adds the ability to customize the default sorting attribute. The sorting column for the index view is `created_at` by default, but now the user can customize that by setting the `default_sort_column` option in the resource file to the desired default sorting column.

We still use `created_at` as the default sorting column if:
* `default_sort_column` is not set
* `default_sort_column` is set, but the model doesn't have that column

This also allow for backwards compatibility.

Fixes #2982

# Checklist:

## Screenshots & recording (**UPDATED**)

https://github.com/user-attachments/assets/0837ffdb-c4b6-40e0-984b-6a7dc73dc6a9

## Manual review steps

Tip: You can use the console log to see what column is being used for sorting when fetching the records (or just look at the UI)

1. Open any model's index page
2. Verify that it's sorting by `created_at`
3. Set the `default_sort_column` to another column
4. Ensure the query reflects the changes approximately

### DIFF
--- a/app/controllers/avo/base_controller.rb
+++ b/app/controllers/avo/base_controller.rb
@@ -305,8 +305,15 @@ module Avo
       # Sorting
       if params[:sort_by].present?
         @index_params[:sort_by] = params[:sort_by]
-      elsif @resource.model_class.present? && @resource.model_class.column_names.include?("created_at")
-        @index_params[:sort_by] = :created_at
+      elsif @resource.model_class.present?
+        available_columns = @resource.model_class.column_names
+        default_sort_column = @resource.default_sort_column
+
+        if available_columns.include?(default_sort_column.to_s)
+          @index_params[:sort_by] = default_sort_column
+        elsif available_columns.include?("created_at")
+          @index_params[:sort_by] = :created_at
+        end
       end
 
       @index_params[:sort_direction] = params[:sort_direction] || :desc
--- a/lib/avo/resources/base.rb
+++ b/lib/avo/resources/base.rb
@@ -75,6 +75,7 @@ module Avo
       class_attribute :link_to_child_resource, default: false
       class_attribute :map_view
       class_attribute :components, default: {}
+      class_attribute :default_sort_column, default: :created_at
 
       # EXTRACT:
       class_attribute :ordering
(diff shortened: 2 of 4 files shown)


icaroryan:
Hey guys, feel free to nitpick. Any feedback is welcome! 😄

adrianthedev:
I somehow thought that this would be a configuration on the resource, not app-wide.
Interesting.
Let's think about pros and cons.

If we do it like this, then we won't be able to control it resource-wide.
If we do it as a resource configuration, you can override it in `BaseResource` (pending https://github.com/avo-hq/avo/pull/2998 to be merged) for all resources and still maintain the ability to change it for just a few resources.

What do you think?

icaroryan:
@adrianthedev I totally agree with you! I think I misunderstood the issue ticket because I actually wrote a draft message asking if having it on the resource would be better, but for some reason, I decided not to send it. I should’ve clarified it before getting started, but fortunately, the ticket is small enough that I can just change it to the resource. 
Thank you for the feedback!

icaroryan:
@adrianthedev I just updated the PR to move the option to the resource level instead! Thanks once again

Paul-Bob:
> @adrianthedev I totally agree with you! I think I misunderstood the issue ticket
 
No @icaroryan, you understood it right, I wrote it with the global configuration in mind, but I also agree with @adrianthedev's comment, is much more flexible to be configurable at the resource level.

> If we do it as a resource configuration, you can override it in BaseResource (pending https://github.com/avo-hq/avo/pull/2998 to be merged) for all resources and still maintain the ability to change it for just a few resources.

This is a great example of how to use the `BaseResource`, let's ensure that we document this.

Paul-Bob (approved):
Looking great @icaroryan thanks for the contribution!

Are you up to make the documentation for this PR?

adrianthedev:
Thank you for your contribution @icaroryan 💪

icaroryan:
@Paul-Bob, I can take care of the documentation for this one, too. Thanks for creating an issue for it

PR #3022, opened 2024-07-19
Add comment and lint

binarygit:
I still couldn't push directly @adrianthedev so created a pr for this one.


spec/dummy/app/models/user.rb
```diff
@@ -48,6 +48,8 @@ class User < ApplicationRecord
   scope :admins, -> { where "(roles->>'admin')::boolean is true" }
   scope :non_admins, -> { where "(roles->>'admin')::boolean != true" }
 
+  # Setter for permissions because it's a
+  # non-db backed boolean group
```
adrianthedev, line 52:
```suggestion
  # We're using a setter here because we want to test that the field is working properly with a non-db backed field.
```

Issue #2978, opened 2024-07-12
Fix eject generator --scope option


Paul-Bob:
`rails generate avo:eject --component Avo::Fields::TextField::ShowComponent --scope team`

This command should either work or display a notice on how to use it correctly but not raise errors.

PR #3027, opened 2024-07-19
fix: assets outputted multiple times

adrianthedev:
# Description

Fixes an issue where, sometimes, the asset files would be added multiple times.
Also fixes https://github.com/avo-hq/avo/issues/2978

# Checklist:

### DIFF
--- a/lib/generators/avo/eject_generator.rb
+++ b/lib/generators/avo/eject_generator.rb
@@ -146,7 +146,7 @@ module Generators
             [dest_rb, dest_erb].each do |path|
               if component.starts_with?("avo/views/")
                 modified_content = File.read(path).gsub("Avo::Views::", "Avo::Views::#{options[:scope].camelize}::")
-              elsif component.starts_with?("avo/fields/")
+              elsif component.starts_with?("avo/fields/") && options["field-components"].present?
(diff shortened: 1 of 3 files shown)


lib/generators/avo/eject_generator.rb
```diff
@@ -146,7 +146,7 @@ def eject_component(component_to_eject = options[:component], confirmation: true
             [dest_rb, dest_erb].each do |path|
               if component.starts_with?("avo/views/")
                 modified_content = File.read(path).gsub("Avo::Views::", "Avo::Views::#{options[:scope].camelize}::")
-              elsif component.starts_with?("avo/fields/")
+              elsif component.starts_with?("avo/fields/") && options["field-components"].present?
```
Paul-Bob, line 149: Is this fixing this issue? https://github.com/avo-hq/avo/issues/2978

Paul-Bob (approved):
Any idea why there are duplicated values inside the `@stylesheets` and `@javascripts` arrays? Can't we avoid the duplication from where the values are added to the array?

adrianthedev:
That would be the best fix. This is a hotfix.

lib/generators/avo/eject_generator.rb
Paul-Bob: I'm supposing it is

PR #3034, opened 2024-07-21
feature: custom search results

### LINKED ISSUE #3012: Implement support for custom search system
### Feature
Accoring to @adrianthedev we should be able to pass an `array` or `object` (rather than the standard `ActiveRecord::Relation`) within `query`.
Yet the exact syntax is not documented anywhere. 

@adrianthedev pulled out this snippet but doesn't seem to be the right one.

```ruby
class Avo::Resources::Person < Avo::BaseResource
  self.search = {
    query: -> { 
      [{
        _id: "123", _label: "abc", _url: "#"
      }]
    }
end
```

Could someone document how we can return output from our own search system (e.g. coming from elasticsearch, API, ...)

Paul-Bob:
# Description

Fixes #3012 

```ruby
  self.search = {
    query: -> {
      array = []

      30.times do |iteration|
        array << {
          _id: iteration,
          _label: "Label nr #{iteration}",
          _url: "https://#{iteration}.ro"
        }
      end

      array
    }
  }
```

[search_results.webm](https://github.com/user-attachments/assets/e5f1e5d3-c9c2-4a51-b192-e3d48fbd8adc)

# Checklist:

  - https://github.com/avo-hq/docs.avohq.io/pull/253

### DIFF
--- a/app/controllers/avo/search_controller.rb
+++ b/app/controllers/avo/search_controller.rb
@@ -47,14 +47,15 @@ module Avo
         query: resource.query_scope
       ).handle
 
+      # When using custom search services query should return an array of hashes
       if query.is_a?(Array)
         # Apply highlight
         query.map do |result|
           result[:_label] = highlight(result[:_label].to_s, CGI.escapeHTML(params[:q] || ""))
         end
 
-        # Get the count
-        results_count = query.size
+        # Force count to 0 until implement an API to pass the count
+        results_count = 0
 
         # Apply the limit
         results = query.first(search_results_count(resource))
--- a/app/controllers/avo/search_controller.rb
+++ b/app/controllers/avo/search_controller.rb
@@ -47,15 +47,28 @@ module Avo
         query: resource.query_scope
       ).handle
 
-      query = apply_scope(query) if should_apply_any_scope?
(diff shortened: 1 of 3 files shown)


app/controllers/avo/search_controller.rb
```diff
@@ -212,5 +204,33 @@ def search_results_count(resource)
...
+    def parse_results(query)
+      # When using custom search services query should return an array of hashes
+      if query.is_a?(Array)
+        # Apply highlight
+        query.map do |result|
+          result[:_label] = highlight(result[:_label].to_s, CGI.escapeHTML(params[:q] || ""))
+        end
+
+        # Force count to 0 until implement an API to pass the count
+        results_count = 0
```
adrianthedev, line 217: Is this hidden to the user or can they see `(0)` in the search results?

adrianthedev (approved):

app/controllers/avo/search_controller.rb
Paul-Bob: Is hidden. The count is added only when bigger than 0:

```ruby
      if results_count > 0
        header = "#{header} (#{results_count})"
      end
```

PR #3043, opened 2024-07-22
Use Rails.application.secret_key_base directly

jpawlyn:
# Description
Port of https://github.com/avo-hq/avo/pull/2417 to the main branch

Fixes # (issue)

Removes the deprecation message output when running tests:
```
DEPRECATION WARNING: `Rails.application.secrets` is deprecated in favor of `Rails.application.credentials` and will be removed in Rails 7.2. (called from secret_key_base at /Users/jpawlyn/.rbenv/versions/3.3.4/lib/ruby/gems/3.3.0/gems/avo-3.10.6/lib/avo/services/encryption_service.rb:42)
```

# Checklist:

### DIFF
--- a/lib/avo/services/encryption_service.rb
+++ b/lib/avo/services/encryption_service.rb
@@ -28,28 +28,7 @@ module Avo
       private
 
       def encryption_key
-        secret_key_base[0..31]
-      end
-
-      def secret_key_base
-        # Try to fetch the secret key base from ENV or the credentials file
-        key = ENV["SECRET_KEY_BASE"] || Rails.application.credentials.secret_key_base
-
-        # If key is blank and Rails version is less than 7.2.0
-        # Try to fetch the secret key base from the secrets file
-        # Rails 7.2.0 made secret_key_base from secrets obsolete
-        if key.blank? && (Gem::Version.new(Rails.gem_version) < Gem::Version.new("7.2.0"))
-          key = Rails.application.secrets.secret_key_base
-        end
-
-        return key if key.present?
-
-        # Avoid breaking in production
-        # All features relying on encryption will not work properly without a configured secret key base
-        return SecureRandom.random_bytes(32) if Rails.env.production?
-
-        raise "Unable to fetch secret key base. Please set it in your credentials or environment variables\n" \
-          "For more information check https://docs.avohq.io/3.0/encryption-service.html#secret-key-base"
+        Rails.application.secret_key_base[0..31]
       end
     end
   end


Paul-Bob:
What if `ENV["SECRET_KEY_BASE"] || credentials.secret_key_base || secrets.secret_key_base` (called by `Rails.application.secret_key_base` in production) returns `nil`?

`Rails.application.secret_key_base[0..31]` will break right?

lib/avo/services/encryption_service.rb
```diff
@@ -28,28 +28,7 @@ def decrypt
...
-
-        return key if key.present?
-
-        # Avoid breaking in production
-        # All features relying on encryption will not work properly without a configured secret key base
-        return SecureRandom.random_bytes(32) if Rails.env.production?
-
-        raise "Unable to fetch secret key base. Please set it in your credentials or environment variables\n" \
-          "For more information check https://docs.avohq.io/3.0/encryption-service.html#secret-key-base"
+        Rails.application.secret_key_base[0..31]
```
Paul-Bob, line 31:
```suggestion
        Rails.application.secret_key_base[0..31]
      rescue
        # Avoid breaking in production
        # All features relying on encryption will not work properly without a configured secret key base
        return SecureRandom.random_bytes(32) if Rails.env.production?

        raise "Unable to fetch secret key base. Please set it in your credentials or environment variables\n" \
          "For more information check https://docs.avohq.io/3.0/encryption-service.html#secret-key-base"
```

jpawlyn:
> Thank you for the contribution @jpawlyn!
> 
> What if `ENV["SECRET_KEY_BASE"] || credentials.secret_key_base || secrets.secret_key_base` (called by `Rails.application.secret_key_base` in production) returns `nil`?
> 
> `Rails.application.secret_key_base[0..31]` will break right?
> 
> Let's add a rescue, which generates a random key on production just to keep the app running and show an error on development.

Looks like Rails takes care of this scenario - please see https://github.com/rails/rails/blob/a17aa6376d7974e38af90a2fe28e548b201baa63/railties/lib/rails/application/configuration.rb#L529

And when I ran my Rails 7.1.3.4 app server in production mode without a secret key base configured, puma never started due to the error:
```
3.3.0/gems/railties-7.1.3.4/lib/rails/application.rb:658:in `validate_secret_key_base': Missing `secret_key_base` for 'production' environment, set this string with `bin/rails credentials:edit` (ArgumentError)
```

adrianthedev:
Looks great! Thanks @jpawlyn!

Issue #3014, opened 2024-07-18
Pagy 9.0.0 not compatible with Avo 3.10.6


jetienne:
### Describe the bug

Pagy 9.0.0 not compatible with Avo 3.10.6

### System configuration
**Avo version**:
avo (3.10.6)

**Rails version**:
rails (7.2.0.beta3)

**Ruby version**:
ruby 3.3.4 (2024-07-09 revision be1089c8ec) [arm64-darwin23]

**License type**:

 - [ ] Community
 - [ ] Pro
 - [x] Advanced

**Are you using Avo monkey patches, overriding views or view components?**

 - [ ] Yes. If so, please post code samples.
 - [x] No

### Screenshots or screen recordings

![image](https://github.com/user-attachments/assets/e8bbecf1-a586-4057-a643-90c8226dc823)

#### Impact

 - [ ] High impact (It makes my app un-usable.)
 - [ ] Medium impact (I'm annoyed, but I'll live.)
 - [x] Low impact (It's really a tiny thing that I could live with.)

#### Urgency

 - [ ] High urgency (I can't continue development without it.)
 - [ ] Medium urgency (I found a workaround, but I'd love to have it fixed.)
 - [x] Low urgency (It can wait. I just wanted you to know about it.)

binarygit:
Hey @jetienne Thanks for creating this issue. We really appreciate you letting us know about this. We will have a fix for this shortly!

giedriusr:
Same here, but today was too lazy to open an issue :))

tvararu:
To work around this for now, you can manually specify an older Pagy version in your `gemfile`:

```ruby
gem "pagy", "< 9"
```

rickychilcott:
Thanks. I didn't find this issue immediately because I searched for `@pagy.items`. 

I'm adding this comment so hopefully, others will see it!

gkunwar:
Thanks @jetienne creating this issue. As I'm first time user, was struggling to fix. Finally found work around. Thanks @tvararu

adrianthedev:
I found this to be the list of breaking changes.
https://ddnexus.github.io/pagy/changelog/#simple-renaming

Would someone be able to take this and make Avo compatible with pagy 9 as well?
This version shoul dbe added to the current pagy versions we support, not replace them. Some folks are using pagy outside Avo and they need compatibility with versions 7 or 8.

Thank you!

Paul-Bob:
[...]

PR #3046, opened 2024-07-23
pagy v9 compatibility, backwards compatible

yshmarov:
Fixes https://github.com/avo-hq/avo/issues/3014
pagy upgrade docs: https://ddnexus.github.io/pagy/changelog/#version-900

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -185,6 +185,6 @@ gem "avo-money_field"
 gem "avo-record_link_field"
 
 # gem "pagy", "< 8.0.0"
-gem "pagy", "> 8"
+gem "pagy", "> 9"
 
 gem "csv"
--- a/Gemfile
+++ b/Gemfile
@@ -185,6 +185,6 @@ gem "avo-money_field"
 gem "avo-record_link_field"
 
 # gem "pagy", "< 8.0.0"
-gem "pagy", "> 9"
+gem "pagy", "> 8"
 
 gem "csv"
--- a/config/initializers/pagy.rb
+++ b/config/initializers/pagy.rb
@@ -1,5 +1,6 @@
 require "pagy/extras/trim"
 require "pagy/extras/countless"
+require 'pagy/extras/size'
 
 # For locales without native pagy i18n support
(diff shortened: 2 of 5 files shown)


Gemfile
```diff
@@ -185,6 +185,6 @@ gem "avo-money_field"
 gem "avo-record_link_field"
 
 # gem "pagy", "< 8.0.0"
-gem "pagy", "> 8"
+gem "pagy", "> 9"
```
yshmarov, line 188: We don't have to commit the Gemfile changes right now, to ensure that we don't break other apps.

adrianthedev:
Thank you @yshmarov!
We appreciate your contribution ❤️

Gemfile
Paul-Bob: It's ok, this Gemfiles does not reflect on avo's dependencies, those are specified in `avo.gemspec`

Paul-Bob (approved):
Thank you @yshmarov!

config/initializers/pagy.rb
```diff
@@ -1,5 +1,6 @@
 require "pagy/extras/trim"
 require "pagy/extras/countless"
+require "pagy/extras/size"
```
Paul-Bob, line 3: Let's keep this for now.

I'm creating an issue to use integer sizes in favor of using the array legacy API.

iainbeeston:
This has actually broken my rails application that uses avo. On app startup I get a `cannot load such file -- pagy/extras/size`
error from line 3 of `config/initializers/pagy.rb`in avo. Looking at pagy on GitHub, it looks like `pagy/extras/size` was only added a week ago in pagy 9.0 and older versions of page don't have this file (and will crash). I'm using 7.0.11.

I can see #3049 there's a change scheduled that removes the need for this PR, can that get prioritised or this PR reverted please?

adrianthedev:
yup. On it @iainbeeston

Paul-Bob:
Fixed on 3.10.9

simi:
Any chance to backport to 2.x?

PR #3002, opened 2024-07-17
fix: errors when attaching

Paul-Bob:
# Description

Fixes #2883 by showing attachment errors.

[attachments_error.webm](https://github.com/user-attachments/assets/88f301c7-d692-4b88-8dde-fdee93389691)

# Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the [documentation](https://github.com/avo-hq/avodocs)
- [x] I have added tests that prove my fix is effective or that my feature works

PR #3048, opened 2024-07-24
Feature: Display additional fields when attaching a record to another

### LINKED ISSUE #2472: Fields for additional column values/associations while attaching a record to another
### Feature
_Apologies if this does exist or there is a suggested way to handle this, but we could not find a mention of it in the 3.0 docs._

The ability to specify additional column values or associations while attaching a record, via a join record, to another.

### Example scenario:
Assume you have a `team` record that has many `users` via a join record called `team_users`. In addition to a `team_id` and `user_id` column on the join record, assume there is a third non-nullable column/association on it called `role_id`.

When trying to an attach a `user` to a `team` via team's `show` page, an attachment will fail because a `team_user` record needs that `role` association value set at the time of creation. In reference to the popup overlay that shows when clicking "Attach user", is there any way to specify additional column values or associations to be set on the join record at that time?

### Current workarounds
- Only current workaround would be to set a column default value and adjust it after attachment/creation.

binarygit:
# Description

Fixes #2472

Documentation PR: https://github.com/avo-hq/docs.avohq.io/pull/255

# Checklist:

### When attaching fails:
https://github.com/user-attachments/assets/d0ed33f9-a03f-4f79-9fa7-02974f185b4c

### When attaching is successful:
https://github.com/user-attachments/assets/a7e80753-d971-4f54-bc26-ed97a2bb57ae

## Manual review steps

1. Step 1
1. Step 2


Paul-Bob (changes requested):
There are some conflicts (I guess from [this PR](https://github.com/avo-hq/avo/pull/3002)) and i18n test is failing.

Let's fix those first please

When fixing the conflicts let's keep the turbo stream response and incorporate the `t("avo.attachment_failed", attachment_class: @related_resource.name)` in the errors list.

binarygit:
@Paul-Bob I can fix the i18n tests before we merge this. It's complaining about missing keys in other i18n files. Currently I will need to go into every one of them and put the key in and we might change the words or something so it might get repetitive.

Could you please review this and I'll make sure I add those keys in before we merge.

Paul-Bob (changes requested):
.

app/controllers/avo/associations_controller.rb
```diff
@@ -194,5 +204,22 @@ def has_many_reflection?
...
+    end
+
+    def new_join_record_params
+      field_names = Avo::FieldsExecutionContext.new(target: @field.extra, parent: @resource).detect_fields.items.map { |field| field.name.tr(" ", "_").downcase }
+      additional_params.permit(field_names).merge({source_foreign_key => @attachment_record.id, through_foreign_key => @record.id})
+    end
+
+    def new_join_record
+      reflection.through_reflection.klass.new(new_join_record_params)
+    end
```
Paul-Bob, line 223: Params are directly used as values. 

As example, we can see that `update_using` option is not respected:
```ruby
field :patrons, as: :has_many, through: :patronships,
      translation_key: "patrons",
      extra: -> {
        field :review, as: :text, update_using: -> { ">> #{value} <<"}
      }
```

Let's try to use the `fill_field` for each value to ensure that all field options are respected and applied.

Paul-Bob (changes requested):
It's looking good! Nice progress @binarygit!

Let's try to use the fields to parse the values that come through params.

app/controllers/avo/associations_controller.rb
binarygit: Sorry, didn't know about `fill_record` :eyes: . I just addressed this and pushed @Paul-Bob. Please take a look and let me know :pray:

lib/avo/resources/base.rb
```diff
@@ -359,7 +361,7 @@ def divider(label = nil)
       end
 
       def hydrate(...)
-        super(...)
+        super
```
Paul-Bob, line 364: Will `super` have access to all the method arguments without explicitly passing them?

lib/avo/resources/base.rb
```diff
@@ -462,6 +464,25 @@ def fill_record(record, params, extra_params: [])
...
+        end
+
+        # Write the user configured extra params to the record
+        if extra_params.present?
+          # Let Rails fill in the rest of the params
+          record.assign_attributes params.permit(extra_params)
+        end
+
+        record
+      end
```
Paul-Bob, line 484: There is only 1 line different from `fill_record`. Can we adapt the `fill_record` method to also work with this? Something like: (UNTESTED)

```ruby
      def fill_record(record, params, extra_params: [], fields: nil)
        # Write the field values
        params.each do |key, value|
          field = if fields.present?
            fields.find { |f| f.id == key.to_sym }
          else
            fields_by_database_id[key]
          end

          next unless field.present?

          record = field.fill_field record, key, value, params
        end

        # Write the user configured extra params to the record
        if extra_params.present?
          # Let Rails fill in the rest of the params
          record.assign_attributes params.permit(extra_params)
        end

        record
      end
```

Otherwise we need to maintain 2 identic methods

Paul-Bob (changes requested):
Hey @binarygit we're almost there!

I left some comments, let's prioritize the `fill_record` method duplication.

lib/avo/resources/base.rb
binarygit: Yes, super has access to all the method args without explicitly passing them. This change was done by standardrb when I linted btw.
image

lib/avo/resources/base.rb
binarygit: @Paul-Bob I pushed just now addressing all of your comments. Could you take a look please? :pray:

app/controllers/avo/associations_controller.rb
```diff
@@ -78,7 +80,13 @@ def create_association
...
 
       perform_action_and_record_errors do
-        if has_many_reflection?
+        if through_reflection?
+          if additional_params?
+            new_join_record.save
+          else
+            @record.send(association_name) << @attachment_record
+          end
+        elsif has_many_reflection?
```
Paul-Bob, line 89:
```suggestion
        if through_reflection? && additional_params?
          new_join_record.save
        elsif has_many_reflection? || through_reflection?
```

Paul-Bob: Let's try to avoid repeated code like `@record.send(association_name) << @attachment_record

app/views/avo/associations/new.html.erb
```diff
@@ -41,6 +41,9 @@
                 class: input_classes('w-full'),
               }
             %>
+          <% @extra&.each_with_index do |ex, index| %>
+            <%= render(Avo::Items::SwitcherComponent.new(resource: @related_resource, item: ex, index: index + 1, view: @view, form: form, field_component_extra_args: {stacked: true})) %>
```
Paul-Bob, line 45: Why forcing `stacked` to `true`?

Paul-Bob (changes requested):
I just noticed that on the associations controller we're using the `reflection` method which fetches the reflection on each usage. 

There is this `@reflection` ivar that is setted by `set_reflection` on each request.

Let's use `@reflection` instead `reflection` and remove the `reflection` method for performance sake.

app/controllers/avo/associations_controller.rb
```diff
@@ -78,7 +80,9 @@ def create_association
       association_name = BaseResource.valid_association_name(@record, association_from_params)
 
       perform_action_and_record_errors do
-        if has_many_reflection?
+        if through_reflection? && additional_params?
+          new_join_record.save
+        elsif has_many_reflection?
```
Paul-Bob, line 85:
```suggestion
        elsif has_many_reflection? || through_reflection?
```

Let's ensure that `@record.send(association_name) << @attachment_record` is executed when `through_reflection?` but there are no additional params

app/views/avo/associations/new.html.erb
```diff
@@ -33,14 +34,21 @@
...
+                as: :select,
+                form: form,
+                name: "related_id",
+                options: options_for_select(@options,
+                nil),
+                stacked: true,
+                classes: 'w-full',
+                name: @related_resource.name.downcase
+              %>
+              <% @extra_fields&.each_with_index do |ex, index| %>
```
adrianthedev, line 48: What is `ex`? can we rename it?

app/views/avo/associations/new.html.erb
```diff
@@ -33,14 +34,21 @@
...
+                <%= render(Avo::Items::SwitcherComponent.new(resource: @related_resource, item: ex, index: index + 1, view: @view, form: form, field_component_extra_args: {compact: true, stacked: true})) %>
```
adrianthedev, line 49:
```suggestion
                <%= render(
                  Avo::Items::SwitcherComponent.new(
                    resource: @related_resource,
                    item: ex,
                    index: index + 1,
                    view: @view,
                    form: form,
                    field_component_extra_args: {compact: true, stacked: true}
                  )
                ) %>
```

app/controllers/avo/associations_controller.rb
```diff
@@ -190,5 +194,37 @@ def has_many_reflection?
...
+          .items
+      end
+    end
+
+    def new_join_record
+      field_names = @extra_fields.map { |field| field.name.tr(" ", "_").downcase }
+
+      @resource.fill_record(
+        @reflection.through_reflection.klass.new,
+        additional_params.permit(field_names).merge({source_foreign_key => @attachment_record.id, through_foreign_key => @record.id}),
```
Paul-Bob, line 224:
```suggestion
        additional_params.permit(field_names).merge(
          {
            source_foreign_key => @attachment_record.id,
            through_foreign_key => @record.id
          }
        ),
```

Paul-Bob (changes requested):
Let's rename `extra_fields` to `attach_fields`

binarygit:
@Paul-Bob sorry, I forgot about that. Thank you for reminding. I've renamed and pushed just now. I think it's meeeerrrge timeeeeeee :partying_face:

lib/avo/resources/base.rb
adrianthedev, line 364: Does this work with Ruby 3.1?

adrianthedev (approved):
Nice work! This is going to be such a nice addition!

lib/avo/resources/base.rb
binarygit: Super always passes every argument to the parent method. It's always been that way, I think. Paul also noticed this and had asked: https://github.com/avo-hq/avo/pull/3048#discussion_r1698293700

Paul-Bob (changes requested):
@binarygit we're 99% there! :D

image

When attaching let's ensure that instead of `RELATED_ID` we get the title from the related resource and add a check for it on the test.

Let's also ensure that when the field is searchable the attach fields are rendered. This can be tested on avo pro

lib/avo/resources/base.rb
binarygit: I tested this on ruby:
```
(ins)$ ruby -v
ruby 3.0.3p157 (2021-11-24 revision 3fb7d2cadc) [x86_64-linux]
```
and it works!
```ruby
class Parent
  def hydrate(**args)
    puts args
  end
end

class Child < Parent
  def hydrate(...)
    super
  end
end

Child.new.hydrate(tim: 'cook', steve: 'jobs')
#=> {:tim=>"cook", :steve=>"jobs"}
```

This other article I found says that `...` is exactly like calling super without any arguments: https://blog.saeloun.com/2019/12/04/ruby-2-7-adds-new-operator-for-arguments-forwarding/

image

app/controllers/avo/associations_controller.rb
```diff
@@ -190,5 +194,42 @@ def has_many_reflection?
...
+    def through_reflection?
+      @reflection.instance_of? ActiveRecord::Reflection::ThroughReflection
+    end
+
+    def additional_params?
+      additional_params.keys.count >= 1
+    end
+
+    def additional_params
+      params[:fields].except("related_id")
```
adrianthedev, line 207: Is this ok? Taking all the params that the user gives us?
It feels like a security issue. A bad actor can add fields in the DOM.

Can we just allow the fields that we need and know that the user added from the field DSL?

app/views/avo/associations/new.html.erb
```diff
@@ -22,27 +22,44 @@
...
+                  <%= avo_edit_field :related_id,
+                    as: :select,
+                    form: form,
+                    name: @related_resource.name,
+                    options: options_for_select(@options,
+                    nil),
+                    include_blank: t('avo.choose_an_option'),
+                    stacked: true,
+                    classes: 'w-full'
+                  %>
```
adrianthedev, line 47: This is not properly indented

adrianthedev (changes requested):
I left a comment about security.

Also, should we make the modal say `patron` instead of `user`? The user clicked `Attach patron`.

CleanShot 2024-08-05 at 14 00 03@2x

app/controllers/avo/associations_controller.rb
```diff
@@ -190,5 +194,42 @@ def has_many_reflection?
...
+          .detect_fields
+          .items_holder
+          .items
+      end
+    end
+
+    def new_join_record
+      field_names = @attach_fields.map { |field| field.name.tr(" ", "_").downcase }
+
+      @resource.fill_record(
```
binarygit, line 222: @adrianthedev We use `fill_record` which permits and assigns only those values whose key match the fields declared by the user in their resources file. I think this mitigates the chances of such attacks, no?

https://github.com/avo-hq/avo/blob/90a64f938798049e4749b498400a0c0091623508/lib/avo/resources/base.rb#L448

Paul-Bob: > I think this mitigates the chances of such attacks, no?

Yes, but let's apply the `permit` directly inside the `additional_params` method to increase readability.

```ruby
def additional_params
  params[:fields].permit(@attach_fields.map { |field| field.name.tr(" ", "_").downcase })
end
```

binarygit:
I pushed a commit that turns the texts to:
image

Could you take a look @adrianthedev :smile:

Paul-Bob:
@binarygit we're attaching a patron. The field should be named `Patron` (singular) and not `Patrons` (plural). Let's verify how we compute the title for the "Attach patron" text and apply the same strategy

Paul-Bob (changes requested):
- Some tweaks to improve readability
- Permit params through fields' `id` instead of fields' `name`.

app/controllers/avo/associations_controller.rb
binarygit: okayy. Got it :smile:

app/controllers/avo/associations_controller.rb
```diff
@@ -80,7 +80,7 @@ def create_association
       association_name = BaseResource.valid_association_name(@record, association_from_params)
 
       perform_action_and_record_errors do
-        if through_reflection? && additional_params?
+        if through_reflection? && !additional_params.empty?
```
Paul-Bob, line 83:
```suggestion
        if through_reflection? && additional_params.any?
```

binarygit: There's no `any?` for `ActionControl::Params`. I could do `additional_params.to_h.any?` WDYT?

app/controllers/avo/associations_controller.rb
```diff
@@ -78,7 +80,9 @@ def create_association
       association_name = BaseResource.valid_association_name(@record, association_from_params)
 
       perform_action_and_record_errors do
-        if has_many_reflection?
+        if through_reflection? && additional_params.any?
```
Paul-Bob, line 83:
```suggestion
        if through_reflection? && additional_params.present?
```

app/views/avo/associations/new.html.erb
```diff
@@ -18,31 +18,48 @@
     } do |form| %>
     <%= render Avo::ModalComponent.new do |c| %>
       <% c.with_heading do %>
-        <%= t 'avo.choose_item', item: @related_resource.name.downcase %>
+        <%= t 'avo.choose_item', item: @field.name.downcase %>
```
Paul-Bob, line 21:
```suggestion
        <%= t 'avo.choose_item', item: @field.name.singularize %>
```

Paul-Bob (approved):
🎆

app/controllers/avo/associations_controller.rb
```diff
@@ -190,5 +194,36 @@ def has_many_reflection?
         ActiveRecord::Reflection::HasAndBelongsToManyReflection
       ]
     end
+
+    def through_reflection?
+      @reflection.instance_of? ActiveRecord::Reflection::ThroughReflection
+    end
+
+    def additional_params
+      params[:fields].permit(@attach_fields.map(&:id))
```
adrianthedev, line 203: One more thing.
Let's see if this `permit` statement emits a red warning in the console for other params which are sent by the modal.

I'd like to prevent that behavior. Some users when they have a problem with that actions, they will look at that and think that that is the cause of the problem.
So they are led astray.

app/views/avo/associations/new.html.erb
```diff
@@ -18,31 +18,48 @@
     } do |form| %>
     <%= render Avo::ModalComponent.new do |c| %>
       <% c.with_heading do %>
-        <%= t 'avo.choose_item', item: @related_resource.name.downcase %>
+        <%= t 'avo.choose_item', item: @field.name.singularize %>
```
Paul-Bob, line 21:
```suggestion
        <%= t 'avo.choose_item', item: @field.name.singularize.downcase %>
```

spec/system/avo/for_attribute_spec.rb
```diff
@@ -54,7 +54,7 @@
 
       click_on "Attach even review"
 
-      expect(page).to have_text "Choose review"
+      expect(page).to have_text "Choose even reviews"
```
Paul-Bob, line 57:
```suggestion
      expect(page).to have_text "Choose even review"
```

app/controllers/avo/associations_controller.rb
Paul-Bob, line 203:
```suggestion
      params[:fields].permit(@attach_fields&.map(&:id))
```

app/controllers/avo/associations_controller.rb
```diff
@@ -190,5 +194,36 @@ def has_many_reflection?
         ActiveRecord::Reflection::HasAndBelongsToManyReflection
       ]
     end
+
+    def through_reflection?
+      @reflection.instance_of? ActiveRecord::Reflection::ThroughReflection
+    end
+
+    def additional_params
+      params[:fields].permit(@attach_fields&.map(&:id))
```
Paul-Bob, line 203:
```suggestion
      @additional_params ||= params[:fields].permit(@attach_fields&.map(&:id))
```

PR #3066, opened 2024-07-29
performance: don't fetch fields on action's initialization

### LINKED ISSUE #3025: Check if we can stop running the `fields` method on each resource index page load
### Context

We found out that when loading an Index page, when we fetch and initialize the actions, all the `fields` methods are being executed.
These methods might contain other queries which the user does not want them to be run.

Let's investigate and see if we can stop that behavior.

### System configuration
**Avo version**: 3.x

Paul-Bob:
# Description

Fixes #3025 

Fetch fields on each action request instead of on each action initialization.

# Checklist:


adrianthedev (changes requested):
Can we add some tests for this, please?

PR #3069, opened 2024-07-30
chore: `reflect_on_association` instead `_reflections`

Paul-Bob:
# Description

Fixes # (issue)

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/app/controllers/avo/associations_controller.rb
+++ b/app/controllers/avo/associations_controller.rb
@@ -106,7 +106,7 @@ module Avo
     private
 
     def set_reflection
-      @reflection = @record._reflections.with_indifferent_access[association_from_params]
+      @reflection = @record.class.reflect_on_association(association_from_params)
     end
 
     def set_attachment_class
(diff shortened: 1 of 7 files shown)


app/controllers/avo/associations_controller.rb
```diff
@@ -188,5 +188,9 @@ def join_record
       reflection.through_reflection.klass.find_by(source_foreign_key => @attachment_record.id,
         through_foreign_key => @record.id)
     end
+
+    def has_many_reflection?
+      reflection_class.in? ["HasManyReflection", "HasAndBelongsToManyReflection"]
```
adrianthedev, line 193: I noticed that above some mentions are classes and these are strings.
Is there anything we can do to make them the same or do we have to know when to call as a string and as a class?

adrianthedev (approved):
Looks good! Thanks for the refactor!

app/controllers/avo/associations_controller.rb
```diff
@@ -188,5 +188,9 @@ def join_record
       reflection.through_reflection.klass.find_by(source_foreign_key => @attachment_record.id,
         through_foreign_key => @record.id)
     end
+
+    def has_many_reflection?
```
binarygit, line 192: Noice extracting this method! LGTM!

binarygit (approved):

app/controllers/avo/associations_controller.rb
Paul-Bob: Makes sense. There is no reason to use `class.name.demodulize.to_s` and check against a string.

PR #3097, opened 2024-08-06
fix: false positive on unpermitted parameters

### LINKED ISSUE #3095: Permit params only once per request
Let's ensure the `permit` is called only once on params per request to avoid unpermitted params logs.

Paul-Bob:
# Description

Fixes #3095

## TLDR

This PR ensures that unnecessary and "false positive" warnings about unpermitted parameters are **not** raised by using only the `permit` strong parameters method where we're permitting. We were using the same `permit` method in places where permitting params weren't necessary and we could use `slice` or `extract!`

## Context
When configuring `self.extra_params` on a resource and submitting a form, the `permit` method is utilized in two distinct stages:

1. [In the Base Controller](https://github.com/avo-hq/avo/blob/main/app/controllers/avo/base_controller.rb#L253): All parameters, [including both base and extra params](https://github.com/avo-hq/avo/blob/main/app/controllers/avo/base_controller.rb#L264), are processed through Strong Parameters, ensuring they are permitted.

2. [On the Resource](https://github.com/avo-hq/avo/blob/main/app/controllers/avo/base_controller.rb#L264): When updating the record, `permit` is called again but only for the extra params. This step isolates these extra params from the combined base and extra params hash.

### Actual approach
Instead of using `permit` on `params` to select the extra parameters, we now use `slice`, which returns only the specified extra parameters. In order to use `slice` we need to pass only the keys from `extra_params`. That's safe since the `params` at this point are already permitted (by using also the hashes values from extra_params) and there is no chance to pick unpermitted ones.

### Previous approach

We temporarily disable the action on unpermitted parameters during this second call. This is necessary because `permit` is applied exclusively to the extra params, which means any base params are excluded and would otherwise trigger errors or logs for being unpermitted.

This approach does not impact the rest of the application because the action on unpermitted parameters is only temporarily disabled within this specific context. After handling the extra parameters, we restore the original setting for handling unpermitted parameters.

# Checklist:


binarygit:
Thanks for working on this Paul. I spent a little time trying to understand this and this is tricky. Kudos to you for coming up with the solution. I also have a small idea I want to share:

Could we do something like this?
```ruby
   # base_controller.rb
    def permitted_params
      <current_code>.merge(extra_params)
    end
    
 #base.rb
 def fill_record(record, params, extra_params_keys: [], fields: nil)
         ...
         if extra_params.present?
          # get_key_value returns a hash of objects like permit does...but without permitting
          record.assign_attributes params.get_key_value(extra_params)
        end
```
We'll need to code up the `get_key_value` method (I checked the API and couldn't find anything similar) which essentially replicates permit but without the permitting part. It just takes our array of keys (eg: `[:fish_type, :something_else, {:properties=>[], :information=>[:name, :history], :reviews_attributes=>[:body, :user_id]}]`) and then returns us the key_value pair like permit does.

Do you think something like this would work?

Paul-Bob:
> We'll need to code up the `get_key_value` method

I thought about it but changed my mind after checking the [`permit` source code](https://github.com/rails/rails/blob/19eebf6d33dd15a0172e3ed2481bec57a89a2404/actionpack/lib/action_controller/metal/strong_parameters.rb#L642).

adrianthedev (approved):
To-do:

- let's fill out this PR with what we spoke on the call
- let's PR to Rails to suport `permit_without_warnings`

binarygit (approved):
.

spec/rails_helper.rb
```diff
@@ -49,6 +49,10 @@
 
 # ActiveRecord::Migrator.migrate(File.join(Rails.root, 'db/migrate'))
 
+# Ensure that there are no unpermitted_parameters logs
+ActionController::Parameters.action_on_unpermitted_parameters = :raise
+
+
```
Paul-Bob, line 55:
```suggestion
```

app/controllers/avo/actions_controller.rb
```diff
@@ -64,7 +64,7 @@ def handle
     private
 
     def action_params
-      params.permit(:authenticity_token, :resource_name, :action_id, :button, fields: {})
+      @action_params ||= params.permit(:id, :authenticity_token, :resource_name, :action_id, :button, fields: {})
```
Paul-Bob, line 67: Memoizing and permitting `id`.

The `id` is present when an action is clicked from a show page.

app/controllers/avo/associations_controller.rb
```diff
@@ -132,7 +132,7 @@ def set_reflection_field
     end
 
     def attachment_id
-      params[:related_id] || params.require(:fields).permit(:related_id)[:related_id]
+      params[:related_id] || params.dig(:fields, :related_id)
```
Paul-Bob, line 135: `require` and `permit` is not really necessary here, we just want to access the `related_id`

app/controllers/avo/associations_controller.rb
```diff
@@ -200,7 +200,7 @@ def through_reflection?
     end
 
     def additional_params
-      @additional_params ||= params[:fields].permit(@attach_fields&.map(&:id))
+      @additional_params ||= params[:fields].slice(*@attach_fields&.map(&:id))
```
Paul-Bob, line 203: Applying `permit` only on the `attach_fields` would raise the error of the unpermitted parameters for the remaining parameters.

`permit` is unnecessary here because `fill_record` enures that only params that match `attach_fields` ids will be sent to the record.

spec/system/avo/tags_spec.rb
```diff
@@ -143,8 +143,12 @@
...
       type(:down, :return)
 
-      sleep 0.3
-      click_on "Run"
+      # TODO: fix "found unpermitted parameter: :user_id-dummy"
+      # tags field passes a dummy param generated from a fake input which is always unpermitted
+      with_temporary_class_option(ActionController::Parameters, :action_on_unpermitted_parameters, :log) do
+        sleep 0.3
+        click_on "Run"
+      end
```
Paul-Bob, line 151: Using a tags field will always raise the unpermitted error since it is generating a dummy param that isn't permitted anywhere.

Paul-Bob (commented):
Just added a couple of comments with a brief change explanation

binarygit (approved):
LGTM :smile: . Thanks Paul!

app/views/avo/home/failed_to_load.html.erb
```diff
@@ -2,7 +2,7 @@
   <%
   classes = 'absolute inset-auto left-1/2 top-1/2 transform -translate-x-1/2 -translate-y-1/2'
   label = t 'avo.failed_to_load'
-  src_url = params[:src].present? && !params[:src].starts_with?('http://') ? params.permit(:src).fetch(:src) : nil
+  src_url = params[:src].present? && !params[:src].starts_with?('http://') ? params[:src] : nil
```
adrianthedev, line 5: Let's sanitize the `src_url` below where we display it. There's a security advisory there.

lib/avo/resources/base.rb
```diff
@@ -461,8 +461,12 @@ def fill_record(record, params, extra_params: [], fields: nil)
 
         # Write the user configured extra params to the record
         if extra_params.present?
+          # Pick only the extra params
+          # params at this point are already permited, only need the keys to access them
+          extra_attributes = params.slice(*flatten_keys(extra_params))
```
adrianthedev, line 466: How about we `params` to `permitted_params` so it's clear to anyone who's reading this and wondering why we are slicing instead of permitting.

adrianthedev (approved):
I left a couple of pieces of feedback.

GG! Looking forward to not showing unpermitted params useless to users.

Thanks!

PR #3099, opened 2024-08-06
chore: gh actions' concurrency

Paul-Bob:
# Description

Implements gh actions' [concurrency](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#concurrency) which will stop previous CI actions on each new commit, avoiding creating big workflow queues.


adrianthedev (approved):
Let's do it on `support` as well, please

PR #3110, opened 2024-08-09
performance: sym_link

### LINKED ISSUE #3107: Extract `avo:sym_link` task to bash script
Extracting the symlinking task to a bash script should improve it's performance

Image

Paul-Bob:
# Description

Fixes #3107

_All tests were done on my local machine._

`rails avo:sym_link` was taking around 6 seconds to execute.

With this improvement, it went down to around  3 seconds, where 90% of the execution time is starting the task, not the actual task code.

The execution time of the code inside the task was decreased drastically.

If the code from the task were executed on a bash script it would cut all the time required for starting the task but we need it on the task to enhance on `assets:precompile`.

If from the task we call a script instead the execution time is almost the same, there are no significant improvements.

# Checklist:

### DIFF
--- a/lib/tasks/avo_tasks.rake
+++ b/lib/tasks/avo_tasks.rake
@@ -79,8 +79,10 @@ task "avo:sym_link" do
     `touch #{packages_path}/.keep`
   end
 
+  gem_paths = `bundle list --paths 2>/dev/null`
+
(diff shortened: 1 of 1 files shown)


adrianthedev (approved):
Nice! let's ship it!

PR #3116, opened 2024-08-09
Add ability to specify `default_sort_direction` on a resource

stevegeek:
# Description
The ability to specify a `default_sort_column` was added recently, this PR adds the ability to also specify the direction of the sort.

# Checklist:

## Manual review steps

Simply add a `.default_sort_direction = :asc` to a Resource


stevegeek:
I will need to update the docs when I get a chance

adrianthedev (approved):

adrianthedev:
FYI, these are the docs for sort column
https://github.com/avo-hq/docs.avohq.io/pull/269/files

app/controllers/avo/base_controller.rb
```diff
@@ -325,7 +325,7 @@ def set_index_params
         end
       end
 
-      @index_params[:sort_direction] = params[:sort_direction] || :desc
+      @index_params[:sort_direction] = params[:sort_direction] || @resource.default_sort_direction || :desc
```
Paul-Bob, line 328: Since `class_attribute :default_sort_direction, default: :desc` default is `:desc` the `|| :desc` here is redundant.

```suggestion
      @index_params[:sort_direction] = params[:sort_direction] || @resource.default_sort_direction
```

stevegeek: Good spot!

stevegeek:
Docs added https://github.com/avo-hq/docs.avohq.io/pull/274

Paul-Bob (changes requested):
Feature tests are breaking

stevegeek:
@Paul-Bob ok seems to be related to the change to remove the `|| :desc` , will look at it now

stevegeek:
@Paul-Bob so activesupport `Class.class_attribute` sets the default only on definition of the attribute, ie it does not act as a default when the attribute is specifically set to `nil`.  Im not sure what the desired behaviour here is. My initial implementation meant that if `default_sort_direction` was specifically set to `nil` then the sort direction would default to `:desc` , now in the current implementation, setting `default_sort_direction` to `nil` explicitly removes a sort order parameter.

Paul-Bob:
> `default_sort_direction`

Why would someone set `default_sort_direction = nil`?

stevegeek:
I imagine one would only do so to explicitly force to use the default, or in the case of it being set by some other configuration value say which may end up being `nil`. Hence I would suggest we revert back to the original implementation with `|| :desc`  which is more defensive, setting the sort direction to `:desc` in the event `self.default_sort_direction` is set to nil (on purpose or inadvertently)

spec/features/avo/default_sort_column_and_direction_spec.rb
```diff
@@ -2,15 +2,16 @@
...
+RSpec.describe "DefaultSortColumnAndDirection", type: :feature do
+  context "default_sort_column and default_sort_direction" do
     let(:courses) do
       create_list(:course, 3) do |course, i|
         course.update(created_at: (i + 1).days.ago)
       end
     end
     let(:course_index) { "/admin/resources/courses" }
     let(:default_sort_column) { :created_at }
+    let(:default_sort_direction) { nil }
```
Paul-Bob, line 14:
```suggestion
    let(:default_sort_direction) { Avo::Resources::Base.default_sort_direction }
```

Paul-Bob:
In my opinion, setting `self.default_sort_direction = nil` to enforce an explicit default is not the right approach  and we should not maintain it. The intention behind this line is to explicitly set `default_sort_direction` to `nil`. In some specific edge cases, this behavior might be necessary, and we should preserve the ability to set it to `nil` when required.

I made [this change](https://github.com/avo-hq/avo/pull/3116#discussion_r1713609105), which resolved the test issues.

Thank you for your contribution, @stevegeek! We appreciate it.

stevegeek:
@Paul-Bob ok! You are welcome

Issue #3192, opened 2024-08-24
current_user should not be limited to ActiveRecord::Base class


xeron:
### Describe the bug

Avo 3.11.8 introduced a change which checks if `current_user` has the `ActiveRecord::Base` class. PR https://github.com/avo-hq/avo/pull/3102. This breaks authentication schemes which don't use AR and set `current_user` using custom code (e.g. from HTTP Basic Auth).

https://github.com/avo-hq/avo/blob/3a12d4a72607dad9abddcdaadc7b8303546b9e1b/app/components/avo/sidebar_profile_component.rb#L3-L4

### Steps to Reproduce

Code sample:

`/config/initializers/avo.rb`:

```ruby
Rails.configuration.to_prepare do
  Avo::ApplicationController.include AvoCurrentUser
end
```

`/app/avo/concerns/avo_current_user.rb`:

```ruby
module AvoCurrentUser
  extend ActiveSupport::Concern

  User = Struct.new(:name)

  protected

    def current_user
      user, _pass = ActionController::HttpAuthentication::Basic.user_name_and_password(request)
      User.new(user)
    end
end
```

### Expected behavior & Actual behavior

Custom `current_user` should work as long as it implements required readers (e.g. `name`). However starting with Avo 3.11.8 it fails with a error:

```
13:41:36 web.1  | ActionView::Template::Error (Expected `#<struct AvoCurrentUser::User name=nil>` to be of type: `_Nilable(ActiveRecord::Base)`.):
13:41:36 web.1  |
13:41:36 web.1  | Causes:
13:41:36 web.1  | Literal::TypeError (Expected `#<struct AvoCurrentUser::User name=nil>` to be of type: `_Nilable(ActiveRecord::Base)`.)
13:41:36 web.1  |     32:         <%= render partial: "avo/partials/navbar" %>
13:41:36 web.1  |     33:         <div data-sidebar-target="mainArea" class="content-area flex-1 flex pt-16 relative <%= "sidebar-open" if @sidebar_open %>">
13:41:36 web.1  |     34:           <div class="hidden lg:flex">
13:41:36 web.1  |     35:             <%= render Avo::SidebarComponent.new sidebar_open: @sidebar_open %>
13:41:36 web.1  |     36:           </div>
13:41:36 web.1  |     37:           <div class="flex lg:hidden">
13:41:36 web.1  |     38:             <%= render Avo::SidebarComponent.new sidebar_open: false, for_mobile: true %>
13:41:36 web.1  |
[...]

PR #3195, opened 2024-08-27
fix: remove type inferences from components

adrianthedev:
# Description

Fixes https://github.com/avo-hq/avo/issues/3192

# Checklist:

### DIFF
--- a/app/components/avo/sidebar_profile_component.rb
+++ b/app/components/avo/sidebar_profile_component.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class Avo::SidebarProfileComponent < Avo::BaseComponent
-  prop :user, _Any
+  prop :user, _Nilable(_Any)
 
   def avatar
     if @user.respond_to?(:avatar) && @user.avatar.present?
--- a/app/components/avo/sidebar_profile_component.rb
+++ b/app/components/avo/sidebar_profile_component.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class Avo::SidebarProfileComponent < Avo::BaseComponent
-  prop :user, _Nilable(ActiveRecord::Base)
+  prop :user, _Any
 
   def avatar
     if @user.respond_to?(:avatar) && @user.avatar.present?
--- a/app/components/avo/fields/common/badge_viewer_component.rb
+++ b/app/components/avo/fields/common/badge_viewer_component.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class Avo::Fields::Common::BadgeViewerComponent < Avo::BaseComponent
-  prop :value, String
+  prop :value, prop :value, _Union(_String, _Symbol)
   prop :options, Hash
 
   def after_initialize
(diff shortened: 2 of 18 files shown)


rickychilcott:
For my error (following https://docs.avohq.io/3.0/guides/rest-api-integration.html), I need to make this change to make it pass. I'm happy to make a separate PR, but it feels like it should be together. 

FYI`_Any` does not include `nil` so you _may_ need to do the same to fix https://github.com/avo-hq/avo/issues/3192 for @xeron.

```ruby
class Avo::Views::ResourceEditComponent
   ...
   prop :record, _Nilable(_Any)
   ...
end
```

xeron:
> FYI`_Any` does not include `nil` so you _may_ need to do the same to fix #3192 for @xeron.

I'm not setting `current_user` to `nil`, I'm passing a `Struct`, only `name` attribute is `nil`.

app/components/avo/sidebar_profile_component.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class Avo::SidebarProfileComponent < Avo::BaseComponent
-  prop :user, _Nilable(ActiveRecord::Base)
+  prop :user, _Any
```
Paul-Bob, line 4:
```suggestion
  prop :user, _Nilable(_Any)
```

Paul-Bob (approved):

rickychilcott:
We discovered a second error today. So far the following would fix it for us:

```ruby
class Avo::PanelComponent < Avo::BaseComponent
  ...
  prop :name, _Nilable(Union(_String, _Integer)) do |value|
    value || @args&.dig(:title)
  end
  ...
end

class Avo::Views::ResourceEditComponent < Avo::ResourceComponent
  ...
  prop :resource, _Nilable(_Any)
  ...
end
```

rickychilcott:
Lies... This is what I had to do (so far):

```ruby
# in config/initializers/avo.rb

module OverrideAvoPanelComponent
  extend ActiveSupport::Concern

  included do
    prop :name, _Nilable(_Union(_String, _Integer)) do |value|
      value || @args&.dig(:title)
    end
  end
end

module OverrideAvoPanelNameComponent
  extend ActiveSupport::Concern

  included do
    prop :name, _Nilable(_Union(_String, _Integer))
  end
end

module OverrideAvoResourceEditComponent
  extend ActiveSupport::Concern

  included do
    prop :resource, _Nilable(_Any)
  end
end

module OverrideAvoFieldBadgeViewerComponent
  extend ActiveSupport::Concern

  included do
    prop :value, _Union(_String, _Symbol)
  end
end

Rails.configuration.to_prepare do
  Avo::PanelComponent.include(OverrideAvoPanelComponent)
  Avo::PanelNameComponent.include(OverrideAvoPanelNameComponent)
  Avo::Fields::Common::BadgeViewerComponent.include(OverrideAvoFieldBadgeViewerComponent)
  Avo::Views::ResourceEditComponent.include(OverrideAvoResourceEditComponent)
end
```

adrianthedev:
Thanks @rickychilcott for brainstorming this with us.
I don't quite get why cast the resource as any here `prop :resource, _Nilable(_Any)`?
I mean... we are expecting a resource to be given otherwise it will most likely crash.

PR #3205, opened 2024-09-01
[2.x] Remove last usage of _reflections in favor of reflect_on_association.

simi:
- better compatible with newer Rails, since it casts to needed type on its own

successor to https://github.com/avo-hq/avo/pull/2927

I'm trying to setup Rails 7.2 for Avo 2.x CI at https://github.com/avo-hq/avo/pull/3204 as well, but it will probably need more time.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -110,7 +110,7 @@ jobs:
       id: run_tests
       run: bundle exec rspec spec/features
 
-    - uses: actions/upload-artifact@v2
+    - uses: actions/upload-artifact@v4
       if: always() && steps.run_tests.outcome == 'failure'
       with:
         name: rspec_failed_screenshots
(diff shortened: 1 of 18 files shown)


simi:
I'm not 100% sure the CI failure is related to the change, I have seen same problem in other PR without this change also. :thinking:

Paul-Bob:
Thanks for looking into this.

> I'm not 100% sure the CI failure is related to the change, I have seen same problem in other PR without this change also. 🤔

I'll have a look.

Is this still a draft?

simi:
@Paul-Bob it is draft, since I wasn't able to get passing CI. Code-wise looks ok to me.

Paul-Bob (approved):
Thanks for your work on this, @simi!

I've pushed some commits to fill in the missing pieces. It's all set now, and we'll release it next week.

PR #3208, opened 2024-09-03
fix: resource fails to show when title is numeric

### LINKED ISSUE #3201: Resource fails to show when self.title is numeric
### Describe the bug

After updating to 3.11.10 I noticed **some** of my resources to fail with Internal Server Error 500 on Show action. After investigation I figured it out that it happened only in resources with self.title set as a number (including numeric id). Changing this is not a problem, but investigating what  happened was problematic.

### Steps to Reproduce
Steps to reproduce the behaviour:
1. Create any resource
2. Set self.title to point to a method returning numeric value
3. Create `def fields` method with at least one field
4. Open Show page of that resource
5. Something Went Wrong

### Expected behaviour & Actual behaviour

Return numeric id as a string or a nice error and a note in documentation that is has to be a string
vs 
Something Went Wrong page and 500 error

### System configuration
**Avo version**:
3.11.10

**Rails version**:
7.2

**Ruby version**:
3.2.2

**License type**:

 - [ ] Community
 - [x] Pro
 - [ ] Advanced

 - [x] No

#### Impact

 - [x] High impact (It made my app broken - until I figured out the problem source)

#### Urgency

 - [x] Low urgency (It can wait. I just wanted you to know about it as others may have same issue)

Paul-Bob:
# Description

Fixes #3201 

Enforce `String` type on `record_title` method

# Checklist:

### DIFF
--- a/lib/avo/resources/base.rb
+++ b/lib/avo/resources/base.rb
@@ -395,21 +395,22 @@ module Avo
       end
 
       def record_title
-        if @record.nil?
-          name
-        elsif title.nil?
-          # Get the title from the record if title is not set, try to get the name, title or label, or fallback to the to_param
-          @record.try(:name) || @record.try(:title) || @record.try(:label) || @record.to_param
+        fetch_record_title.to_s
+      end
 
-        else
-          # If the title is a symbol, get the value from the record else execute the block/string
-          case title
-          when Symbol
-            @record.send title
-          when Proc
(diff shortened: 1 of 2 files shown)


adrianthedev (changes requested):
Let's not do nested if's

```ruby
def record_title
  fetch_record_title.to_s
end

private

def fetch_record_title
  ce avem acum
end

```

PR #3246, opened 2024-09-14
chore: remove unwanted development deps

adrianthedev:
# Description

After a quick debugging session with @segiddins, we found a few things that shouldn't be loaded in the app, thus making the boot process faster.


Paul-Bob:
Apply this changes on other packages

PR #3292, opened 2024-10-01
feature: implicit authorization

### LINKED ISSUE #2125: Implicit authorization
This disables everything (access to the resource/search/attachments/uploads, everything) until someone implicitly enables them.

Maybe we should introduce a roles system that will handle this.

From https://github.com/avo-hq/avo/discussions/1574

Paul-Bob:
# Description

Fixes #2125
[avo-pro PR](https://github.com/avo-hq/avo-pro/pull/90)

- ~Add `raise_error_on_missing_policy_method` that behaves the same as the existing `raise_error_on_missing_policy` but for methods.~
- Add `whitelisting_authorization` that is false per default (true on new apps by template). This option enables a more strict authorization control where the action is granted only if is explicitly defined. If policy class is missing OR if method for action is missing, action will be unpermitted.

TODO:
- [x] Docs
- [x] Tests

# Checklist:

  - https://github.com/avo-hq/docs.avohq.io/pull/294

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -107,7 +107,7 @@ GIT
 PATH
   remote: .
   specs:
-    avo (3.13.0)
+    avo (3.13.1)
       actionview (>= 6.1)
       active_link_to
       activerecord (>= 6.1)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -642,7 +642,7 @@ GEM
     unicode-display_width (2.6.0)
     uri (0.13.1)
     useragent (0.16.10)
-    view_component (3.14.0)
+    view_component (3.15.1)
       activesupport (>= 5.2.0, < 8.0)
       concurrent-ruby (~> 1.0)
       method_source (~> 1.0)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -107,7 +107,7 @@ GIT
 PATH
   remote: .
   specs:
-    avo (3.13.1)
+    avo (3.13.3)
       actionview (>= 6.1)
       active_link_to
       activerecord (>= 6.1)
--- a/app/components/avo/index/resource_table_component.rb
+++ b/app/components/avo/index/resource_table_component.rb
@@ -14,7 +14,7 @@ class Avo::Index::ResourceTableComponent < Avo::BaseComponent
   prop :parent_record, _Nilable(_Any)
   prop :parent_resource, _Nilable(Avo::BaseResource)
   prop :pagy, _Nilable(Pagy)
-  prop :query, _Nilable(ActiveRecord::Relation)
+  prop :query, _Nilable(_Any)
   prop :actions, _Nilable(_Array(Avo::BaseAction))
 
   def encrypted_query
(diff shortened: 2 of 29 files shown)


adrianthedev (changes requested):
Great job!
I can't wait to have this merged!

Some feedback and thoughts.

- I would remove the `raise_error_on_missing_policy_method` as we want most apps to use the new default with whitelisting.
- rename the new config from `whitelisting_authorization` to `implicit_authorization`?
- when `implicit_authorization` is set to true, then `raise_error_on_missing_policy` becomes obsolete and the result would be `false`. What do you think about this approach? I would advise our users to set the default `ApplicationPolicy` to have all methods return `false` and then just override the new ones.

This all goes towards having new apps with the new `implicit_authorization` option turned on and slowly have it the default in Avo 4.
So this code will be much more simplified when we do the switch.
What do you think?

Paul-Bob:
**I think we're looking at `raise_error_on_missing_policy_method` and `raise_error_on_missing_policy` from a different perspective. I'm looking at those not to protect the app on production but to help during development and those definitively helped me while testing. IMHO letting each user configure them as they want would serve both perspectives.**

#

> I would remove the `raise_error_on_missing_policy_method` as we want most apps to use the new default with whitelisting.

I used `raise_error_on_missing_policy_method` on the dummy app on development and was super useful to understand what methods I'm missing from certain policy files (like `act_on?`), otherwise, those methods were returning `false` without me knowing about them. IMO `raise_error_on_missing_policy_method` is a great development help. ( Can't see any value on `production` )

#

> rename the new config from `whitelisting_authorization` to `implicit_authorization`?

Yes, it's easy to write/spell.

#

> when `implicit_authorization` is set to true, then `raise_error_on_missing_policy` becomes obsolete and the result would be `false`.

I've also used both (`implicit_authorization` and `raise_error_on_missing_policy`) on `true` simultaneously on development to figure out what policy files I'm missing. It's a great feature to have configured on `true` (on development) when you don't want to forget to create a policy when you create a new resource.

# 

The idea of overriding `raise_error_on_missing_policy` when `implicit_authorization` is set to true makes total sense on `prduction` environment since the app is already protected and the error would only make it break without any added value. 

If we look from a production environment perspective `raise_error_on_missing_policy_method` and `raise_error_on_missing_policy` are not helpful when using `implicit_authorization`. But for `development` env those are great whether `implicit_authorization` is used or not.

#

> I would advise our users to set the default `ApplicationPolicy` to have all methods return `false` and then just override the new ones.

If `implicit_authorization` is set on `true`  I think there is no need to keep any method returning `false` on the `ApplicationPolicy` since by default on the missing method the authorization will already return `false`. With `implicit_authorization` it makes more sense to add methods that you want to always return `true` on the `ApplicationPolicy`

adrianthedev:
Ok. I see what you mean to use it as a development helper.
Ok. Let's have both `missing` methods then that run only in development.

adrianthedev (changes requested):
Let's add those methods.

adrianthedev (approved):

Issue #3245, opened 2024-09-14
Remove `OpenStruct` from the codebase


adrianthedev:
### Context

`OpenStruct` seems to be "deprecated" and not that performant.
Let's try to replace it with `Data` (carefull. it's not mutable).

Yash-Singh-Pathania:
hi @adrianthedev  would you be okay assigning this to me .

adrianthedev:
Done!  Thanks @Yash-Singh-Pathania!

adrianthedev:
Maybe this is a good alternative?
https://api.rubyonrails.org/classes/ActiveSupport/OrderedOptions.html

Yash-Singh-Pathania:
@adrianthedev Makes sense. I'll try to implement it and have the pull request ready by today or the day after. Thanks for the help!

Yash-Singh-Pathania:
@adrianthedev : https://github.com/avo-hq/avo/actions/runs/11178246617/job/31075442563?pr=3307
could you please help me with the test cases ?

PR #3307, opened 2024-10-04
refactor: replace OpenStruct with Data for performance and readability improvement

Yash-Singh-Pathania:
# Description

This PR refactors the codebase by replacing OpenStruct with Ordered Options for improved performance and immutability. The change ensures that Data is used to represent immutable objects, which provides better memory usage and execution speed compared to OpenStruct. No additional dependencies are required for this change.

Fixes # (issue)
[#3245](https://github.com/avo-hq/avo/issues/3245)

# Checklist:

## Screenshots & recording

N/A (No UI changes involved)

## Manual review steps

1. Verify that the `accounts` method returns the correct objects using the new `Data` class.
2. Ensure that `perform_request` correctly handles the refactored `Response` class in test mode.


docs.avohq.io
adrianthedev, line 1: I dont' think this should be here.
This is why the CI is failing.

Yash-Singh-Pathania:
@adrianthedev  can you please  review this , I think its done sorry for the delay I just got started with my masters 
thanks a ton.

Yash-Singh-Pathania:
@Paul-Bob  would you be reviewing this ?

Paul-Bob (changes requested):
It feels unnatural to create a new class and override the initializer; it seems like we're not using the right tool for the job.

I came across this article discussing alternatives to `OpenStruct`:
https://dev.to/lorankloeze/alternative-for-rubys-openstruct-i72

I think we should consider using a more appropriate tool, such as `Struct` or `Data`.

`Struct` is significantly faster than `OpenStruct` (~80 times faster), although I’m unsure how it compares to `Data`. Let’s either research or run a quick benchmark, similar to the one in the article, to determine which option `Struct` or `Data` is faster, and proceed with that.

Yash-Singh-Pathania:
@Paul-Bob  let me go through this quickly and run a quick benchmark for the same.

spec/dummy/app/models/user.rb
```diff
@@ -19,7 +19,10 @@
 #  active                 :boolean          default(TRUE)
 #  slug                   :string
 #
+
```
Paul-Bob, line 22:
```suggestion
```

Paul-Bob (approved):
Thank you for working on this @Yash-Singh-Pathania 

I've completed the PR by using `Struct` instead of previous approaches.

PR #3315, opened 2024-10-08
fix: undefined tags in field when editing

coatezy:
# Description

This PR fixes an undefined tag issue when editing existing tags on a resource.

# Checklist:

## Screenshots & recording

**Before:**

CleanShot 2024-10-08 at 13 12 48@2x
CleanShot 2024-10-08 at 13 00 44@2x

**After:**
CleanShot 2024-10-08 at 13 13 08@2x

## Manual review steps

1. Step 1
1. Step 2


Paul-Bob:
Hey @coatezy thanks for the contribution! I noticed that some tests are failing. Could you please share the tags field code from your resource and more details about how to reproduce the issue?

coatezy:
> Hey @coatezy thanks for the contribution! I noticed that some tests are failing. Could you please share the tags field code from your resource and more details about how to reproduce the issue?

Hey, @Paul-Bob! Not problem. I'd be happy to try and understand why these are failing. The tags field is configured in the following way on the resource.

```
    field :tags, as: :tags, acts_as_taggable_on: :tags
```

Paul-Bob:
I was able to reproduce the issue, I guess the tests are failing because tags cover several use cases and this change fixes this particular use case but maybe is breaking other use case, I'll also have a look

Paul-Bob (changes requested):
I suggested some changes that might be effective.

When using `suggestions` https://docs.avohq.io/3.0/fields/tags.html#possible-values as an array of hashes the `possibleSuggestion` will respond to `label` but when used as array of values or not used at all it will NOT respond to `label`

Paul-Bob (approved):
Thank you for this contribution @coatezy!

Some tests were failing, so I made a change to ensure that tags function correctly in other scenarios.

PR #3320, opened 2024-10-09
✨ Controls placement on resource level

RocKhalil:
# Description
You can now have the placement of the actions set on the resource too. So you can have a resource with controls on the left and another one with controls on the right.

# Checklist:

## Manual review steps

1. On a resource level, add `self.controls_placement = :left`
2. View the controls being rendered on the `left` instead of the default `right` value

### DIFF
--- a/app/components/avo/index/table_row_component.html.erb
+++ b/app/components/avo/index/table_row_component.html.erb
@@ -1,7 +1,4 @@
 <%# hover:z-[21] removed from tr class to solve flickering actions component on row controls and z-20 changed to z-21%>
-<%
-  resource_controls_placement = @resource.controls_placement || Avo.configuration.resource_controls_placement
-%>
 
 <%= content_tag :tr,
   class: class_names("bg-white hover:bg-gray-50 hover:shadow-row z-21 border-b", {"cursor-pointer": click_row_to_view_record}),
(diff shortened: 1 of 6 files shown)


app/components/avo/index/table_row_component.html.erb
```diff
@@ -1,4 +1,8 @@
 <%# hover:z-[21] removed from tr class to solve flickering actions component on row controls and z-20 changed to z-21%>
+<%
+  resource_controls_placement = @resource.controls_placement || Avo.configuration.resource_controls_placement
+%>
```
Paul-Bob, line 4: Let's extract this logic to a concern, something like

```
def controls_placement
  @resource.controls_placement || Avo.configuration.resource_controls_placement
end

def resource_controls_on_the_right?
  controls_placement == :right
end

def resource_controls_on_the_left?
  controls_placement == :left
end
```

Then include the concern on the components that uses this logic.

Let's also ensure that the original methods `resource_controls_on_the_right?` and `resource_controls_on_the_left?` are deleted if they are no longer used.

Paul-Bob (changes requested):
Thank you for this contribution @RocKhalil!

We should wrap this logic on a concern and remove unused methods, otherwise the logic is done well!

RocKhalil:
@Paul-Bob refactored into a concern + added `both` so we can render the controls on both sides if someone wants it

Paul-Bob (approved):
It's looking good @RocKhalil!

I noticed that you added the `:both` option and I can see its value.

I'm approving the PR but there is still some work to do to pass this to the finish line:

1. Add docs for the `:both` option, including the version in which it was introduced.
2. Add tests

Feel free to let me know if you need any help.

RocKhalil:
documentation PR: https://github.com/avo-hq/docs.avohq.io/pull/296

PR #3325, opened 2024-10-10
Made TestHelpers include methods from WaitForLoaded

iainbeeston:
# Description

Right now many of the test helpers don't work in any other project because they rely on methods from WaitForLoaded and the test helpers crash when they are called. This isn't an issue in the avo project itself because the WaitForLoaded methods are required automatically by the avo test suite. It isn't possible to load WaitForLoaded in projects using the rubygem of avo because files inside the spec directory aren't included (and even if they were it would be a bit weird to load something from `spec/support` in a gem).

To fix this I've moved WaitForLoaded alongside the TestHelpers module and included it from there so the helper methods can be used in other projects.

Fixes # (issue)

# Checklist:

## Manual review steps

1. Confirm avo test suite still works correctly
1. Check test helpers (eg. add_tag) work in a project using avo


iainbeeston:
I haven't fixed any of the code climate comments because I didn't want to change that code.

Paul-Bob (approved):
Thank you @iainbeeston!

PR #3328, opened 2024-10-10
Made the send_keys helper work in selenium as well as cuprite

iainbeeston:
# Description

A few of the test helpers use the `type` helper method but in selenium it raises this error:

     NoMethodError:
       undefined method `keyboard' for #<Selenium::WebDriver::Chrome::Driver:0x000000013ebbb950>

I've changed this to do the same thing (type in the focused element) but in a way that should work in every capybara driver.

Fixes # (issue)

# Checklist:

## Manual review steps

1. Tests using `add_tag` and other helpers that use the keyboard should still pass.

### DIFF
--- a/lib/avo/test_helpers.rb
+++ b/lib/avo/test_helpers.rb
@@ -315,7 +315,11 @@ module Avo
     end
 
     def type(...)
-      page.driver.browser.keyboard.type(...)
+      if page.driver.browser.respond_to?(:keyboard)
+        page.driver.browser.keyboard.type(...)
+      else
+        page.send_keys(...)
+      end
     end
 
     def accept_custom_alert(&block)


Paul-Bob (changes requested):
Hey @iainbeeston thank you for looking into this.

It seems that `page.send_keys(...)` is breaking our tests on cuprite.

Not sure if we can make a method that is 100% compatible with both `selenium` and `cuprite`, can you have a look into it?

If it's blocking, an alternative approach would be to override it on your side or to make 2 methods:

```ruby
def cuprite_type(...)
  page.driver.browser.keyboard.type(...)
end

def selenium_type(...)
  page.send_keys(...)
end
```

Paul-Bob (approved):

Paul-Bob:
Thank you for this contribution @iainbeeston

Issue #3334, opened 2024-10-11
Exception is raised when association uses a composite primary/foreign key


filipegiusti:
### Describe the bug

`foreign_key.to_sym` on https://github.com/avo-hq/avo/blob/v3.13.5/lib/avo/fields/belongs_to_field.rb#L197 raises `NoMethodError: undefined method `to_sym' for an instance of Array` when an association uses a composite primary key/foreign key.

### Steps to Reproduce

Have a relationship that uses composite primary/foreign keys.

```ruby
class Membership
  belongs_to :profile, primary_key: %i[organization_id id], query_constraints: %i[organization_id profile_id]
end

class Profile
  has_one :membership, primary_key: %i[organization_id id], query_constraints: %i[organization_id profile_id]
end
```

Try to update any attribute on Profile through Avo.

### Expected behavior & Actual behavior

Expected: Update works.
Actual: Exception it raised with the following backtrace.

```
NoMethodError: undefined method `to_sym' for an instance of Array
/webapp/vendor/bundle/ruby/3.3.0/gems/avo-3.13.5/lib/avo/fields/belongs_to_field.rb:197:in `to_permitted_param': undefined method `to_sym' for an instance of Array (NoMethodError)
	from /webapp/vendor/bundle/ruby/3.3.0/gems/avo-3.13.5/app/controllers/avo/base_controller.rb:277:in `map'
	from /webapp/vendor/bundle/ruby/3.3.0/gems/avo-3.13.5/app/controllers/avo/base_controller.rb:277:in `permitted_params'
	from /webapp/vendor/bundle/ruby/3.3.0/gems/avo-3.13.5/app/controllers/avo/base_controller.rb:266:in `model_params'
	from /webapp/vendor/bundle/ruby/3.3.0/gems/avo-3.13.5/app/controllers/avo/application_controller.rb:200:in `fill_record'
	from /webapp/vendor/bundle/ruby/3.3.0/gems/activesupport-7.2.1/lib/active_support/callbacks.rb:362:in `block in make_lambda'
	from /webapp/vendor/bundle/ruby/3.3.0/gems/activesupport-7.2.1/lib/active_support/callbacks.rb:179:in `block in call'
	from /webapp/vendor/bundle/ruby/3.3.0/gems/actionpack-7.2.1/lib/abstract_controller/callbacks.rb:34:in `block (2 levels) in <module:Callbacks>'
[...]

PR #3335, opened 2024-10-11
Fix exception when using a resource with a composite primary/foreign key association

filipegiusti:
Fixes https://github.com/avo-hq/avo/issues/3334

Fix `NoMethodError: undefined method 'to_sym' for an instance of Array` raised by this line when an association uses a composite primary key/foreign key.

# Checklist:


adrianthedev:
Thanks @filipegiusti!
Would you be able to add some tests for this as well?

The approach I would take is to create a new model & resource which uses composite foreign keys and test that.

Thank you for tackling this!

filipegiusti:
> The approach I would take is to create a new model & resource which uses composite foreign keys and test that.

I searched for a test file at `spec/lib/avo/fields/belongs_to_field.rb` or using the search term `spec fields belongs_to` but I couldn't find anything. I'm not sure where such test would be and how it would look like, I don't have much experience testing libraries.

adrianthedev:
Can you create a new spec next to [this one](https://github.com/avo-hq/avo/blob/main/spec/features/avo/belongs_to_spec.rb#L3) and test only the composite behavior?

Paul-Bob:
@filipegiusti I can take over and create the test. Could you share some details about your environment setup? We don’t have any tables with composite primary or foreign keys, so I’m unsure how to configure the environment to test this change.

filipegiusti:
It all started when we added the primary_key and foreign_key. We have these models.

```ruby
class Profile
  has_one :membership, primary_key: %i[organization_id id], foreign_key: %i[organization_id profile_id], dependent: :destroy, inverse_of: :profile
end

class Membership
  belongs_to :profile, primary_key: %i[organization_id id], foreign_key: %i[organization_id profile_id], inverse_of: :membership
end
```
Both models have a primary key of id and an unique index of `[organization_id, id]`.

The edit page of Profile works, the exception is raised when Profile is attempted to be updated.

We also have:

```ruby
class Resources::Profile < Avo::BaseResource
  def fields
    field :membership, as: :belongs_to, only_on: :show
  end
end
```

Paul-Bob:
> I might be able to look at this next week.

That would be great,

Thank you!

filipegiusti:
Hello! This was prioritized and I'll try to make this work in Avo.

I see there are test models for Location and TeamMembership, could I modify and add a relationship between these? It would be something like:

```ruby
# in a migration

add_column :team_memberships, :location_id, :bigint
add_foreign_key :team_memberships, :locations, primary_key: %i[team_id id], column: %i[team_id location_id] # introduced on https://github.com/rails/rails/pull/47637

# in models

class Location
  has_many :team_memberships, primary_key: %i[team_id id], foreign_key: %i[team_id location_id], inverse_of: :location
end

class TeamMembership
  belongs_to :location, primary_key: %i[team_id id], foreign_key: %i[team_id location_id], inverse_of: :team_memberships
end
```

This enforces a TeamMembership to only be linked to Location that shares the same Team.

With that I should be able to create a test case that fails with the current code.

Paul-Bob:
> could I modify and add a relationship between these?

If it doesn't affect current tests you can, but I think there are some tests around those models / associations so the safest approach would be to set up a new association between existing models or creating new ones for this use case

filipegiusti:
I had a hard time  creating the tests so I solved the problem in another way. We don't have a need to support composite foreign keys in associations anymore.

Paul-Bob:
Thanks for letting us know @filipegiusti 

I'm closing this PR to keep the queue clean, we might get back to it and reopen if necessary.

PR #3343, opened 2024-10-16
Avoid index_active_storage_attachments_uniqueness violation by attaching multiple files in a single call to attach

jpawlyn:
# Description
We recently tried to switch to using `direct_upload: true` on the `Files` field. However when creating a new record with multiple files, we got the error `duplicate key value violates unique constraint "index_active_storage_attachments_uniqueness"`.

The fix for this is to attach files in a single call to `attach` instead of calling `attach` for each file.

Fixes # (issue)

# Checklist:

## Screenshots & recording

index-violation

## Manual review steps

1. Step 1
1. Step 2


camilodelvasto (approved):
💯

lib/avo/fields/files_field.rb
```diff
@@ -32,12 +32,9 @@ def to_permitted_param
...
 
-        value.each do |file|
-          # Skip empty values
-          next unless file.present?
-
-          record.send(key).attach file
-        end
+        # attach files in one call to avoid index_active_storage_attachments_uniqueness violation
+        files = value.compact_blank
+        record.send(key).attach(files) unless files.none?
```
jpawlyn, line 37: On second thoughts, this looks better. Let me just check that it's the same.
```suggestion
        record.send(key).attach(value.compact_blank)
```

adrianthedev:
@camilodelvasto can we add a test to check for the bad behavior that was happening before? Also to check for `nil` values (to see that the `compact_blank`) works.

jpawlyn:
> Looking great! Thank you @jpawlyn for the contribution!
> 
> Good to merge after adding some tests

I haven't been able to re-create the issue in the Avo test suite I'm afraid but I have recreated it in the Active Storage test suite.

Looking at https://github.com/rails/rails/blob/ee661852313ff697d5f3a73a68aab99101ea3ead/activestorage/test/models/attached/many_test.rb#L404-L415 the call to `attach` is taking multiple signed ids. When I change that to 2 separate calls to `attach` ie

jpawlyn:
> > @Paul-Bob I tried writing a test but unfortunately it doesn't fail with the original code. Somehow my app running in production and the activestorage test suite re-creates the issue but not for the Avo test suite.
> 
> Does it fail on Avo local?

Yes it does

Paul-Bob:
> When running my app locally, yes it does fail

Ok, thank you for your effort on this @jpawlyn. I'll have a look

Issue #3263, opened 2024-09-24
Show trix / rhino always_show variation


pjmuller:
### Feature
Show a part of the content with a click to see more
See loom: https://www.loom.com/share/d95cfc272bb84e1797870265c9a710fb

to be discussed how we introduce this without creating regressions for users  that really want to hide everything.
introduce an extra boolean? `partial_show` ?

### Current workarounds
Define a custom partial for the show action -> copying the app/components/avo/fields/trix_field/show_component.html.erb and using our own styles?

or overwriting the `hidden` CSS class within this wrapper?

pjmuller:
FYI: we 80/20 fixed this withi this css, #sharing-is-caring ;)
This works for both trix as rhino

```css
.trix-content.hidden {
  display: block !important;
}

.trix-content.hidden {
  clip-path: inset(0 0 calc(100% - 50px) 0);
  -webkit-mask-image: linear-gradient(to bottom, black 30px, transparent 50px);
  mask-image: linear-gradient(to bottom, black 30px, transparent 50px);
  max-height: 50px;
  min-height: 0px;
}

.trix-content:not(.hidden) {
  clip-path: none;
  -webkit-mask-image: none;
  mask-image: none;
}

a[data-action="click->hidden-input#showContent"].hidden {
  display: none !important;
}
```

adrianthedev:
Yup!
Awesome recommendation!

We should have a few junior folks that want to contribute to OSS soon so maybe we can get some help on it.


### Approach

- tweak the field component `Avo::Fields::TrixField::ShowComponent` to show a bit of the body
- this body would be in a div
- the div would have overflow: hidden and a fixed height
- create a new stimulus controller to remove the overflow and height classes

**Nice to have:** 
- the bottom part of the div would have a gradient that passes from transparent (top) to white (bottom) so the user peeks that there is more content below (see how we did it on [this PR](https://github.com/avo-hq/avo/pull/3280))

adrianthedev:
#4724 added Avo::Fields::Common::CollapsableContentComponent + collapsible-content.css: 50px faded preview, More/Less, always_show opts out

PR #3359, opened 2024-10-24
Refactor : change trix field show

Nevelito:
# Description
- Change trix CSS
- change show_component for trix field
- add new js controller for trix body
- add new locales in every language

Fixes # (issue)
https://github.com/avo-hq/avo/issues/3263

# Checklist:

## Screenshots & recording
image
image


KamilMilewski (approved):

app/components/avo/fields/trix_field/show_component.html.erb
```diff
@@ -1,14 +1,13 @@
...
+  <div data-controller="trix-body">
+    <div class="trix-content py-2 max-w-4xl hidden" data-trix-body-target="content">
       <%= sanitize @field.value.to_s %>
     </div>
+    <div class="hidden" data-trix-body-target="moreContentButton">
+      <%= link_to t('avo.more_content'), 'javascript:void(0);', class: 'font-bold inline-block pt-3', data: { action: 'click->trix-body#toggleContent' } %>
+    </div>
+    <div class="hidden" data-trix-body-target="lessContentButton">
+      <%= link_to t('avo.less_content'), 'javascript:void(0);', class: 'font-bold inline-block pt-3', data: { action: 'click->trix-body#toggleContent' } %>
+    </div>
```
Paul-Bob, line 11: `@field.always_show` seems to be ignored after these changes.

`always_show` is an option for the Trix field, it is documented here https://docs.avohq.io/3.0/fields/trix.html#always_show

We should ensure that setting `always_show` to true bypasses the entire show/hide content logic.

app/javascript/js/controllers/trix_body_controller.js
```diff
@@ -0,0 +1,27 @@
...
+  toggleContent() {
+    this.contentTarget.classList.toggle('hidden')
+    this.toggleButtons()
+  }
+
+  toggleButtons() {
+    this.moreContentButtonTarget.classList.toggle('hidden')
+    this.lessContentButtonTarget.classList.toggle('hidden')
+  }
+}
```
Paul-Bob, line 27: Clean, nice work here!

spec/system/avo/trix_field_spec.rb
```diff
@@ -31,11 +31,17 @@
 
         save
 
-        click_on "Show content"
-
```
Paul-Bob, line 35: In all instances where `click_on "Show content"` was removed, update the test logic to use `click_on "More content"` instead.

Since "More content" only appears when there is additional content, add extra text to `trix_post_body`.

It's essential to keep each test case focused on its original purpose. Here, we only need to add more content, and the test should confirm that the entire content displays correctly after clicking "More content."

spec/system/avo/trix_field_spec.rb
```diff
@@ -46,10 +52,33 @@
...
+          expect(page).to have_link("More content", href: "javascript:void(0);")
+        end
+
+        it "displays correct button after extended content" do
+          visit "/admin/resources/posts/#{post.id}"
+
+          click_on "More content"
+
+          expect(page).to have_link("Less content", href: "javascript:void(0);")
+        end
```
Paul-Bob, line 80: It's possible to speed up the tests by joining those, performance and speed on tests are always great!

Paul-Bob (changes requested):
Hey @Nevelito thank you for this contribution!

I left some comments on specific parts of the code, feel free to continue a conversation on each comment.

Additionally, I noticed that some locale files were generated under the `dummy/config/locales` path. Since these won't be used during development, we can remove them to keep the repo clean.

The PR is looking good overall! Let's focus on the tests and on ensuring that `always_show` option works as expected. We should also add a test case for `always_show`

Paul-Bob:
Just one more thing, Lint action is failing, you can check what is failing by clicking on "Details"

image

spec/system/avo/trix_field_spec.rb
Nevelito: the js controller checks if the body has more then one line, so if it is not then More content button will not appear

Paul-Bob (approved):
Thanks for this contribution @Nevelito it looks great!

I've added a test for the `always_show` field option.

Issue #3368, opened 2024-10-29
Non-standard record IDs can cause a record not found error


davidlormor:
### Describe the bug

Avo can't handle non-standard primary keys/IDs that don't follow Rails' out-of-the-box conventions (e.g. periods in a string-based ID). Rails documents how to work around this via constraints in their guides. Unfortunately, it appears that segment-level constraints cannot be passed down to routes in a mounted engine.

### Steps to Reproduce

1. Create a new model with a string-based ID (use a generator to create the Avo resource, too):
```bash
bin/rails g model book code:string:uniq title:string
```
2. Edit the migration to use a custom primary key
```ruby
class CreateBooks < ActiveRecord::Migration[7.0]
  def change
    create_table :books, id: false, primary_key: :code do |t|
      t.string :code, null: false, index: { unique: true }
      t.string :title

      t.timestamps
    end
  end
end
```
3. Run the migration to set up the books table
```bash
bin/rails db:migrate
```
4. Tell the model about the primary key:
```ruby
class Book < ApplicationRecord
  self.primary_key = "code"
end
```
5. Create a book with an unsupported ID param via Rails console (or Avo):
```bash
Book.create(code: '11-2003.34', title: 'Some book')
```
6. Navigate to the book record in Avo

### Expected behavior & Actual behavior

I would expect the book to be rendered via Avo, but instead I get a 500 page. Inspecting the Rails server output, it appears that the book cannot be found, due to an incorrectly parsed ID param. The request params are logged as `params: { id: "11-2003" }` (note the missing content starting at the `.` character).

According to the Rails guides, this issue can be [solved via a constraint](https://guides.rubyonrails.org/routing.html#dynamic-segments):
```ruby
Rails.application.routes.draw do
  mount Avo::Engine, at: Avo.configuration.root_path, constraints: { id: /[*\/]+/ }
end
```
Unfortunately, it appears that because the constraint applies at the route segment level, it is not passed to the Rails engine (I've confirmed this on a separate example engine). I've logged a question on the Rails discussion board to see if anything can/should be done at the Rails level, since this affects all engines, not just Avo.

### Models and resource files

See above for setup.

### System configuration
**Avo version**: 3.13.7

**Rails version**: 8.1.0.alpha

**Ruby version**: 3.3.5

**License type**:

 - [x] Community
 - [x] Pro
 - [x] Advanced

**Are you using Avo monkey patches, overriding views or view components?**

 - [ ] Yes. If so, please post code samples.
[...]

PR #3369, opened 2024-10-29
Fix issue with loading dynamic routes

davidlormor:
# Description

Per the [Rails guides](https://guides.rubyonrails.org/routing.html#breaking-up-very-large-route-file-into-multiple-small-ones for details), you should not call `Avo::Engine.routes.draw` again when loading routes from another file.

This is related to #3368. Doesn't actually fix the issue, but this change is required in order to ensure that an `:id` constraint can be applied to all resource-related routes.

Based on the Rails guides, I would suggest the following change, but didn't want to do anything too drastic:

```ruby
# config/routes.rb
Avo::Engine.routes.draw.do
  # ...other routes
  scope "resources", as: "resources" do
    draw :resources
  end
end
```

```ruby
# config/routes/resources.rb
Avo::Resources::ResourceManager.fetch_resources.map do |resource|
  resources resource.route_key do
    member do
      get :preview
    end
  end
end
```

If you want to take it even a step further, the Rails guides suggests only breaking up your routes file when it goes extremely large and complex, so an even simpler version would be:

```ruby
# config/routes.rb
Avo::Engine.routes.draw.do
  # ...other routes
  scope "resources", as: "resources" do
    Avo::Resources::ResourceManager.fetch_resources.map do |resource|
      resources resource.route_key do
        member do
          get :preview
        end
      end
    end
  end
end
```

Either method would ensure that any potential fix to #3368 would be able to properly apply a routing constraint to all resource routes.

# Checklist:

## Manual review steps

1. Referenced this change in an existing Avo implementation and confirmed that all resource routes are still working as expected.


Paul-Bob (changes requested):
This approach isn't generating the routes as expected.

In my opinion, `Avo::DynamicRouter` no longer needs this level of abstraction. Previously, it had more code, making the extraction useful, but now it’s compact enough to apply directly in `config/routes.rb` instead of a separate file.

Let's remove the `lib/avo/dynamic_router.rb`  and directly apply the code where `Avo::DynamicRouter.routes` was called.

Replace
```ruby
    # Generate resource routes as below:
    # resources :posts
    Avo::DynamicRouter.routes
```

With
```ruby
    # Generate resource routes as below:
    # resources :posts
    Avo::Resources::ResourceManager.fetch_resources.map do |resource|
      resources resource.route_key do
        member do
          get :preview
        end
      end
    end
```

davidlormor:
🙌 no problem @Paul-Bob - happy to help get that cleaned up! Especially now that I've realized my challenges with ID constraints (#3368) are a bit more broad than I'd originally realized. Turns out the proprietary data set that I'm working with has several tables with composite keys in addition to some of the other ID formats already mentioned. I also realized that all of the workarounds (`to_param`, `friendly_id`, etc.) all seem to be strictly on or off, which isn't great because we'll likely need to use some of these IDs in their original format for SEO purposes.

I'm taking some time to work through the whole data set so I can have a better grasp of the scope of the issue, and which parts are Avo-related vs. Rails-related.

Issue #3317, opened 2024-10-09
tags field select mode UX glitches


pjmuller:
### Describe the bug
see loom https://www.loom.com/share/99ec158c8a86415a8b79a680b57b4793
loom 2: https://www.loom.com/share/bd47d51e8aad4764aee4fa8e04f46dfd

1. when clicking the "x" and then clicking out of the element it does not close
2. whenever you selected an item, it stops doing the `fetch_values_from` call until you fully clear and click "x"
3. whenever `fetch_values_from` is working, when you do backspace you remove ALL of your characters while I would expect
4. copy paste in the field is not working

### Steps to Reproduce
Code I used in my loom
```ruby
def fields
    field :id, as: :tags, mode: :select, suggestions: -> { 
      [
        {value: "1", label: "One"},
        {value: "2", label: "Two"},
      ]
     },
     format_using: -> {[{value: value, label: value}]},
     enforce_suggestions: true,
     close_on_select: true
    
    field :id_2, for_attribute: :id, as: :tags, mode: :select,
      fetch_values_from: "/t/dev_org/avo/associations/people",
      format_using: -> { [{value: value, label: value}]},
     enforce_suggestions: true,
     close_on_select: true
```

### System configuration
**Avo version**: 3.13.1

**Rails version**: 7.2.1

**Ruby version**: 3.2.2

**License type**:

 - [ ] Community
 - [ ] Pro
 - [X] Advanced

**Are you using Avo monkey patches, overriding views or view components?**

 - [ ] Yes. If so, please post code samples.
 - [X] No

#### Impact

 - [ ] High impact (It makes my app un-usable.)
 - [X] Medium impact (I'm annoyed, but I'll live.)
 - [ ] Low impact (It's really a tiny thing that I could live with.)

#### Urgency

 - [ ] High urgency (I can't continue development without it.)
 - [X] Medium urgency (I found a workaround, but I'd love to have it fixed.)
 - [ ] Low urgency (It can wait. I just wanted you to know about it.)

Issue #2711, opened 2024-04-21
Searchable `belongs_to` as a default option


adrianthedev:

rctneil:
@adrianthedev Thanks for merging my suggestion with this one.

I did have a go at using the tags field last night instead of text, but was getting an error and i think it was to do with it having trouble joining the value from the database with the array of suggestions from my model method.

Not got any code on me right now but i'll look this evening and post a complete issue if needed, but if you have any suggestions for how to use the tags field instead of text to allow a suggestion list that would be incredible!

adrianthedev:
I don't have a tutorial but that's achievable using `mode: :select` and `fetch_values_from`.

rctneil:
Ah, I don't think i can use that as it looks to save the value as an id.  I need to keep the text field as a text field storing it as a text value, but with an auto-complete or auto suggestion of other values in the same field on other records.

Paul-Bob:
`update_using` helps in this scenario, you can format the input before it's saved.

PR #3381, opened 2024-11-02
feature: combobox field

adrianthedev:
# Description

> [!IMPORTANT]  
> This needs to be another gem.

Fixes https://github.com/avo-hq/avo/issues/3317
Related to https://github.com/avo-hq/avo/issues/2711

Other PRs: https://github.com/avo-hq/avo/pull/2646 & https://github.com/avo-hq/avo/pull/2718

# Checklist:

## Manual review steps

1. Step 1
1. Step 2


adrianthedev:
Still waiting for a response on this thread in regards to supporting objects as values.

https://github.com/josefarias/hotwire_combobox/discussions/221

adrianthedev:
This field needs to be another gem.
The reason is that it does not support Rails 6.1 so it will be difficult to ignore it in tests but also because it adds another dependency (in gemspec and some styles) to installations that potentialy don't need it.

PR #3383, opened 2024-11-03
refactor: `via_resource_class` param

### LINKED ISSUE #2100: refactor `via_resource`
### Context

Right now `via_resource` param look like `Avo%3A%3AResources%3A%3AUser` we should make it something like `avo/resources/user`.

It's important to keep the namespace

thiagoyoussef:
# Description

Fixes #2100 

# Checklist:

## Screenshots & recording

CleanShot 2024-11-03 at 11 14 59@2x

### DIFF
--- a/app/components/avo/fields/belongs_to_field/edit_component.rb
+++ b/app/components/avo/fields/belongs_to_field/edit_component.rb
@@ -71,7 +71,7 @@ class Avo::Fields::BelongsToField::EditComponent < Avo::Fields::EditComponent
   private
 
   def visit_through_association?
-    @field.target_resource.to_s == params[:via_resource_class].to_s
+    @field.target_resource.to_s.demodulize == params[:via_resource_class].to_s.demodulize
   end
 
   def model_keys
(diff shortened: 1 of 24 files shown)


thiagoyoussef:
Considered using:

- `avo/resources/user` -> `avo%2Fresources%2Fuser`
- `avo__resources__user`

I found the simpler solution is to `demodulize` the resource class to `User`. This approach allows us to avoid encoding and at the same time is not needed to format the string back to the class inside `Avo.resource_manager.get_resource` method:

```ruby
def get_resource(resource)
  # Already do the job
  resource = "Avo::Resources::#{resource}" unless resource.to_s.starts_with?("Avo::Resources::")
  ...
end
```

Paul-Bob:
Closing this one until we research a propper approach. Thank you again @thiagoyoussef for all the effort on this one

thiagoyoussef:
100%, I didn't think about the context when there is `Avo::Resources::Link` and `Avo::Resource::Course::Link`.
@Paul-Bob thanks for the feedback!

PR #3398, opened 2024-11-08
Correct how the tags helper gets the text for tags

iainbeeston:
# Description

Fixes the tags test helper so it returns the text of each visible tag.

The tags helper was getting the `label` attribute of `.tagify__tag` elements before but so far as I can see `.tagify__tag` elements do not have a label. Because of that the tags helper was always returning an array of `nil` values (one for each tag). Instead I've changed them to get the text of the tag.

I'm not sure why this wasn't causing the avo test suite to fail before, as there is a test for this helper and the other tags helpers (which return the result of the tags helper) but this seems to work for both the avo test suite and my own code (and matches what I can see when I try avo myself in the browser). It also matches what the `wait_for_tag_to_appear` helper does (ie. looks for tag text, not label).

# Fixes # (issue)

The tags test helper was always returning an array of `nil` values (one for each tag).

# Checklist:

## Manual review steps

So long as the avo test suite still passes this should be ok.

### DIFF
--- a/spec/system/avo/test_helpers_spec.rb
+++ b/spec/system/avo/test_helpers_spec.rb
@@ -222,6 +222,14 @@ RSpec.describe "TestHelpers", type: :system do
         expect(remove_tag(field: :tags, tag: "one")).to eq ["three"]
         expect(tag_suggestions(field: :tags, input: "")).to eq ["one", "two"]
       end
+
+      let!(:post) { create :post, tag_list: ["test", "five", "seven"] }
+
+      it "verify tags" do
+        visit avo.edit_resources_post_path(post)
+
+        expect(tags(field: :tags)).to eq ["test", "five", "seven"]
+      end
       # end
     end
   end
--- a/spec/system/avo/test_helpers_spec.rb
+++ b/spec/system/avo/test_helpers_spec.rb
@@ -223,12 +223,12 @@ RSpec.describe "TestHelpers", type: :system do
         expect(tag_suggestions(field: :tags, input: "")).to eq ["one", "two"]
       end
 
-      let!(:post) { create :post, tag_list: ["test", "five", "seven"] }
+      let!(:post) { create :post, tag_list: ["one", "two"] }
 
       it "verify tags" do
         visit avo.edit_resources_post_path(post)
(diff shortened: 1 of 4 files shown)


spec/system/avo/test_helpers_spec.rb
```diff
@@ -222,6 +222,14 @@
         expect(remove_tag(field: :tags, tag: "one")).to eq ["three"]
         expect(tag_suggestions(field: :tags, input: "")).to eq ["one", "two"]
       end
+
+      let!(:post) { create :post, tag_list: ["test", "five", "seven"] }
+
+      it "verify tags" do
+        visit avo.edit_resources_post_path(post)
+
+        expect(tags(field: :tags)).to eq ["test", "five", "seven"]
```
iainbeeston, line 231: It looks like in the test this is coming out as `[“seven”, “five”, “test”]`

Paul-Bob: True, it was passing locally, but the tags field on the post has enforced suggestions so it's flaky... Just changed it to `["one", "two"]` should work as expected not

Paul-Bob (approved):
> I'm not sure why this wasn't causing the avo test suite to fail before, as there is a test for this helper and the other tags helpers (which return the result of the tags helper) 

The test case wasn't covering it enough, thanks for this contribution @iainbeeston .

I've committed a test for it.

Paul-Bob:
@iainbeeston this helper seems a bit unpredictable, it returns tags in different orders each time. Any ideas why?

iainbeeston:
@Paul-Bob I think that's happening because acts as taggable does not preserve the order of tags by default. For this test to work the way you want you'll need to update `Post` from using `acts_as_taggable_on :tags` to `acts_as_ordered_taggable_on :tags` (documentation is [here](https://github.com/mbleigh/acts-as-taggable-on?tab=readme-ov-file#usage))

iainbeeston:
The other option would be to use the `contains_exactly` matcher on the list of tags returned.

Paul-Bob:
That was it, thanks for looking into this @iainbeeston !

PR #3418, opened 2024-11-13
fix: associations query limit on select options

### LINKED ISSUE #2406: Prevent memory bloat in associations when searchable is set to false

thiagoyoussef:
# Description

Fixes #2406 

# Checklist:

## Screenshots & recording

On attach model:
CleanShot 2024-11-13 at 15 13 56@2x

On edit belongs to:
CleanShot 2024-11-13 at 15 17 32@2x


Paul-Bob (changes requested):
The `query.all.count` will perform sub-optimal on large datasets

From [this stackoverflow thread](https://stackoverflow.com/questions/55018986/postgresql-select-count-query-takes-long-time):
>  I just ran into a situation where a table of 1.3 million rows was taking about 35 seconds to perform a simple SELECT COUNT(*).

> For our table with 20M records, it took 4 minutes.

Instead, I suggest fetching records with the defined limit, then checking the size of the retrieved dataset. If the result size matches the limit, two scenarios are likely:

- There are more records beyond the limit (most probable).
- The limit precisely matches the total record count (less less less probable).

---

Let's discuss the option name and the default limit.

Otherwise this PR looks good! Thank you @thiagoyoussef

Paul-Bob (approved):

Paul-Bob (approved):

Issue #3140, opened 2024-08-14
Implement radio field


Paul-Bob:
### Context

Similar to the boolean group field but using radio input so the options become exclusive

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio

Originally requested [here](https://discord.com/channels/740892036978442260/1273023803168915608)

zhephyn:
Hey @Paul-Bob. I'll take up this one

Paul-Bob:
Great @zhephyn! This one is challenging! Let me know if you find any obstacles

Paul-Bob:
Thank you!

zhephyn:
Hey  @Paul-Bob. I'm trying to access the discord channel where the feature is stated to have been requested for initially, but it seems the link is broken. Could you provide some more context about the feature to be implemented or maybe the specific page in the application where the feature ought to be implemented. Thank you

Paul-Bob:
Hey @zhephyn, there's not much happening on the Discord thread, it's mainly here to track the source of the original request.

>Could you provide some more context about the feature to be implemented or maybe the specific page in the application where the feature ought to be implemented. Thank you

Sure. The idea is to implement a new field that will allow this DSL:

```ruby
field :gender, as: :radio
```

This will use a radio input element: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio

Let's take the `boolean_group` field as a reference for structuring it:

- Field definition file here: `lib/avo/fields/boolean_group_field.rb`
- Component files here: `app/components/avo/fields/boolean_group_field`

We need to build a file to keep the `radio` field and components.

zhephyn:
Hey @Paul-Bob. It seems this issue received a PR that could result into it becoming a closed one. I was still working through my solution, but now I'm not sure how to proceed. Should i abandon working on and  submitting a PR for this issue and pick up another one? Or should i continue working on it? Any guidance is appreciated. Thanks in advance.

Paul-Bob:
The PR for this issue is on track to be merged soon, which will indeed close it.

You're still welcome to open a PR for your current work if you'd like some feedback, though it won't be merged, I can review it to provide any insights or suggestions

I'm sorry that it happened again. I've just created this issue: https://github.com/avo-hq/avo/issues/3426. If you're interested, we can assign it exclusively to you. Let me know!

Thank you @zhephyn !

PR #3425, opened 2024-11-14
feature: radio field

thiagoyoussef:
# Description

Fixes https://github.com/avo-hq/avo/issues/3140

# Checklist:

## Screenshots & recording

CleanShot 2024-11-14 at 16 30 34@2x

### DIFF
--- a/spec/dummy/app/avo/resources/person.rb
+++ b/spec/dummy/app/avo/resources/person.rb
@@ -10,7 +10,7 @@ class Avo::Resources::Person < Avo::BaseResource
 
   def fields
     field :name, as: :text, link_to_record: true, sortable: true, stacked: true
-    field :type, as: ::Pluggy::Fields::RadioField, name: "Type", options: {Spouse: "Spouse", Sibling: "Sibling"}, include_blank: true, filterable: true
+    field :type, as: :pluggy_radio, name: "Type", options: {Spouse: "Spouse", Sibling: "Sibling"}, include_blank: true, filterable: true
     field :link, as: :text, as_html: true
     field :spouses, as: :has_many, hide_search_input: true
     field :person, as: :belongs_to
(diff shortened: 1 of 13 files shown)


spec/dummy/app/avo/resources/person.rb
```diff
@@ -10,7 +10,7 @@ class Avo::Resources::Person < Avo::BaseResource
 
   def fields
     field :name, as: :text, link_to_record: true, sortable: true, stacked: true
-    field :type, as: ::Pluggy::Fields::RadioField, name: "Type", options: {Spouse: "Spouse", Sibling: "Sibling"}, include_blank: true, filterable: true
```
adrianthedev, line 13: Awesome!

spec/system/avo/radio_field_spec.rb
```diff
@@ -0,0 +1,60 @@
+require "rails_helper"
+
+RSpec.describe "RadioField", type: :system do
```
adrianthedev, line 3: I wonder if we can get away with a feature test instead of a system one.

adrianthedev (approved):
Love it!
Thank you!

spec/dummy/app/avo/resources/person.rb
```diff
@@ -10,7 +10,7 @@ class Avo::Resources::Person < Avo::BaseResource
...
+    field :type, as: :pluggy_radio, name: "Type", options: {Spouse: "Spouse", Sibling: "Sibling"}, include_blank: true, filterable: true
```
Paul-Bob, line 13: Is this change really necessary? I think it's a good test for declaring fields using the constant.

Paul-Bob (approved):
Looking great! Thank you @thiagoyoussef !

spec/dummy/app/avo/resources/person.rb
thiagoyoussef: Will change to use the constant then! It's not necessary, just thought would be a better practice.

PR #3442, opened 2024-11-19
Feature: Title and description for tabs

### LINKED ISSUE #3421: Title and description for tabs
### Feature

Create a header section for tabs that displays a title and a description, similar to the resource's title and description.

DSL should look something like this:

```ruby
tabs "Title here", description: "Description here" do
  # ...
end
```

Image

Nevelito:
# Description
- Refactor tab_groups to have title and description

Fixes #3421

image

# Checklist:

  - https://github.com/avo-hq/docs.avohq.io/pull/321


Nevelito:
I think I mede it as I should, but I do not know what can I put into title and description of tabs. Also I don't know why tab_group_component_spec and tab_swicher_component_spec are empty.

Paul-Bob (changes requested):
While manually testing this I noticed something, title and description would look nicer outside of the tabs border:

image

And let's try to extract the header section from `app/components/avo/panel_component.html.erb` and reuse it for tabs instead of creating a new component.

app/components/avo/tab_group_component.html.erb
```diff
@@ -10,7 +10,8 @@
   } do %>
   <% visible_tabs.each_with_index do |tab, index| %>
     <%= content_tag :div, **args(tab) do %>
-      <div class="border rounded-lg p-2 -mx-2 -my-2 lg:p-4 lg:-mx-4 lg:-my-4 space-y-4">
+      <%= render Avo::PanelComponent.new name: group.title, description: group.description %>
```
Paul-Bob, line 13: Can we extract the header section from the `Avo::PanelComponent` and reuse it here? I.E Having a `Avo::PanelHeaderCompontent` and render it here and inside `Avo::PanelComponent`?

Paul-Bob (changes requested):
Seems a bit to much to render the whole `Avo::PanelComponent` when we only want to render the header section. Le'ts create a new component, extract the header section into that component and use the new `Avo::PanelHeaderComponent`

app/components/avo/panel_header_component.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  prop :name
+  prop :description
+  prop :display_breadcrumbs, default: false
+  prop :profile_photo
+
+  private
+
+  def display_breadcrumbs?
+    @display_breadcrumbs == true && Avo.configuration.display_breadcrumbs == true
```
KamilMilewski, line 17:
```suggestion
    @display_breadcrumbs && Avo.configuration.display_breadcrumbs
```

spec/components/avo/panel_header_component_spec.rb
```diff
@@ -0,0 +1,11 @@
+require "rails_helper"
+
+RSpec.describe Avo::PanelHeaderComponent, type: :component do
+  # it "renders something useful" do
```
KamilMilewski, line 4: should we actually implement this spec?

adrianthedev: I think we should. Not sure why it's commented out.

Nevelito: All of spec are made like this.

Nevelito: image

Paul-Bob: > should we actually implement this spec?

Not really, it was generated while generating the component

> I think we should. Not sure why it's commented out.

It was generated during the component generation, we don't implement component tests yet so it can be removed for this PR

Paul-Bob (approved):
Thank you @Nevelito! It's looking good!

I've made some minor commits to complete this PR and include it in tomorrow's release.

PR #3452, opened 2024-11-21
Fix Turbo.config.forms.confirm deprecation message

iainbeeston:
# Description

Turbo 8 deprecated `Turbo.setConfirmMethod()` and instead moved it to `Turbo.config.forms.confirm` instead (see https://github.com/hotwired/turbo/pull/1216). Because of that, avo is throwing deprecation messages in the javascript console:

    Please replace `Turbo.setConfirmMethod(confirmMethod)` with `Turbo.config.forms.confirm = confirmMethod`. The top-level function is deprecated and will be removed in a future version of Turbo.

I've updated the code to use the new config object instead, which should resolve the deprecation warning.

Fixes # (issue)

# Checklist:

## Manual review steps

1. Confirm that turbo confirmation popups still work.

### DIFF
--- a/app/javascript/js/custom-confirm.js
+++ b/app/javascript/js/custom-confirm.js
@@ -1,6 +1,6 @@
 import { Turbo } from '@hotwired/turbo-rails'
 
-Turbo.setConfirmMethod((message) => {
+Turbo.config.forms.confirm = (message) => {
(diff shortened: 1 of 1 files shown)


Paul-Bob (approved):
Looking good! Thank you @iainbeeston

PR #3461, opened 2024-11-25
Don't refer to more content button if always_true is set

iainbeeston:
# Description

If you set `always_show: true` on a trip field and it has more than 50 characters there is a javascript error:

    Error: Missing target element "moreContentButton" for "trix-body" controller

This is because the trix field show component only adds the more content button when `always_show` is `false` but the trix controller always refers to that component without checking if it exists first.

I've updated it so that it checks if that target exists before manipulating it.

Fixes # (issue)

# Checklist:

## Manual review steps

1. Find a trix field that has `always_show: true` set
2. Add a lot of text to it (more than 50 characters)
3. Go to the show view for that component
4. Confirm that it shows a javascript error on `main` but on this branch it doesn't show an error (on both it should not show an error)
5. Change the field to `always_show: false`
6. Reload the show view
7. Confirm that the show more button is visible and works.


Paul-Bob (approved):
Thank you for this contribution @iainbeeston!

Issue #2769, opened 2024-05-15
Click to copy field option


adrianthedev:
### Feature

We'd love to have a Click to copy field option.

- [ ] It would show up with icon and text on the Show view to the far right.
- [ ] It would show up with just the icon on the Index view right next to the value to the right.

Similar PR https://github.com/avo-hq/avo/issues/2771

### Current workarounds

Custom field.

### Screenshots or screen recordings

![CleanShot 2024-05-23 at 10 36 56@2x](https://github.com/avo-hq/avo/assets/1334409/ccd82f59-b197-4fb4-aed9-aaa1fd5c85aa)

### Additional context

Part of the code I used to get the screenshot. The `@index` part need to be taken out.
```erb
      <% if on_show? %>
        <% if field.value.blank? and dash_if_blank %>
          —
        <% else %>
        <div class="flex">
          <% if @index.in?([1,5]) %>
          <span class="flex-1"><%= content %></span>
          <a href="javascript:void(0)" class=" text-xs">
            <%= helpers.svg 'heroicons/outline/clipboard', class: "h-4 inline" %> click to copy
          </a>
        <% end %>
        </div>
        <% end %>

```

This is a JS controller we use on avohq.io (on the hero section)

```js
import { Controller } from '@hotwired/stimulus'

export default class extends Controller {
  static targets = ['trigger']

  copy() {
    const initialContent = this.triggerTarget.innerHTML

    const str = this.context.element.dataset.text
    /* ——— Derived from: https://hackernoon.com/copying-text-to-clipboard-with-javascript-df4d4988697f
          improved to add iOS device compatibility——— */
    const el = document.createElement('textarea'); // Create a <textarea> element

    let storeContentEditable = el.contentEditable;
    let storeReadOnly = el.readOnly;

    el.value = str; // Set its value to the string that you want copied
    el.contentEditable = true;
    el.readOnly = false;
    el.setAttribute('readonly', false); // Make it readonly false for iOS compatability
    el.setAttribute('contenteditable', true); // Make it editable for iOS
    el.style.position = 'absolute';
    el.style.left = '-9999px'; // Move outside the screen to make it invisible
    document.body.appendChild(el); // Append the <textarea> element to the HTML document
    const selected =
        document.getSelection().rangeCount > 0 // Check if there is any content selected previously
            ? document.getSelection().getRangeAt(0) // Store selection if found
            : false; // Mark as false to know no selection existed before
    el.select(); // Select the <textarea> content
[...]

PR #3468, opened 2024-11-28
Feature/click to copy

Nevelito:
# Description

1. add new feature - add button to copy to clipboard
2. install stimulus-clipboard
3. add visability in hover
4. add new component to optymalize code
5. register JS controller

Fixes # (issue)

https://github.com/avo-hq/avo/issues/2769

# Checklist:

## Screenshots & recording
image

image

image

image

image

### DIFF
--- a/app/components/avo/fields/country_field/index_component.html.erb
+++ b/app/components/avo/fields/country_field/index_component.html.erb
@@ -1,7 +1,12 @@
 <%= index_field_wrapper **field_wrapper_args do %>
-  <% if @field.display_code %>
-    <%= @field.value %>
-  <% else %>
-    <%= @field.countries[@field.value] %>
-  <% end %>
+  <div class="flex flex-row items-center justify-between space-x-1 group">
+    <% if @field.display_code %>
+      <%= @field.value %>
+    <% else %>
+      <%= @field.countries[@field.value] %>
+    <% end %>
+    <% if copyable? %>
+      <%= render Avo::ClipboardComponent.new value: @field.value, view: 'index' %>
+    <% end %>
+  </div>
 <% end %>
--- a/app/components/avo/fields/show_component.rb
+++ b/app/components/avo/fields/show_component.rb
@@ -61,4 +61,8 @@ class Avo::Fields::ShowComponent < Avo::BaseComponent
   def disabled?
     field.is_readonly? || field.is_disabled?
   end
+
+  def copyable?
+    field.respond_to?(:copyable) && field.copyable
+  end
 end
--- a/app/components/avo/clipboard_component.html.erb
+++ b/app/components/avo/clipboard_component.html.erb
@@ -1,4 +1,4 @@
-<div class="flex flex-row justify-between" data-controller="clipboard" data-clipboard-success-content-value=<%= success_value %>>
+<div class="flex flex-row justify-between opacity-0 group-hover:opacity-100 transition-opacity duration-200" data-controller="clipboard" data-clipboard-success-content-value=<%= success_value %>>
   <div class="hidden" data-clipboard-target="source"><%= @value %></div>
   <div class="text-xs xl:text-sm">
     <button type="button" data-action="clipboard#copy" data-clipboard-target="button">
--- a/app/components/avo/clipboard_component.html.erb
+++ b/app/components/avo/clipboard_component.html.erb
@@ -1,0 +1,11 @@
+<div class="flex flex-row justify-between" data-controller="clipboard" data-clipboard-success-content-value=<%= success_value %>>
+  <div class="hidden" data-clipboard-target="source"><%= @value %></div>
+  <div class="text-xs xl:text-sm">
+    <button type="button" data-action="clipboard#copy" data-clipboard-target="button">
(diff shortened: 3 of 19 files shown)


KamilMilewski:
We changed quite a lot in existing components or added new ones. Should we cover those changes with specs?

spec/dummy/db/schema.rb
```diff
@@ -12,7 +12,7 @@
 
 ActiveRecord::Schema[8.0].define(version: 2024_11_14_165947) do
   # These are extensions that must be enabled in order to support this database
-  enable_extension "plpgsql"
+  enable_extension "pg_catalog.plpgsql"
```
KamilMilewski, line 15: Wonder why schema changed, considering there is no migrations present 🤔

app/components/avo/fields/show_component.rb
```diff
@@ -61,4 +61,8 @@ def field_wrapper_args
   def disabled?
     field.is_readonly? || field.is_disabled?
   end
+
+  def copyable?
+    field.respond_to?(:copyable) && field.copyable
```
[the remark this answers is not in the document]
Nevelito: That is solution but only fields with text should have option to be copyable and only in show and index view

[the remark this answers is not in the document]
KamilMilewski: Yeah, my point is to try to avoid usage of `respond_to?` which is a code smell. You are already trying to ask every field if its copyable of not so you could as like allow every fileld to be able to respond to this msg (and return false by default for example by impelemnting this method in some base field and then overriging it with true where it apply)

spec/dummy/db/schema.rb
Paul-Bob: Agree, let's revert this schema change

app/components/avo/fields/show_component.rb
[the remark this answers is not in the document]
Paul-Bob: I agree with @KamilMilewski.  

Let's insert the `copyable` property on the base field instead of adding it to each individual field we want to support. Set the default value to `false`, allowing us to simply check `field.copyable`.  

This approach reduces complexity for now.

app/components/avo/fields/country_field/index_component.html.erb
```diff
@@ -1,7 +1,12 @@
...
-  <% end %>
+  <div class="flex flex-row items-center justify-between space-x-1 group">
+    <% if @field.display_code %>
+      <%= @field.value %>
+    <% else %>
+      <%= @field.countries[@field.value] %>
+    <% end %>
+    <% if copyable? %>
+      <%= render Avo::ClipboardComponent.new value: field_value %>
+    <% end %>
```
Paul-Bob, line 10: Let's explore an alternative approach: render this once on the `index_field_wrapper` instead of manually adding it to each field we want to support. This would significantly reduce complexity.

Let's consider applying this approach to both `index_field_wrapper` and `field_wrapper`.

Paul-Bob (changes requested):
Thank you for all your effort here, @Nevelito!

I've been testing it locally, and it's already working as expected.

Let's make some visual adjustments and apply the same design used in the Docs ([check it here](https://docs.avohq.io/3.0/resources.html#resources-from-model-generation)).

Specifically:

- Display the empty clipboard on hover, without the text.
- After a click, display the same clipboard with a checkmark inside it.
Later, we could add a title to the icon and display it on hover.

I've also included some comments about the code.

Let me know if you have any questions.

Paul-Bob:
> We changed quite a lot in existing components or added new ones. Should we cover those changes with specs?

Let's do a system spec at the end that tests the copyable functionality, not component specific.

app/javascript/js/controllers/clipboard_controller.js
```diff
@@ -0,0 +1,20 @@
...
+
+  copied() {
+    this.iconTarget.classList.add('hidden')
+    this.iconCopiedTarget.classList.remove('hidden')
+
+    // Reset the icon after a 2 seconds delay
+    setTimeout(() => {
+      this.iconTarget.classList.remove('hidden')
+      this.iconCopiedTarget.classList.add('hidden')
+    }, 2000)
```
Paul-Bob, line 18: I think this can be configured with `data-clipboard-success-duration-value` 

From: https://www.stimulus-components.com/docs/stimulus-clipboard

Paul-Bob (changes requested):
Let's use the configurations that https://www.stimulus-components.com/docs/stimulus-clipboard already supports

app/javascript/js/controllers/clipboard_controller.js
Nevelito: It is not that easy, It works with  **data-clipboard-success-content-value** (this work only with plain text) , we want to show another object so we can not probably use **data-clipboard-success-duration-value**. I tried use that but it do not work.

Paul-Bob (approved):
Thank you for this contribution @Nevelito it's looking great! 

I just pushed some commits to complete the PR.

Paul-Bob:
TODO:

- [ ] add tool-tip 
- [ ] add localization key for `copy`

PR #3502, opened 2024-12-09
enhancement: keep all referrer_params on go back link

### LINKED ISSUE #3497: Keep filter when navigating back to page from record
### Context
Is is possible to keep the filter on the page after going into record and back out to the resource

### System configuration
**Avo version**: 3.15.0

**Rails version**: 7.2.0

**Ruby version**: 3.3.1

**License type**

 - [ ] Community
 - [x] Pro

 - [x] No

https://github.com/user-attachments/assets/bea5f287-7a34-45dd-8f77-91b8fcf24c7c

Paul-Bob:
# Description

Fixes #3497

Keeping all `referrer_params` on the "go back" link ensures that the scopes/filters applied, along with other parameters, are preserved.

# Checklist:

## Manual review steps

1. Step 1
1. Step 2


johnvtroy:
Hi @Paul-Bob this doesn't seem to work for has_many associations. Is this it possible to to have this on associations too?

Paul-Bob:
> Hi @Paul-Bob this doesn't seem to work for has_many associations. Is this it possible to to have this on associations too?

Hi @johnvtroy, unfortunately handling associations like this isn't straightforward. Each association is wrapped in a Turbo frame with a src attribute. To preserve the context for all associations when navigating back using the go back button, we'd need to maintain somewhere the state of all the params for each association and pass they to the src.

johnvtroy:
> > Hi @Paul-Bob this doesn't seem to work for has_many associations. Is this it possible to to have this on associations too?
> 
> Hi @johnvtroy, unfortunately handling associations like this isn't straightforward. Each association is wrapped in a Turbo frame with a src attribute. To preserve the context for all associations when navigating back using the go back button, we'd need to maintain somewhere the state of all the params for each association and pass they to the src.

Could i raise an enhancement request for this? Depending on how long it might take to make it into Avo I may need to refactor my implementation.

Paul-Bob:
Sure @johnvtroy! Feel free to open a request and we can continue the discussion there

PR #3515, opened 2024-12-11
fix: search_controller null error

Paul-Bob:
# Description

Fixes the JS search controller.

`.aa-DetachedOverlay` is already getting removed when a value is selected from the search list, there is no need to remove it with code.

Fixes:

```ruby
search_controller.js:223 Uncaught (in promise) TypeError: Cannot read properties of null (reading 'remove')
    at n.handleOnSelect (search_controller.js:223:51)
```

# Checklist:


thiagovsk:
👏

priscilaformaggio:
👏

PR #3520, opened 2024-12-12
fix: standalone action disabled and enabled classes

### LINKED ISSUE #3395: non standalone actions on custom controls
### Describe the bug
related to https://github.com/avo-hq/avo-advanced/pull/50

see https://www.loom.com/share/ee833adc463442ffb462435b77506b3d that demonstrates the problems below

1. icon + text not grayed out while disabled (so not a clear signal to the user what is available/unavailable)
2. when no text present seems to still open standalone=false when no records are selected
3. inconsistent behavior, sometimes (but not always) when unchecking the checkbox, we are able to open the action

feature / code snippet request

- could you explain how we can have a tooltip on the disabled actions saying they should first select an item, this will make it more foolproof

FYI: I also ran the same experiments having chrome developer console open + disable cache, so I should have all latest assets

```ruby
module Avo
  module Resources
    class Tenant < Avo::BaseResource
      self.index_controls = -> do
        action Avo::Actions::Dummy, icon: 'heroicons/outline/envelope', style: :outline, title: 'Stuur Mailsjabloon', label: ''
      end
end

module Avo
  module Actions
    class Dummy < Avo::BaseAction
      self.standalone = false
      self.name = 'Dummy'
      # ...
    end
  end
end

```
### System configuration
**Avo version**: 3.14

**Rails version**: 7.2

 - [X] No

#### Impact

#### Urgency

 - [X] Medium urgency (I found a workaround, but I'd love to have it fixed.)

Paul-Bob:
# Description

Fixes #3395

Adjust opacity for disabled actions and resolve an issue where row actions became disabled after toggling the selection of a record.

https://github.com/user-attachments/assets/69120c1b-b7b9-4b85-9b7c-c4abbb0e120d

# Checklist:


pjmuller:
Hi @Paul-Bob thanks for shipping this,
1. I guess it's not in 3.15.3 yet right? Can you add it in 3.15.4, our customers are getting confused and asking questions about this...

2. when we spoke about this you said I would need to specify specific css classes for inactive. Is this still the case
the code suggests that you implemented it in a generic way I guess ? `disabled:opacity-60 data-[disabled='true']:opacity-60` (which is perfect to me) just want to know how to use it when it ships ;)

Paul-Bob:
Hi @pjmuller this will be included on 3.15.4 which we're releasing during this week.

> when we spoke about this you said I would need to specify specific css classes for inactive. Is this still the case
the code suggests that you implemented it in a generic way I guess ? disabled:opacity-60 data-[disabled='true']:opacity-60 (which is perfect to me) just want to know how to use it when it ships ;)

Yes we opted for the more generic implementation with the opacity, I mentioned it [here](https://github.com/avo-hq/avo/issues/3395#issuecomment-2510923638). No action is required on your part during the update; it will be applied automatically.

PR #3547, opened 2024-12-20
Set "as" attribute on stylesheet link tag

iainbeeston:
# Description

In the browser javascript console I see this error on every page load:

    The resource http://localhost:3000/avo-assets/avo.base.css was preloaded using link preload but not used within a few seconds from the window's load event. Please make sure it has an appropriate `as` value and it is preloaded intentionally."

It's not causing me any issues as such - it's just noisy and can make it harder to see real issues. It seems like a simple thing to fix so I've raised a PR.

# Checklist:

## Manual review steps

1. The avo dashboard should load with the correct styles and show no stylesheet warning in the browser console.


adrianthedev:
I didn't know about the `as` attribute

adrianthedev (approved):

PR #3572, opened 2025-01-03
feature: Implement multiple option for select field

### LINKED ISSUE #3426: Implement `multiple` option for select field.
### Feature

Right now the select field do not allow multiple selections. Let's add the `multiple` option for the select field which will enable the multiple-select mode.

### Approach

https://www.loom.com/share/7cdb77e591de4c27bb6115afab147b7f?sid=0748b730-efc4-45ad-b08e-8c5ad71a4dbc

- [ ] Add `multiple` option on the select field that defaults to `false`
- [ ] Pass the `field.multiple` option to the form helper on the view components
- [ ] Ensure that is working as expected and add tests for the option
- [ ] Document the `multiple` option including the version since its available

https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_select_multiple

zhephyn:
# Description

Add a `multiple` boolean option to the select field, allowing for multiple selections.

Fixes #3426

# Checklist:

  - https://github.com/avo-hq/docs.avohq.io/pull/342

## Manual review steps

Added two extra tests(one for selecting multiple sizes when creating a new product, another for editing the sizes list of an existing product) to the select_field_spec.rb file. These all pass when i run the command "bin/test ./spec/feature/avo/select_field_spec.rb".


lib/avo/fields/select_field.rb
```diff
@@ -55,6 +56,15 @@ def label
         display_value ? value : options.invert[value]
       end
 
+      def to_permitted_param
+        case id
+          when :sizes
+            {"#{id}": []}
+          else
+            id
+          end
```
Paul-Bob, line 65: Let's perform the conditional check on `@multiple`. Right now it will only apply to fields named `sizes`.

```suggestion
        return id unless @multiple

        {"#{id}": []}
```

app/components/avo/fields/select_field/edit_component.html.erb
```diff
@@ -1,9 +1,11 @@
...
-      include_blank: @field.include_blank
-    },
-    aria: {
-      placeholder: @field.placeholder
+      include_blank: @field.include_blank,
+      multiple: @field.multiple,
+      aria: {
+        placeholder: @field.placeholder
+      },
+      include_hidden: false
```
Paul-Bob, line 8: Is it necessary to put the `aria` inside the same hash as `include_blank`? Why aren't we keeping the same syntax as before?

I think `include_hidden: false` is causing an issue:

1. Create a product with 2 random sizes
2. Verify that sizes are correctly assigned
3. Edit that product
4. Deselect all the sizes
5. Save
6. Notice that product is keeping the 2 sizes instead empty

Paul-Bob, line 8:
```suggestion
      include_blank: @field.include_blank,
      multiple: @field.multiple
    },
    aria: {
      placeholder: @field.placeholder
```

spec/features/avo/select_field_spec.rb
```diff
@@ -229,4 +229,33 @@ def test_array
       end
     end
   end
+  describe "when options are multiple select" do
+    context "new" do
+      it "allow selection of multiple values" do
+        Avo::Resources::Product.with_temporary_items do
+          field :sizes, as: :select, multiple: true, options: { 'Large': :large, 'Medium': :medium, 'Small': :small}
+        end
```
Paul-Bob, line 237:
```suggestion
```

No need for `Avo::Resources::Product.with_temporary_items` since the `sizes` field is already configured on the dummy's resource.

spec/features/avo/select_field_spec.rb
```diff
@@ -229,4 +229,33 @@ def test_array
...
+        save
+        expect(Product.last.sizes).to match_array(["large", "medium"])
+      end
+    end
+    context "edit" do
+      let (:product) {create :product, sizes: [:large]}
+      it "allow changing of selected values" do
+        Avo::Resources::Product.with_temporary_items do
+          field :sizes, as: :select, multiple: true, options: { 'Large': :large, 'Medium': :medium, 'Small': :small}
+        end
```
Paul-Bob, line 251:
```suggestion
```

No need for `Avo::Resources::Product.with_temporary_items` since the `sizes` field is already configured on the dummy's resource.

Paul-Bob (changes requested):
Hi @zhephyn 

Thanks for this contribution. It's going on the right direction!

I left some comments on the code.

I also noticed that when `multiple` is true the show and index views are not showing the value as expected.

Let's address the code comments and the displaying issue before the next review.

Let me know if I can help.

Thank you!

app/components/avo/fields/select_field/edit_component.html.erb
zhephyn: The purpose of include_hidden: false was because, without it, when I select the sizes when creating a new product, an empty string is added as part of the submitted array. For example, if I was to select "large" and "small" as the sizes for a new product, on saving the product to the database, the array looks like this ["/","large","small"] without include_hidden: false. Should I remove it altogether or should I solve the error so that it can be kept? All in all, it's purpose was preventing the empty string from being saved as part of the sizes array.

Paul-Bob: I'm not sure if we should remove or keep it, but this is the expected behavior:

If `large` and `small` is selected than the array should look like: `["large", "small"]`

After having it saved as `["large", "small"]` and on edit both are deselected the array should look like: `[]`

lib/avo/fields/select_field.rb
```diff
@@ -55,6 +56,11 @@ def label
         display_value ? value : options.invert[value]
       end
 
+      def to_permitted_param
+        return id unless @multiple
+        {"#{id}": []}
```
Paul-Bob, line 61:
```suggestion
        @multiple ? {"#{id}": []} : id
```

spec/dummy/app/models/product.rb
```diff
@@ -31,4 +33,10 @@ class Product < ApplicationRecord
 
   has_one_attached :image
   has_many_attached :images
+
+  private
+
+  def sanitize_sizes
+    self.sizes = sizes.reject(&:blank?)
+  end
```
Paul-Bob, line 41: This logic doesn't really fit here. We should make sure the behavior discussed [here](https://github.com/avo-hq/avo/pull/3572#discussion_r1906817998) works without requiring extra setup from the user. 

As it stands, this logic has to be manually added for every multiple select in the app, and that's definitely not ideal.

One way to tackle this might be by overriding the `fill_field` method on the select field, like so:

```ruby
      def fill_field(record, key, value, params)
        if @multiple
          value = sizes.reject(&:blank?)
        end

        super
      end
```

I haven't tested this out yet, so let's also brainstorm other potential approaches to solve this.

Paul-Bob (changes requested):
Hey @zhephyn, thanks for the updates!

There's a bit of an issue with the sanitization. It really should be handled automatically by Avo rather than shifting that responsibility to the user. 

Other than that, we're pretty much at the finish line! 🚀

spec/dummy/app/models/product.rb
zhephyn: Ohh, makes sense now. I was sanitizing the sizes for just the product model yet this is a feature that should be available for other models too.

zhephyn: "sizes" when used directly like in the above code, is not  recognized in the context of the Select field class, thus i ended up with an "undefined variable or method "sizes" " error. Made some changes though and used this instead
`value = value.reject(&:blank?)`
Everything worked as expected and all the tests pass.

zhephyn:
> Hey @zhephyn, thanks for the updates!
> 
> There's a bit of an issue with the sanitization. It really should be handled automatically by Avo rather than shifting that responsibility to the user.
> 
> Other than that, we're pretty much at the finish line! 🚀

I've updated the PR to reflect the changes you requested for. In particular, the ones relating to sanitization.

spec/dummy/app/models/product.rb
Paul-Bob: Exactly, @zhephyn! That's precisely what I meant. And yes, sizes are specific to the product model, so using value is indeed the right approach. Thank you!

Paul-Bob:
> I've updated the PR to reflect the changes you requested for. In particular, the ones relating to sanitization.

Awesome! Let's please ensure that lint and the test actions are not failing, you can check the details for each failing action, let me know if you need any help interpreting those.

zhephyn:
> > I've updated the PR to reflect the changes you requested for. In particular, the ones relating to sanitization.
> 
> 
> 
> Awesome! Let's please ensure that lint and the test actions are not failing, you can check the details for each failing action, let me know if you need any help interpreting those.

The current 3 tests I had written to test for the new and edit actions relating to the field all pass as expected. The last 3 tests in the select_field_spec are the ones I'm referring to. Are there any other tests that test for this field that I should ensure "work fine"?

zhephyn:
I've updated the PR to fix the issues that were flagged by rubocop relating to double and single quote syntax

zhephyn:
Hey @Paul-Bob, the issues I'm seeing on checking the PR, the one for rubocop seems understandable and im gonna fix it. As for the system and feature specs, the runner is failing at bundle install and throws an argument error saying that version 8.0 is not recognized with the highest version being 7.2. This doesn't quite make sense given the fact that, locally everything works and this is a rails 8 application

db/migrate/20250103192051_add_sizes_to_products.rb
```diff
@@ -0,0 +1,5 @@
+class AddSizesToProducts < ActiveRecord::Migration[8.0]
```
Paul-Bob, line 1:
```suggestion
class AddSizesToProducts < ActiveRecord::Migration[6.1]
```

> As for the system and feature specs, the runner is failing at bundle install and throws an argument error saying that version 8.0 is not recognized with the highest version being 7.2. This doesn't quite make sense given the fact that, locally everything works and this is a rails 8 application. The tests are being run against rails version 7.2 yet the application's migration version is 8.0.

Tests are designed to run across multiple Rails versions, `6.1`, `7.1`, and `8.0`, to verify compatibility.

If the migration version is set to `8.0`, it will fail on Rails `6.1` and `7.1`.

To resolve this, you can manually adjust the migration version to `6.1`.

zhephyn:
Hey @Paul-Bob. I've updated the branch to solve the migration version error plus the double string rubocop error. 
PS: Sorry about the delay

db/migrate/20250103192051_add_sizes_to_products.rb
```diff
@@ -0,0 +1,5 @@
+class AddSizesToProducts < ActiveRecord::Migration[7.1]
```
Paul-Bob, line 1:
```suggestion
class AddSizesToProducts < ActiveRecord::Migration[6.1]
```

zhephyn:
I thought the migration version had to correspond to the rails version installed by the bundle command when running the CI test. That's why I had specified version 7.1, I'm going to update the PR and use rails version 6.1 instead. Can't wait for the PR to get merged 🤞. 
PS: Thank you so much for your patience. 🙏

Paul-Bob:
> Hey @Paul-Bob. I've updated the branch to solve the migration version error plus the double string rubocop error.
> PS: Sorry about the delay

Thanks @zhephyn, no problem!

> I thought the migration version had to correspond to the rails version installed by the bundle command when running the CI test. That's why I had specified version 7.1, I'm going to update the PR and use rails version 6.1 instead. Can't wait for the PR to get merged 🤞.
PS: Thank you so much for your patience. 🙏

The version must be `6.1` or lower since that's the minimum version used for testing.

I've observed some test failures, could you please investigate those?

zhephyn:
The tests are failing because rails possibly has no inbuilt provision for allowing multiple values for an enum declaration in a model. Rails expects that an enum should allow single and not multiple selection.
In the project model file, we define an enum called "Stage" and provide the options including "On hold, cancelled, drafting" and so on. Then in the schema.rb, we can see that the stage column of the projects table, has array set to "true", which is meant to say that it can accept an array/multiple values instead of a single value. However, in factories.rb, for tests that rely on creating a project first before the test is run, the stage field is populated with a single value instead of an array, which is why we see the 'Malformed array literal" error. To solve this we could EITHER,
- Do away with the enum altogether in the project model file and specify the options for the stage field in the project.rb resource file like we did for the "Sizes" column of the product table, similar to this:
`field :type, as: :select, options: { 'Large container': :large, 'Medium container': :medium, 'Tiny container': :tiny }, display_with_value: true, placeholder: 'Choose the type of the container.'`
- OR we could manipulate the values sent after populating the "Stage" field in the project.rb model file, before saving which i think could allow us to keep the enum declaration.
- Another option would be to run a migration to change the Stage column to a non-array column. Though if my memory serves me right, initially the issue was related to making this particular field a multiple select.

Kindly let me know which approach is you'd prefer and i'll get to solving the error ASAP!

spec/dummy/db/schema.rb
```diff
@@ -173,12 +173,13 @@
...
     t.integer "price_cents", default: 0, null: false
     t.string "price_currency", default: "'USD'::character varying", null: false
+    t.string "sizes", default: [], array: true
   end
 
   create_table "projects", force: :cascade do |t|
     t.string "name"
     t.string "status"
-    t.string "stage"
+    t.string "stage", default: [], array: true
```
Paul-Bob, line 182: @zhephyn let's start by fixing the schema, the `AddSizesToProducts` migration is adding the `sizes` array column to the `products` table however the schema is also changing the `stage` column to array on the `projects` table.

```suggestion
    t.string "stage"
```

Paul-Bob:
> initially the issue was related to making this particular field a multiple select.

The `spec/dummy` directory contains the local application used for development and testing. The issue is about implementing the `multiple` option for select fields. All changes made to `spec/dummy` should be exclusively for testing this feature.

I mentioned that specific field in the video to illustrate how multiple select should work. However, it’s not necessary to modify that exact field or data structure. Since that field is already being used in other tests, it’s preferable to create a new field specifically for testing the multiple select use case, similar to how you created the `sizes` field for products.

Let me know if you have any questions.

spec/dummy/db/schema.rb
Paul-Bob: I'm directly committing this change to speed up the process

zhephyn: Okay no problem. Im not directly in contact with my computer at the moment

zhephyn: @Paul-Bob . Though I didn't quite understand what you referenced, you specified a migration named AddSizesToProducts, yet according to the suggested change, it's like you're changing the stage column to a non array column, and the stage column is part of the projects table not the products table. Could you kindly clarify if this is the case?

zhephyn:
Thanks for the feedback. As for what I meant by the statement, initially what we were aiming for is that for every resource that has a multiple select, whether it's Product or Project, the any field of this resource can use the select option which can allow multiple options to be selected. And so far, we implemented it correctly for the sizes column of the products table. All the tests related to this pass. However, the problem came with the fact that the stage column of the projects table is set to accept array values, yet the Enumerable declaration for the stage in the project model doesn't allow this. This is a missing feature in rails. The enum can't allow multiple values but rather a single value. Which is why those tests failed. And this is why I had suggested as one of the approaches that we could remove the "array: true" option from the stage column by running a migration, this will solve the issue. However, this would defeat the initial suggestion since you had said that we should make changes such that the we can select multiple stages whether two or more for a given project. By removing the "array: true", this wouldn't be possible. Should I create a demo with code and perhaps video to show what I'm referring to approach wise? Please let me know. Thanks for the feedback. I'm very grateful for the learning experience 🙏

Paul-Bob:
@zhephyn, can we hop on a quick call to talk this through?

zhephyn:
> @zhephyn, can we hop on a quick call to talk this through?

Sure. Can we use zoom? Or there's another alternative?

Paul-Bob:
@zhephyn https://meet.google.com/tzw-edpx-zea

zhephyn:
Hey @Paul-Bob . I've seen the updates you pushed to the PR and that all the tests are passing successfully. Does it mean there is no more work to be done on this PR?

Paul-Bob:
There was an issue with how the multiple option was configured on the select helper. I modified it to ensure compatibility with Rails 6.1.

A test is required for the show view, which I fixed in a previous commit, along with documentation updates. Could you please take care of those?

zhephyn:
I'd be glad to. 
A few clarifications though, 
1. For the show view, is a new test to be added to the select field spec? And what are we testing for? I've tried out the feature in the browser and the empty string which was causing issues initially is removed before saving happens.
2. Are there any guidelines regarding how to go about updating the docs, I've never pushed an update to them before?
Thanks

Paul-Bob (approved):
Thank you for this contribution @zhephyn it's awesome!

I done the docs and also committed the missing tests in order to speed up the process and merge this PR

PR #3581, opened 2025-01-13
enhancement: access to the `query` on all action blocks

### LINKED ISSUE #2935: Access to the selected records on actions fields/messages
### Context

Allow to access the selected records when rendering action fields.

From here https://discord.com/channels/740892036978442260/1141725718364045394/1258058534206046269

Related discord thread: https://discord.com/channels/740892036978442260/1141725618694795364/1288893781151584381

## Possible approaches

I see two approaches here:

#### 1. Append the record id's to the action link as params

This is the simpler approach in my opinion.
When a user selects an item, that item's id (or it's identifying param when using `to_param`) should be appended to the URL of every action URL that's present on the page.
Then, in the action controller, we detect those params, find the records and expose them inn the action `def fields` method (where they are missed).

#### 2. Open the action using a POST request

This approach should use a stimulus controller that intercepts the click on the action URL, figures out if the page that the user is on is an Index page and tries to find the selected records.
If it finds selected records, it should capture the id, prevents the browser to follow that link, and initiate a POST request (probably using https://github.com/rails/requestjs-rails) and letting the app to respond with a turbo stream reponse.

### Warning

> [!WARNING]
> We have to ensure the select all feature works as well. so we need to send the encrypted query as a param.

### Misc

We could compress the ids for the param using https://github.com/hopsoft/universalid

Nevelito:
# Description

Fixes #2935

Grants access to the `query` object throughout the action, enabling its use when rendering the action view.

Here is an example with a test action:

```ruby
class Avo::Actions::Test::Query < Avo::BaseAction
  self.name = "Test query access"
  self.message = -> {
    "message #{query.count} selected"
  }
  self.cancel_button_label = -> {
    "cancel_button_label #{query.count} selected"
  }
  self.confirm_button_label = -> {
    "confirm_button_label #{query.count} selected"
  }
  self.standalone = true

  def fields
    field :selected, default: "#{query.count} selected def fields"
  end

  def handle(**args)
    succeed "succeed #{query.count} selected"
  end
end
```

image

# Checklist:

  - https://github.com/avo-hq/docs.avohq.io/pull/333


app/controllers/avo/actions_controller.rb
```diff
@@ -9,9 +9,14 @@ class ActionsController < ApplicationController
...
     end
+    before_action :set_query, only: [:show, :handle], if: ->(request) do
+      # set query will fail in show if user will not select some row
+      # params[:fields] are set during selecting row or in show view
+      request.params[:fields].present?
+    end
     before_action :set_action, only: [:show, :handle]
     before_action :verify_authorization, only: [:show, :handle]
-    before_action :set_query, :set_fields, only: :handle
+    before_action :set_fields, only: [:handle]
```
Paul-Bob, line 19:
```suggestion
    before_action :set_fields, only: :handle
```

app/controllers/avo/actions_controller.rb
```diff
@@ -9,9 +9,14 @@ class ActionsController < ApplicationController
       # set_record will fail if it's tried to be used from the Index page.
       request.params[:id].present?
     end
+    before_action :set_query, only: [:show, :handle], if: ->(request) do
+      # set query will fail in show if user will not select some row
+      # params[:fields] are set during selecting row or in show view
+      request.params[:fields].present?
+    end
```
Paul-Bob, line 16: This approach makes sense for `show`, but we should ensure that the conditional check does not apply to `handle`.

app/controllers/avo/actions_controller.rb
```diff
@@ -65,6 +70,8 @@ def set_query
       resource_ids = action_params[:fields][:avo_resource_ids].split(",")
 
       @query = decrypted_query || (resource_ids.any? ? @resource.find_record(resource_ids, params: params) : [])
+
+      TestBuddy.hi("Query count: #{@query.size}")
```
Paul-Bob, line 74: Ensure that `TestBuddy` messages are triggered exclusively from the dummy app located in `spec/dummy/app`. Otherwise, this could slightly degrade performance across all apps, even though it’s solely for testing purposes on our side.

```suggestion
```

app/javascript/js/controllers/item_select_all_controller.js
```diff
@@ -43,6 +43,10 @@ export default class extends Controller {
       this.selectAllOverlay(allSelected)
       this.resetUnselected()
     }
+
+    const resourceIds = JSON.parse(this.element.dataset.selectedResources).join(',')
+    const allSelectedValue = ''
+    this.updateLinks(resourceIds, allSelectedValue)
```
Paul-Bob, line 49: Let's avoid unnecessary const initialization by giving `allSelected` a default value on the function definition

```suggestion
    this.updateLinks(resourceIds)
```

app/javascript/js/controllers/item_select_all_controller.js
```diff
@@ -51,6 +55,36 @@ export default class extends Controller {
     this.selectedAllValue = !this.selectedAllValue
     this.unselectedMessageTarget.classList.toggle('hidden')
     this.selectedMessageTarget.classList.toggle('hidden')
+
+    const resourceIds = JSON.parse(this.element.dataset.selectedResources).join(',')
```
Paul-Bob, line 59: This line is duplicated, can't we pass the `this.element` to the `updateLinks` and compute the parse logic inside the function?

Paul-Bob (changes requested):
Thanks for working on this, @Nevelito! 

I've manually tested this PR with the following action code, but it seems there's currently no access to the `query` object inside the actions. 
Let's make sure the `query` object is accessible within action messages/fields. 
Additionally, I've left some code review comments, let me know if you have any questions!

```ruby
class Avo::Actions::Sub::DummyAction < Avo::BaseAction
  self.name = -> { "Dummy action" }
  self.message = -> { query.first.name }
  self.cancel_button_label = -> { query.first.name }
  self.confirm_button_label = -> { query.first.name }
  self.standalone = true

  def fields
    field :name, default: query.first.name
  end

  def handle(query:, **args)
    succeed query.first.name
  end
end
```

Paul-Bob (commented):
Thanks for all the work done here @Nevelito it's looking good!

I'll take over from here and adjust the last details before merging.

PR #3594, opened 2025-01-20
Fixed argument prefix warnings

iainbeeston:
# Description

If I run an app using avo with warnings enabled I get a LOT of warnings like this:

    avo-3.16.1/app/components/avo/fields/files_field/edit_component.html.erb:1: warning: `**' interpreted as argument prefix

This is happening because a function is being called with `**` as the first argument but there are no brackets around the arguments. Adding brackets fixes the warning but it does not change the meaning, so I've done that for all field components.

Fixes # (issue)

# Checklist:

## Manual review steps

1. Everything should work as before (this shouldn't change any behavior)

### DIFF
--- a/app/components/avo/fields/area_field/edit_component.html.erb
+++ b/app/components/avo/fields/area_field/edit_component.html.erb
@@ -1,4 +1,4 @@
-<%= field_wrapper **field_wrapper_args do %>
+<%= field_wrapper(**field_wrapper_args) do %>
   <%= @form.text_field field.id,
     value: field.value.to_s,
     class: classes("w-full"),
--- a/app/components/avo/fields/area_field/show_component.html.erb
+++ b/app/components/avo/fields/area_field/show_component.html.erb
@@ -1,4 +1,4 @@
-<%= field_wrapper **field_wrapper_args do %>
+<%= field_wrapper(**field_wrapper_args) do %>
   <% if field.value.present? %>
     <%= area_map field.map_data, **field.mapkick_options %>
   <% else %>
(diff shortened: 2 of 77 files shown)


Paul-Bob:
Hi @iainbeeston thanks for this contribution.

> If I run an app using avo with warnings enabled I get a LOT of warnings like this:

What warnings are you enabling to see this logs?

iainbeeston:
@Paul-Bob Sorry I'm talking about warnings coming from ruby itself. You can enable that by either:

* running ruby with the `-w` flag
* running rspec with the `--warnings` flag
* Setting `$VERBOSE=true` (in a rails app this can be done in your `puma.rb` config file or anywhere called during startup)

In rspec the errors appear in the console, in rails they appear in the server logs when you navigate to any avo resource page.

I'm using ruby 3.3.2 and looking at the source code for ruby it looks like this warning was added in ruby 3.3

Paul-Bob:
Thanks for sharing those @iainbeeston 

Is there any rule that we can add to our linter to avoid this kind of warning in the future?

iainbeeston:
That's a really good question... ideally there would be a rubocop cop about this but I've looked and I can't see one (except always requiring parentheses). Even if there was a rubocop cop you'd also need to be using ere-lint because these warnings are coming from view files (I'm not sure if you are using ere-lint or not?)

Paul-Bob (approved):
Thanks for this contribution @iainbeeston!

adrianthedev:
Has anyone try [`Dangerfile`](https://danger.systems/ruby/) before?

PR #3620, opened 2025-01-30
enhancement: add `saveButton` stimulus target to save button

### LINKED ISSUE #3607: Custom stimulus controller and exposed target on resource form submit button
### Feature
Ability to apply resource-defined stimulus controllers on built-in form buttons and get them as Stimulus targets.

For example:
The "Save" button has `data-controller="loading-button"`. It would be great to have the same logic that already exists for inputs — apply a resource controller and expose the button as a Stimulus target. 

Expected behavior:
```ruby
class Avo::Resources::User < Avo::BaseResource
  self.stimulus_controllers = %w[user-form-autofill]
end
```
The new/edit page of user resource has a "Save" button with such attributes:
```html
<button name="button" type="submit" loading="true" data-controller="loading-button user-form-autofill" data-action="click->loading-button#attemptSubmit" data-user-form-autofill-target="resourceSubmitButton">...
```
(or the controller is applied at the resource level instead of the button)

### Current workarounds
Match it using CSS selector:

Submit button: `document.querySelector('div[data-panel-id="main"] button[type="submit"]')`

I mean this button on the resource new/edit page.
Image

### Additional context
Why I need that: I have dynamic data loading in the edit form, and I want to disable the submit button until the data are loaded and the user makes a selection.
Ideally, I wish it would work for custom actions on the edit page too, but in my particular case, it's not needed.

------
The documentation states that it's possible to target the "button" element.

https://github.com/avo-hq/docs.avohq.io/blob/c676288f198217a1f7a7222370b0de95972f2755/docs/3.0/stimulus-integration.md?plain=1#L77C255-L77C312

But I didn't find any button which exposed as Stimulus target. So, probably it's a case of updating docs too.

Paul-Bob:
# Description

Fixes #3607

Adding the `saveButton` target to all the stimulus controllers registered on a resource.

# Checklist:

### DIFF
--- a/app/components/avo/resource_component.rb
+++ b/app/components/avo/resource_component.rb
@@ -1,6 +1,7 @@
 class Avo::ResourceComponent < Avo::BaseComponent
   include Avo::Concerns::ChecksAssocAuthorization
   include Avo::Concerns::RequestMethods
+  include Avo::Concerns::HasResourceStimulusControllers
 
   attr_reader :fields_by_panel
   attr_reader :has_one_panels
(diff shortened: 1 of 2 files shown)


adrianthedev (approved):
Looks good. Haven't manually tested it.

PR #3635, opened 2025-02-03
Removed defer attribute from the avo.base.css stylesheet tag

iainbeeston:
# Description

When you load a page in avo the html `<link>` tag to load `avo.base.css` it ha a `defer` attribute on it. `defer` isn't a valid attribute for a `<link>` tag - it is for `<script>` tags and having it here is invalid html. I think this is also causing loading issues in google chrome for me when I redirect from avo to my main app.

I've removed defer from the `<link>` tag. This should make the html valid and shouldn't have any bad effects (as it shouldn't be used on a `<link>` tag anyway).

Fixes # (issue)

# Checklist:

## Manual review steps

1. Check that the css for avo still loads correctly.


Paul-Bob (approved):
LGTM

Thank you @iainbeeston

PR #3639, opened 2025-02-05
Fix pagy nav typo

julianrubisch:
# Description
Small typo fix that leads to an AttributeError (`unknown keyword: :xanchor_string`) in the Media Library


Paul-Bob (approved):
LGTM!

Thanks for noticing and contributing this fix @julianrubisch !

Issue #3536, opened 2024-12-16
Detect polymorphic associations in generator


adrianthedev:
### Feature

If a user generates a resource that has polymorphic associations then they will automatically get a error from Avo because actually it's from rails because it does not support to compute the polymorphic association. We should detect if we find an `_type` and `_id` attributes and try to make that polymorphic or we can generate it but leave it commented out if we can't compute to which classes it's polymorphic to. 

This happens in when they have invited by and invited the attributes.

Nevelito:
For assignment

Paul-Bob:
Implemented on #3645

PR #3645, opened 2025-02-06
enhancement: Detect polymorphic associations in generator

Nevelito:
# Description

Changes:
- implement detect_polymorphic_associations to resource_ganerator
- add specs

Fixes # (issue)
https://github.com/avo-hq/avo/issues/3536

# Checklist:


lib/generators/avo/resource_generator.rb
```diff
@@ -25,6 +25,9 @@ class ResourceGenerator < NamedBaseGenerator
       def create
         return if override_controller?
 
+        # Detect polymorphic associations
+        detect_polymorphic_associations
+
         template "resource/resource.tt", "app/avo/resources/#{resource_name}.rb"
         invoke "avo:controller", [resource_name], options
       end
```
Paul-Bob, line 33: I've noticed that all fields are being generated within the `generate_fields` method. Why, then, is `detect_polymorphic_associations` being called separately instead of alongside the other methods inside `generate_fields`? Also, it seems that we're calling it without ever using the return value.

lib/generators/avo/resource_generator.rb
```diff
@@ -266,13 +280,21 @@ def generated_fields_template
           fields_string = ""
 
           fields.each do |field_name, field_options|
-            # if field_options are not available (likely a missing resource for an association), skip the field
```
Paul-Bob, line 269: Why is this comment being removed? Is it outdated in some way?

lib/generators/avo/resource_generator.rb
```diff
@@ -266,13 +280,21 @@ def generated_fields_template
...
+              field_options[:options].each { |k, v| options += ", #{k}: #{v}" }
+            end
 
-            fields_string += "\n    #{field_string field_name, field_options[:field], options}"
+            # Add a comment for polymorphic fields
+            if field_options[:field] == "polymorphic"
+              fields_string += "\n    # Polymorphic association detected for :#{field_name}"
+              fields_string += "\n    # Uncomment and configure the possible types:"
+              fields_string += "\n    # field :#{field_name}, as: :polymorphic, types: [User, Admin]"
+            else
```
Paul-Bob, line 295: There isn't a field type `polymorphic`

[Here](https://github.com/avo-hq/avo/blob/main/spec/dummy/app/avo/resources/comment.rb#L22) is an example of a `belongs_to` field for a polymorphic association.

Paul-Bob (changes requested):
Thanks for working on this, @Nevelito!  

I have a few questions:

lib/generators/avo/resource_generator.rb
```diff
@@ -185,14 +189,28 @@ def fields_from_model_tags
...
           associations.each do |name, association|
-            fields[name] = if association.is_a? ActiveRecord::Reflection::ThroughReflection
-              field_from_through_association(association)
-            else
-              ::Avo::Mappings::ASSOCIATIONS_MAPPING[association.class]
-            end
+            fields[name] =
+              if association.polymorphic?
+                types = find_polymorphic_association_types(association)
+                {field: "polymorphic", options: {types: types}}
```
Paul-Bob, line 195: Can we use `field: "belongs_to"` and add the `polymorphic_as` inside the `options`?

Paul-Bob (changes requested):
We're getting close to merge this @Nevelito 

I let some comments about reusing the code we have instead creating a new method for the polymorphic string constructor.

lib/generators/avo/resource_generator.rb
```diff
@@ -272,7 +290,12 @@ def generated_fields_template
             options = ""
             field_options[:options].each { |k, v| options += ", #{k}: #{v}" } if field_options[:options].present?
 
-            fields_string += "\n    #{field_string field_name, field_options[:field], options}"
+            fields_string +=
+              if field_options[:field] == "polymorphic"
+                "\n    #{polymorphic_field_string field_name, options}"
+              else
+                "\n    #{field_string field_name, field_options[:field], options}"
+              end
```
Paul-Bob, line 298: If we implement https://github.com/avo-hq/avo/pull/3645/files#r1950414758 we can remove `polymorphic_field_string`and reuse `field_string`.

spec/features/avo/generators/resource_generator_spec.rb
```diff
@@ -46,6 +46,23 @@
...
+    it "generates commented polymorphic fields" do
+      files = [
+        Rails.root.join("app", "avo", "resources", "review.rb").to_s,
+        Rails.root.join("app", "controllers", "avo", "reviews_controller.rb").to_s
+      ]
+
+      keeping_original_files(files) do
+        Rails::Generators.invoke("avo:resource", ["review", "-q", "-s"], {destination_root: Rails.root})
+
+        expect(File.read(files[0])).to include("field :reviewable, as: :belongs_to")
```
Paul-Bob, line 60: Lets add the types into this test

lib/generators/avo/resource_generator.rb
```diff
@@ -185,14 +185,35 @@ def fields_from_model_tags
...
         end
 
+        def field_with_polymorphic_association(association)
+          Rails.application.eager_load!
+
+          possible_types = ActiveRecord::Base.descendants.select do |model|
+            model.reflect_on_all_associations(:has_many).any? { |assoc| assoc.options[:as] == association.name }
+          end.map(&:name)
+
+          types = possible_types.empty? ? "[] # Program couldn't compute types" : possible_types.map(&:to_sym)
```
Paul-Bob, line 206:
```suggestion
          types = possible_types.empty? ? "[] #  Types weren't computed correctly. Please configure them." : possible_types.map(&:to_sym)
```

Paul-Bob (approved):
This PR is looking good @Nevelito

Let's just change the message when types are empty and add the types on the test, otherwise is good to merge.

spec/features/avo/generators/resource_generator_spec.rb
Paul-Bob: Also the polymorphic_as

lib/generators/avo/resource_generator.rb
```diff
@@ -185,11 +185,34 @@ def fields_from_model_tags
...
+
+        def field_with_polymorphic_association(association)
+          Rails.application.eager_load! unless Rails.application.config.eager_load
+
+          types = polymorphic_association_types(association)
+
+          {
+            field: "belongs_to",
+            options: {
+              polymorphic_as: association.name,
```
Paul-Bob, line 207:
```suggestion
              polymorphic_as: :"#{association.name}",
```

Nevelito:
I think reduce 3 lines in resource_generator is impossible, we can reduce one here and move first if to first line:

```
fields[name] =
              if association.polymorphic?
                field_with_polymorphic_association(association)
              elsif association.is_a?(ActiveRecord::Reflection::ThroughReflection)
                field_from_through_association(association)
              else
                ::Avo::Mappings::ASSOCIATIONS_MAPPING[association.class]
              end
```

lib/generators/avo/resource_generator.rb
```diff
@@ -185,11 +185,34 @@ def fields_from_model_tags
...
+            options: {
+              polymorphic_as: :"#{association.name}",
+              types: types.presence || "[] # Types weren't computed correctly. Please configure them."
+            }
+          }
+        end
+
+        def polymorphic_association_types(association)
+          ActiveRecord::Base.descendants.filter_map do |model|
+            model.name.to_sym if model.reflect_on_all_associations(:has_many).any? { |assoc| assoc.options[:as] == association.name }
```
Paul-Bob, line 215:
```suggestion
            model.name if model.reflect_on_all_associations(:has_many).any? { |assoc| assoc.options[:as] == association.name }
```

spec/features/avo/generators/resource_generator_spec.rb
```diff
@@ -46,6 +46,28 @@
...
+      keeping_original_files(files) do
+        Rails::Generators.invoke("avo:resource", ["review", "-q", "-s"], {destination_root: Rails.root})
+
+        # Types load in different order every time
+        expect(File.read(files[0])).to include("field :reviewable, as: :belongs_to, polymorphic_as: :reviewable, types:")
+        expect(File.read(files[0])).to include(":Team")
+        expect(File.read(files[0])).to include(":Project")
+        expect(File.read(files[0])).to include(":Post")
+        expect(File.read(files[0])).to include(":Fish")
+
```
Paul-Bob, line 66:
```suggestion
        expect(File.read(files[0])).to include("Team")
        expect(File.read(files[0])).to include("Project")
        expect(File.read(files[0])).to include("Post")
        expect(File.read(files[0])).to include("Fish")

```

Paul-Bob:
@Nevelito, I noticed that the generated types aren't constants.

They should be formatted like this: `types: [Team, Project, Post, Fish]`

Alternatively, `types: [::Team, ::Project, ::Post, ::Fish]` is also valid.

However, my last suggestion mistakenly converted `types: [:Team, :Project, :Post, :Fish]` into `types: ["Team", "Project", "Post", "Fish"]`, both of which are incorrect.

Nevelito:
I tried to use constantize witch should fix this issue but it returns types like this

```
types: [Team(id: integer, name: string, description: text, created_at: datetime, updated_at: datetime, url: string, color: string), Project(id: inte ...
```

Paul-Bob:
> I tried to use constantize witch should fix this issue but it returns types like this
> 
> ```
> types: [Team(id: integer, name: string, description: text, created_at: datetime, updated_at: datetime, url: string, color: string), Project(id: inte ...
> ```

I know, I don't know a easy way to achieve this, could you do some research on this please?

Nevelito:
I created alias with new inspect function and keep original inspect in function original_inspect, now it works as it should but let me know what you think

Paul-Bob:
> I created alias with new inspect function and keep original inspect in function original_inspect, now it works as it should but let me know what you think

@Nevelito you did that change on the `spec/dummy/app/models/application_record.rb` which is inside Avo's dummy application. What will happen on user's applications that will not have this monkey patch? Most provably will not work as expected and the types will be rendered as: `types: [Team(id: integer, name: string, description: text, created_at: datetime, updated_at: datetime, url: string, color: string), Project(id: inte ...` which is not ideal. We want this way of rendering types: `types: [Team, Project, Post, Fish]` on all the applications

Suggesting this change to all the users is not ideal and monkey patching the `ActiveRecord::Base` it's also not reasonable. 

Let's try to achieve this output: `types: [Team, Project, Post, Fish]`  without monkey-patches if possible

Nevelito:
I am not sure if it is possible, I tried to find something but nothing works, only changing inspect gave us wanted result but I see that we can not make like that because of this is in dummy app. I not sure what can we do now.

lib/generators/avo/resource_generator.rb
```diff
@@ -316,5 +316,15 @@ def field(name, type)
...
+
+    class Inspector
+      attr_accessor :name
+      def initialize(name)
+        @name = name
+      end
+      def inspect
+        name
+      end
+    end
```
Paul-Bob:
```suggestion
    class Inspector
      def initialize(name)
        @name = name
      end

      def inspect = @name
    end
```

Paul-Bob:
Thanks you for this contribution @Nevelito !

Issue #3413, opened 2024-11-12
Non-standard id via `primary_key` option on Belongs To field not respected


therabidbanana:
### Describe the bug

We have a table that has an integer id (deprecated) and a uuid used for all new tables. This is easy to define in the activerecord relationship via `primary_key`:

```
belongs_to :user, primary_key: :uuid
```

But the Avo belongs to field for this relationship still tries to save using the `#id` method. Depending on if the id field exists on the table this will either generate a NoMethodError or, in our case since we actually have both, using an invalid integer type id, silently failing. 

### Steps to Reproduce

1. Add uuid identifier column to existing table with integer ids. Define relationship using this new column.
2. Configure a `field :user, as: :belongs_to`
3. Try to edit the newly added field via Avo

### Expected behavior & Actual behavior

Observe `id` param is sent rather than `uuid`

### Models and resource files

```ruby
# == Schema Information
#
# Table name: posts
#
#  id                :uuid             not null, primary key
#  title             :string
# user_id :uuid
#
# Indexes
#
#  index_posts_on_user_id  (user_id)
#
class Post
  belongs_to :user, primary_key: :uuid, optional: true
end
```

```ruby
# == Schema Information
#
# Table name: users
#
#  id                     :bigint(8)        not null, primary key
#  name                   :string           not null
#  uuid                   :uuid             not null
#
# Indexes
#
#  index_users_on_uuid                  (uuid) UNIQUE
class User
end
```

```
class Avo::Resources::Post
  def fields
    field :title, as: :text
    field :user, as: :belongs_to
  end
end
```

### System configuration
**Avo version**:
= 3.13.7

**Rails version**:
= 7.2.1.2

**Ruby version**:

**License type**:

 - [ ] Community
 - [ ] Pro
 - [x] Advanced

### Additional context

My best guess following code path is that it looks like id is assumed to the the primary key here, but reflections should be used like they are for the foreign key option:

https://github.com/avo-hq/avo/blob/main/lib/avo/fields/belongs_to_field.rb#L219

#### Impact

 - [ ] High impact (It makes my app un-usable.)
 - [x] Medium impact (I'm annoyed, but I'll live.)
 - [ ] Low impact (It's really a tiny thing that I could live with.)

#### Urgency

 - [ ] High urgency (I can't continue development without it.)
 - [ ] Medium urgency (I found a workaround, but I'd love to have it fixed.)
 - [x] Low urgency (It can wait. I just wanted you to know about it.)

[...]

PR #3665, opened 2025-02-13
fix: belongs_to field should use association primary_key

Nevelito:
# Description

If record had primary_key set, it was still using **:id**

**Changes:**
- Change fill_field function in belongs_to_firld.rb
- Add primary_key function
- Add specs

Fixes # (issue)
Issue: https://github.com/avo-hq/avo/issues/3413

# Checklist:


Paul-Bob (changes requested):
Thanks for looking into this, @Nevelito!

I tested this PR using the reproduction repo, and it does seem to resolve the issue. However, the test coverage doesn't currently reproduce the problem. Would it be possible to add a `belongs_to` relationship that uses `primary_key` in the dummy app and write a test against that scenario?

Nevelito:
```
FactoryBot.define do
  factory :user do
    uuid { SecureRandom.uuid }
    first_name { Faker::Name.first_name }
    last_name { Faker::Name.last_name }
    email { Faker::Internet.email }
    password { Faker::Internet.password }
    roles { {admin: false, manager: [true, false].sample, writer: [true, false].sample} }
    birthday { Faker::Date.birthday(min_age: 18, max_age: 65) }
    custom_css { ".header {\n  color: red;\n}" }
  end
  
  ```
  
  schema:
  
  ```
  create_table "users", force: :cascade do |t|
    t.string "email", default: "", null: false
    t.string "first_name"
    t.string "last_name"
    t.json "roles"
    t.date "birthday"
    t.text "custom_css"
    t.bigint "team_id"
    t.string "encrypted_password", default: "", null: false
    t.string "reset_password_token"
    t.datetime "reset_password_sent_at", precision: nil
    t.datetime "remember_created_at", precision: nil
    t.datetime "created_at", null: false
    t.datetime "updated_at", null: false
    t.boolean "active", default: true
    t.string "slug"
    t.string "uuid", null: false
    t.index ["email"], name: "index_users_on_email", unique: true
    t.index ["reset_password_token"], name: "index_users_on_reset_password_token", unique: true
    t.index ["slug"], name: "index_users_on_slug", unique: true
    t.index ["team_id"], name: "index_users_on_team_id"
    t.index ["uuid"], name: "index_users_on_uuid", unique: true
  end
  ```
  
  
image



**Do you have any ideas why it is happening ? I could not find any info in Internet.**

**Comment by Paul-Bob** [comment]
> I pushed changes, have a look in free time. I will try to fix it tomorrow one more time.

The `User` and `Post` models are heavily used in many tests. If you notice that this change breaks existing tests, I suggest creating the test scenario with other models that are less burdened, such as `Event`, or even a new model if necessary.

**Comment by Nevelito** [comment]
> > I pushed changes, have a look in free time. I will try to fix it tomorrow one more time.
> 
> The `User` and `Post` models are heavily used in many tests. If you notice that this change breaks existing tests, I suggest creating the test scenario with other models that are less burdened, such as `Event`, or even a new model if necessary.
> 
> Let's ensure that all the tests run successfully.


Hi, I’m still having some issues with testing. Could you take a look and maybe suggest something I could try?

File: lib/avo/fields/belongs_to_field.rb [hunk 1]
```diff
@@ -126,6 +133,7 @@ def values_for_type(model = nil)
         end
 
         query.all.limit(Avo.configuration.associations_lookup_list_limit).map do |record|
+          # to_param uses slug so checking primary_key is unnecessary
```
Paul-Bob, line 136:
```suggestion
```

Comments are important, but I’m not sure how this one is relevant. Do you think it adds value or highlights something worth remembering? If so, could you elaborate on it? Thanks!

Paul-Bob (changes requested):
Thanks for looking into this, @Nevelito !

Everything looks clean, and the setup appears to be correct. I’ve left some comments on the code, let’s continue the discussion there.

spec/features/avo/belongs_to_spec.rb
```diff
@@ -159,6 +159,16 @@
...
 
+  describe "with custom primary key set" do
+    let(:event) { create(:event, name: "Sample Event") }
+
+    let!(:volunteer) { create(:volunteer, event: event) }
+
+    it "displays the event link using the UUID and stores the correct foreign key" do
+      expect(volunteer.event_uuid).to eq(event.uuid)
+    end
+  end
```
Paul-Bob, line 170: This test is not effectively validating the changes. If we use the same setup on `main` without the logic changes to the `belongs_to` field, it still runs successfully without any failures.

Let's correctly set the volunteer with `field :event, as: :belongs_to` and write a test that ensures a volunteer is created, during creation it selects an event, and is correctly associated with that event after creation.

lib/avo/fields/belongs_to_field.rb
```diff
@@ -212,14 +220,16 @@ def fill_field(record, key, value, params)
           if valid_model_class.blank? || id_from_param.blank?
             record.send(:"#{polymorphic_as}_id=", nil)
           else
-            record_id = target_resource(record:, polymorphic_model_class: value.safe_constantize).find_record(id_from_param).id
+            found_record = target_resource(record:, polymorphic_model_class: value.safe_constantize).find_record(id_from_param)
+
+            record_id = found_record&.send(primary_key.presence || :id)
```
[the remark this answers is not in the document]
Paul-Bob: Addressed here https://github.com/avo-hq/avo/pull/3665/commits/912c9fdd6449b3ed0aa73af3dcf31346b30b6d29

Paul-Bob (approved):
Thanks for all your effort here @Nevelito!

PR #3683, opened 2025-02-21
fix: remove source maps for prod build

veelenga:
# Description

Source maps are useful for debugging by linking compiled code back to its original source but pose security risks when deployed in production. 

# Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "build:custom-js": "esbuild spec/dummy/app/javascript/*.js --bundle --sourcemap --minify --outdir=app/assets/builds",
     "build:css": "tailwindcss -i ./app/assets/stylesheets/avo.base.css -o ./app/assets/builds/avo.base.css --postcss",
     "prod:build": "yarn prod:build:js && yarn prod:build:css",
-    "prod:build:js": "esbuild app/javascript/*.js --bundle --sourcemap --minify --outdir=public/avo-assets",
+    "prod:build:js": "esbuild app/javascript/*.js --bundle --minify --outdir=public/avo-assets",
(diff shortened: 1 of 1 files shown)


adrianthedev:
Hey @veelenga.

I understand why you send this PR.

The thing is that the files are still being pushed to the server. If there's a bad actor they can unminify the files quite easily using an LLM (they are pretty good at that).

Removing sourcemaps will also make it more difficult to debug issues.

We're not going to merge this right now.

Thanks for understanding.

veelenga:
Thanks for the clarification.

Do you have any recommendations on how to prevent Avo's sourcemap from being included in our final product? I understand the value of sourcemaps for debugging within your own applications, but it doesn’t have any sense for them to be part of an app that you, as the Avo developer, don’t have access to debug.

veelenga:
@adrianthedev just to clarify, this is reported as a part of penetration testing, which makes sense to me because it violates the security practices. 

I tried different ways on my own to disable Avo's sourcemaps during the assets compilation steps, but it looks very hacky. 

Would you guys be able to provide any recommendation how it can be disabled?

adrianthedev:
Hey @veelenga.

We didn't take that decision by ourselves.
We had a conversation with our [security partners](https://greenhats.com/) and they concurred.
They gave us a few reasons why they advised against this change:
1. The code can be unminified quite easily
2. The code is open source (so un-minified from the start)
3. This will make it more difficult to debug code for the rest of our users

That being said, you still can have your source maps removed from your Avo apps.
The approach is to have a rake task run after the asset precompile step (or another step that better suits your deployment tactic), go into your app (probably `Rails.root_path.join('public', 'avo-assets', FIND_THE_SOURCEMAP_FILE)`) and remove it.
This task should automatically run on every deploy.

Let us know if you need more assistance with this.

Issue #3596, opened 2025-01-21
fields :title and :description to allow lambda


jetienne:
### Feature

When defining a `:has_many` field, it would be great to allow lambdas for `:name` and `:description` attribute. We could then display dynamic content like:

```ruby
field :sent_invoices, 
         as: :has_many, 
         scope: lambda { query.sent }, 
         name: 'Sent Invoices', 
         description: lambda { "Total of #{query.sent.sum(:amount)}" }
```

marlena-b:
I would like to give it a try. Could you please assign me to it?

Paul-Bob:
Hi @jetienne,  

How would you use `name` as a lambda? Is it essential to include `query` there?  
I'm asking because, in the first iteration, we might not have access to `query` inside `name`.  

For `description`, however, `query` will definitely be available.

Paul-Bob:
> [@Paul-Bob](https://github.com/Paul-Bob) `query` seems to be the main dynamic content for my use cases

How would you use the `query` when defining the field name?

jetienne:
it could be "{query.count} Invoices" for instance

PR #3693, opened 2025-02-25
enhancement: allow dynamic content in `description` and `name`

marlena-b:
# Description

Fixes https://github.com/avo-hq/avo/issues/3596

I'm not sure about this solution as there is a lot of metaprogramming that I don't fully understand. 
The description already supported lambdas so I just extended it so it now has access to `query`. I tried to do the same for `name` but it cannot have access to the query as it is also used to display the text of the tab (see screenshot). When rendering these tabs we don't have access to the query.

I appreciate the feedback, and if this is not the correct way to solve this I don't mind closing this PR :) 

# Checklist:

- [X] I have made corresponding changes to the [documentation](https://github.com/avo-hq/avodocs)

## Screenshots & recording

Zrzut ekranu 2025-02-25 o 18 50 10

## Manual review steps

1. Start the dummy app
2. Open any user
3. Notice the description of Posts relationship has a counter
4. Notice that we use lambda in code to display the name of the relationship

### DIFF
--- a/lib/avo/fields/base_field.rb
+++ b/lib/avo/fields/base_field.rb
@@ -116,7 +116,12 @@ module Avo
       # Secondly we'll try to find a translation key
       # We'll fallback to humanizing the id
       def name
-        return @name if custom_name?
+        if custom_name?
+          return Avo::ExecutionContext.new(
+            target: @name,
+            **name_attributes
+          ).handle
+        end
 
         if translation_key
           translated_name default: default_name
@@ -316,6 +321,11 @@ module Avo
         resource = Avo.resource_manager.get_resource_by_model_class(model_class)
 
         resource || (raise Avo::MissingResourceError.new(model_class, id))
+      end
+
+      # Override this method to add custom attributes to the name execution context.
+      def name_attributes
+        {}
       end
     end
   end
--- a/lib/avo/fields/base_field.rb
+++ b/lib/avo/fields/base_field.rb
@@ -117,7 +117,7 @@ module Avo
       # We'll fallback to humanizing the id
       def name
         if custom_name?
-          return Avo::ExecutionContext.new(target: @name).handle
+          Avo::ExecutionContext.new(target: @name).handle
         elsif translation_key
           translated_name default: default_name
         else
--- a/lib/avo/fields/base_field.rb
+++ b/lib/avo/fields/base_field.rb
@@ -117,13 +117,8 @@ module Avo
       # We'll fallback to humanizing the id
       def name
         if custom_name?
-          return Avo::ExecutionContext.new(
-            target: @name,
-            **name_attributes
-          ).handle
-        end
-
-        if translation_key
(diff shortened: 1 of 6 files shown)


lib/avo/fields/base_field.rb
```diff
@@ -116,7 +116,12 @@ def translated_plural_name(default:)
       # Secondly we'll try to find a translation key
       # We'll fallback to humanizing the id
       def name
-        return @name if custom_name?
+        if custom_name?
+          return Avo::ExecutionContext.new(
+            target: @name,
+            **name_attributes
+          ).handle
+        end
```
Paul-Bob, line 124: Do we currently have any use for `name_attributes`? If not, we should consider removing it

```ruby
def name
  if custom_name?
    Avo::ExecutionContext.new(target: @name).handle
  elsif translation_key
    translated_name default: default_name
  else
    default_name
  end
end
```

Paul-Bob (changes requested):
Thank you for this contribution @marlena-b it's already looking great!

> I tried to do the same for name but it cannot have access to the query as it is also used to display the text of the tab (see screenshot).

I understand the challenge. Let's keep it without `query` access for now, and we can iterate on it later if needed. 

Let's focus on the `name_attributes` comment.

Thank you!

lib/avo/fields/base_field.rb
marlena-b: I wanted to keep it consistent with how the description attribute is [implemented](https://github.com/avo-hq/avo/blob/701c29182e40f543fc99d983e24ab0d513a8f63a/lib/avo/concerns/has_description.rb#L18). But I also don't see how this could be used, so I will remove it.

Paul-Bob: Got it.  

`Avo::Concerns::HasDescription` is included in multiple classes, some within the `avo` repository and others across private repositories.  

Each class that includes `Avo::Concerns::HasDescription` can define its own `description_attributes`. When `description` is called, `description_attributes` will vary based on the class from which it is invoked.  

For `name_attributes`, however, we currently don’t have a similar use case.

lib/avo/fields/base_field.rb
```diff
@@ -116,9 +116,9 @@ def translated_plural_name(default:)
       # Secondly we'll try to find a translation key
       # We'll fallback to humanizing the id
       def name
-        return @name if custom_name?
-
-        if translation_key
+        if custom_name?
+          return Avo::ExecutionContext.new(target: @name).handle
```
Paul-Bob, line 120:
```suggestion
          Avo::ExecutionContext.new(target: @name).handle
```

We don’t need an explicit `return` since everything is already wrapped within the `if/else` condition.

Paul-Bob (approved):
I just added a small tweak suggestion, otherwise, it's ready for documentation and merge!  

Thanks again, @marlena-b!

marlena-b:
I added docs: https://github.com/avo-hq/docs.avohq.io/pull/354

Issue #3679, opened 2025-02-21
Bulk update back-end


adrianthedev:
### Context

We'd like to enable users to bulk-update records from the `Index` page.
We need a back-end for it.

### Approach

Create a `BulkUpdateController` which will take in a payload and update those records accordingly.

- execute field validation from the model (Rails should do it for us)
- accepted params should be verified
  - fields marked as `disabled` should not be updated
- ensure the values are updated through `update_using` for every field

Possible payload this controller action can take:

```ruby
{
  payload: [
    {
      param: "user_12o3211o3i3h" # or just the id
      resource_type: Avo::Resources::User,
      body: {
        first_name: "John",
        age: 33
      }
    },
    {
      param: "user_k3k6no2i" # or just the id
      resource_type: Avo::Resources::User,
      body: {
        first_name: "John",
        age: 33
      }
    },
    {
      param: 123
      resource_type: Avo::Resources::Post,
      body: {
        title: "What a great day",
        body: "# Lorem ipsum"
        user_id: "user_12o3211o3i3h"
      }
    }
  ]
}
```

Nevelito:
Comment for assignment, I will give it a try.

adrianthedev:
Some inspiration:

- https://x.com/bradgessler/status/1887187002459877425
- https://superails.com/posts/ruby-on-rails-86-bulk-actions-edit-multiple-selected-records

PR #3695, opened 2025-02-26
feature:  add bulk update

Nevelito:
# Description

**Add bulk update to app**

**Changes:**

1. create new controller
2. adjust the code in edit resource component to handle normal edit and update and bulk edit
3. make changes in base controller to handle bulk update and fix codeclimat errors
4. add new routes
5. add new function to url_halpers.rb to handle bulk update
6. adjust item_select_all_controller.js to handle adding resource ids to Bulk update button
7. add locales
8. add system specs

Fixes # (issue)<br>[https://github.com/avo-hq/avo/issues/3679](<https://github.com/avo-hq/avo/issues/3679>)

# Checklist:

- [ ] I have made corresponding changes to the [documentation](<https://github.com/avo-hq/avodocs>)

## Screenshots & recording

[c4112a6b-a05a-474b-a1a6-25ea3faf24c1](https://uploads.linear.app/bbbbb234-e734-4a7a-929b-5115c5c8399e/1b27686e-7dbc-4c5f-bd4f-7413b4c816e1/c4112a6b-a05a-474b-a1a6-25ea3faf24c1?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2JiYmJiMjM0LWU3MzQtNGE3YS05MjliLTUxMTVjNWM4Mzk5ZS8xYjI3Njg2ZS03ZGJjLTRjNWYtYmQ0Zi03NDEzYjRjODE2ZTEvYzQxMTJhNmItYTA1YS00NzRiLWExYTYtMjVlYTNmYWYyNGMxIiwiaWF0IjoxNzg0ODA5MTQ2LCJleHAiOjE4MTYzNzk3MDZ9.i7Vq781eeAAngE_N9UPm9v1oKcObVzScwIK4S5yFg6g)


Nevelito:
@Paul-Bob I’ve tested this task and it works correctly on the backend. Now we just need to take care of the frontend part. Do you have any suggestions ?

app/components/avo/views/resource_edit_component.html.erb
```diff
@@ -9,7 +9,7 @@
     **@resource.stimulus_data_attributes
   } do %>
   <%= render_cards_component %>
-  <%= form_with model: @resource.record,
+  <%= form_with model: model,
```
Paul-Bob, line 12: Why are we extracting `@resource.record` to the `model` method?

app/components/avo/views/resource_edit_component.html.erb
```diff
@@ -23,6 +23,13 @@
     },
     multipart: true do |form| %>
     <%= render Avo::ReferrerParamsComponent.new back_path: back_path %>
+
+    <% if @prefilled_fields.present? %>
+      <% @prefilled_fields.each do |field, value| %>
+        <%= hidden_field_tag "prefilled[#{field}]", value %>
+      <% end %>
+    <% end %>
```
Paul-Bob, line 31: Why do we need this?

I imagined that each field would return the correct value when pre-filed

app/components/avo/views/resource_edit_component.rb
```diff
@@ -9,6 +9,14 @@ class Avo::Views::ResourceEditComponent < Avo::ResourceComponent
   prop :view, default: Avo::ViewInquirer.new(:edit).freeze
   prop :display_breadcrumbs, default: true, reader: :public
 
+  attr_reader :query
+
+  def initialize(resource:, query: nil, prefilled_fields: nil, **args)
+    @query = query
+    @prefilled_fields = prefilled_fields
+    super(resource: resource, **args)
+  end
```
Paul-Bob, line 18:
```suggestion
  prop :query
  prop :prefilled_fields
```

Lets use the props instead initializer

app/components/avo/views/resource_edit_component.rb
```diff
@@ -18,15 +26,13 @@ def title
...
-
-    return resource_view_path if via_resource?
-    return resources_path if via_index?
-
-    if is_edit? && Avo.configuration.resource_default_view.show? # via resource show or edit page
-      return helpers.resource_path(record: @resource.record, resource: @resource, **keep_referrer_params)
-    end
+    return back_path_from_via_resource if via_resource?
+    return back_path_from_via_index if via_index?
+    return back_path_from_edit if is_edit? && Avo.configuration.resource_default_view.show?
```
Paul-Bob, line 35: Why are we doing this refactor?

If this refactor was introduced due to CodeClimate warnings, let's ignore them for now and revert the change.

For feature PRs, it’s important to keep changes strictly relevant to the feature itself.

If refactor opportunities are identified during development, they should be proposed in a separate issue and addressed in a dedicated PR.

This practice helps maintain clear version history, simplifies tracking changes, and makes reverting specific work much easier if needed.

app/components/avo/views/resource_edit_component.rb
```diff
@@ -71,13 +97,19 @@ def is_edit?
   end
 
   def form_method
-    return :put if is_edit?
+    return :put if is_edit? && params[:controller] != "avo/bulk_update"
```
Paul-Bob, line 100: I think we should use the put as well for bulk updates, any argument why we should use post?

app/components/avo/views/resource_index_component.rb
```diff
@@ -29,6 +30,28 @@ def title
     end
   end
 
+  def view_type
+    @index_params[:view_type]
+  end
```
Paul-Bob, line 35: This method was refactored and is not necessary, `view_type` can now be accessed from the resource: `resource.view_type`

```suggestion
```

app/components/avo/views/resource_index_component.rb
```diff
@@ -29,6 +30,28 @@ def title
...
+        via_record_id: @parent_record.to_param
+      }
+    end
+
+    helpers.edit_bulk_update_path(resource: @resource, **args)
+  end
+
+  def available_view_types
+    @index_params[:available_view_types]
+  end
```
Paul-Bob, line 53:
```suggestion
```

This method was refactored and can be accessed from the resource: `resource.avialable_view_types`

lib/avo/resources/base.rb
```diff
@@ -641,6 +641,7 @@ def entity_loader(entity)
       end
 
       def record_param
+        return nil if @record.nil?
```
Paul-Bob, line 644: When is the `record_param` called and the `@record` is nil?

Paul-Bob (changes requested):
Hi @Nevelito,

Thanks for the great work! 🙌
I did some manual testing, and the controller already seem to be working as expected.
I'll perform a more detailed review of the controllers later.

In the meantime, I have a few questions about parts of the current implementation. Let's continue the discussion within the individual review comments.

Thanks again for your patience and collaboration!

lib/avo/resources/base.rb
Nevelito: for now there is no situation like this, i forgot to delete it

app/components/avo/views/resource_edit_component.html.erb
Nevelito: we use it to compare data, the case is that progress bar can not be null, it will automaticlly appear at 50%, so to avoid problems we have to double check if it was moved or delete it from the bulk update

Nevelito: btw during todays tasting I saw that tere is problems with stage in Projects model but it seems easy to solve, but I will solve it after weekend

app/components/avo/views/resource_edit_component.html.erb
Nevelito: The model function had some conditions in early versions and forgot to delete it

app/components/avo/views/resource_edit_component.html.erb
Paul-Bob: Sure, please ask another review when all the review conversations are resolved

Nevelito: done

Paul-Bob (changes requested):
Hi @Nevelito I left some more questions and suggestions

app/components/avo/views/resource_edit_component.html.erb
Paul-Bob: @Nevelito regarding this use case. I completely removed the related code but couldn't identify the specific problem it is solving. Could you clarify a concrete scenario where this is truly needed? For example:

1. Navigate to Projects and select two projects, one with progress bar at X, the other at Y  
2. Initiate a bulk edit and save without making any changes  
3. Observe the issue that occurs  
4. Reintroduce the prefilled values, and the issue is resolved

app/components/avo/views/resource_index_component.rb
Paul-Bob: I pushed a commit that removes this method https://github.com/avo-hq/avo/pull/3695/commits/b0ea7501520a42e1deb0ed3486619f676773b01a

app/components/avo/fields/select_field/edit_component.html.erb
```diff
@@ -1,6 +1,6 @@
 <%= field_wrapper(**field_wrapper_args) do %>
   <%= @form.select @field.id, options_for_select(@field.options_for_select, selected: @field.value), {
-      include_blank: @field.include_blank
+      include_blank: (params[:controller] == "avo/bulk_update") || @field.include_blank
```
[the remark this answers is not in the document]
Nevelito: couse if include blank will be false the bulk update will always change it so it have to be true in bulk update to not update things that we do not want

app/components/avo/views/resource_index_component.rb
```diff
@@ -29,6 +30,20 @@ def title
...
+
+    if @parent_record.present?
+      args = {
+        via_resource_class: parent_resource.class.to_s,
+        via_record_id: @parent_record.to_param
+      }
+    end
+
+    helpers.edit_bulk_update_path(resource: @resource, **args)
+  end
```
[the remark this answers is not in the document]
Nevelito: you right, fixed

app/components/avo/views/resource_edit_component.rb
```diff
@@ -18,6 +20,8 @@ def title
   end
 
   def back_path
+    return helpers.resources_path(resource: @resource) if from_bulk_update?
```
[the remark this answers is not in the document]
Nevelito: done

app/components/avo/views/resource_index_component.rb
```diff
@@ -29,6 +30,20 @@ def title
     end
   end
 
+  def bulk_edit_path
+    # Add the `view` param to let Avo know where to redirect back when the user clicks the `Cancel` button.
+    args = {via_view: "index"}
+
+    if @parent_record.present?
+      args = {
+        via_resource_class: parent_resource.class.to_s,
```
Paul-Bob, line 39:
```suggestion
        via_resource_class: @parent_resource.class.to_s,
```

Nevelito:
Hey @Paul-Bob, is there anything else to do in this PR for now?

Nevelito:
Progress fields require special handling because during editing, when the progress bar value is nil, it defaults to 50%. This causes issues during bulk updates — if values are not prefilled, the progress bar will always be set to 50%, even if no change was intended. You can verify this behavior yourself. The prefilled fields logic is used specifically to avoid unintentionally setting all progress bars to the same value during bulk updates, which is why it's not changed.

app/components/avo/views/resource_edit_component.html.erb
Nevelito: I see, this code is not necessary here.

Paul-Bob:
I think I’ve got it now. I believe we should look into the progress field and see why it isn’t allowing the submission of a `nil` value. I did some tests on the project creation form (not the bulk update), and when visiting the create form, `@record.progress` is `nil`, but upon submission, it sends `50`. Let's do some research around this. I feel that the correct approach would be to fix the progress field to allow it to submit `nil` values.

Let’s please work on a separate PR branched from `main` for this, and put the current PR on hold until we fix the progress field. I’ll create an issue to track it.

Paul-Bob (changes requested):
After we fix the progress bar field issue, we should be able to remove the specific complex logic around progress fields in this PR.

Nevelito:
@Paul-Bob, I Love this approach! Jumping into the progress fields issue now. Thanks a lot!

Paul-Bob (changes requested):
This PR is taking longer than usual because it's a more substantial feature, and we need to be careful with the complexity it introduces. Typically, we build things iteratively, meaning the first iteration should be clean and simple enough to maintain and extend later. Thanks so much for your patience with all the back-and-forth on this one, @Nevelito .

I went ahead and [pushed some tweaks](https://github.com/avo-hq/avo/pull/3695/commits/657c5ca6e98d5c740e053aec31be6b3a250fe856) around areas I consider important for the first version. You can see it in the commit history, but I'm summarizing the key changes here:

- I noticed `@fields` was being used to store a hash of attributes and values. In Avo, we refer to variables as fields only when they are actual Avo fields.
- The method `prefill_fields` was receiving two arguments, but it was only ever called once, using instance variables. I updated it to take zero arguments and use the instance variables directly.
- The logic to construct the final hash with prefilled and `nil` values was split across `edit` and `prefill_fields`. I consolidated that logic into `prefill_fields`, which now returns a unified hash. Common values are preserved, mismatches become `nil`. For example, for two projects that only share the same progress value (7), the result would be:

  ```ruby
  {
    "id"=>nil, "name"=>nil, "status"=>nil, "stage"=>nil, "budget"=>nil,
    "country"=>nil, "users_required"=>nil, "started_at"=>nil, "meta"=>nil,
    "created_at"=>nil, "updated_at"=>nil, "description"=>nil, "progress"=>7
  }
  ```

- Since the contents weren't actually Avo fields, I renamed the method to `bulk_values`, and added a comment explaining what it does for clarity.
- I found that `set_fields` was unnecessary. It just grabbed the first record and its attributes. I replaced it with a direct call to the resource model class. I'm aware there was a conditional guard there, and I'll come back to that in this review with an improvement that makes the guard redundant.
- I switched the direct assignment of the record on the resource to use the hydration pattern. It's not essential, but aligns better with our not documented guidelines.
- The `edit` response now uses the computed component approach like the base controller does. This gives developers flexibility to set custom bulk edit components, while defaulting to the resource's edit component. This needs to be documented in `self.components`.

Since the reasoning behind these changes might not be fully clear from the commits alone, I felt it was worth writing it out here.

---

During this review process, I also noticed a few UI/UX aspects that need improvement:

- [ ] The bulk edit button should be part of the index controls, alongside the create and actions buttons, rather than being manually inserted into the filters area.
- [ ] It should be disabled until at least one record is selected (similar to how non-standalone actions work), and re-disabled when all records are deselected. This also ties into the earlier point about the conditional guard, we're now preventing code execution without a valid query, so the guard becomes unnecessary. In fact, if a blank query does slip through, we want it to raise an error because that should never happen.

On the topic of the query itself: I noticed we're passing it directly through params, which opens the door for malicious actors to tamper with the payload and possibly target records beyond their authorization scope. This could be mitigated using the built-in encryption service, but that's a larger topic. For now, let's focus on the two points I just mentioned.

Thanks again for your patience throughout this PR. It's a big one, and each review cycle takes time, but it's worth it to ensure we get it right.

Nevelito:
@Paul-Bob have a look at new changes, I think it starting to look as it should. There is some issue with system spec in old rails version but I will try to figure it out tomorrow.

Paul-Bob (commented):
The `bulk_update_no_records` and `bulk_update` translations don't seem to be used anymore.

Other than that, I think we're really close to completing this PR.

Nevelito:
Hi @Paul-Bob I'm going on vacation for two months starting in two weeks. It would be great to wrap up this feature before then. Could you please take another look and let me know if there's anything I can do on my side to move it forward?

Thanks in advance!

Paul-Bob:
> Hi @Paul-Bob I'm going on vacation for two months starting in two weeks.

Sounds great! Enjoy your time off!

> It would be great to wrap up this feature before then. Could you please take another look and let me know if there's anything I can do on my side to move it forward?
> Thanks in advance!

All good on your end, thanks so much for the effort you put into this! I'll take it from here and wrap up the PR. There are still a few loose ends and the last time I spoke with Adrian, we discussed possibly moving this PR to a private repo.

Really appreciate all the work you put in, this one was a bit of a beast!

Let me know if there's another issue you'd like to pick up next.

Paul-Bob:
Closing this PR to keep the queue clean since we are not focusing on it right now.

We will probably reopen and rebase it on Avo 4.

Thank you again Newel for all the great work on this one!

PR #3705, opened 2025-03-01
chore: add app_id if provided

adrianthedev:
# Description

When the install command has the app_id, it will enable the `license_key` config.

# Checklist:


lib/generators/avo/install_generator.rb
```diff
@@ -24,6 +25,10 @@ def create_resources
           end
         end
 
+        def app_id
+          options[:"app-id"] || SecureRandom.hex(6).first(12)
```
adrianthedev, line 29:
```suggestion
          options[:"app-id"]
```

Issue #3604, opened 2025-01-22
No cancel button showing when creating new user via new review


ianhearneIE:
### Describe the bug

There doesnt appear to be any cancel button when trying to create a new user via the Create new review process. I managed to reproduce this within the Avo demo instance.

### Steps to use in the reproduction repository
1. Go to Reviews > Create new review
2. Under User click Create new user
3. Only the Save button will appear with no cancel option showing

### Expected behavior & Actual behavior

There should be the option to cancel creating a new user when only the following is showing
![Image](https://github.com/user-attachments/assets/407e98ee-cb56-4fea-adb9-8268221e8dbe)

Here is the option when directly creating a new user by navigating to Users > Create new user
![Image](https://github.com/user-attachments/assets/84958809-e082-4c5b-9704-95951a5b5c74)

### System configuration
**Avo version**: 3.15.4 (my app) 3.16.1 (avo demo)

**Rails version**: 7.2

**Ruby version**: 3.3

**License type**:

 - [ ] Community
 - [ ] Pro
 - [x] Advanced

**Are you using Avo monkey patches, overriding views or view components?**

 - [ ] Yes. If so, please post code samples.
 - [x] No

### Screenshots or screen recordings

https://github.com/user-attachments/assets/831eeb72-848c-40aa-91cb-61a4dc02162e

#### Impact

 - [ ] High impact (It makes my app un-usable.)
 - [ ] Medium impact (I'm annoyed, but I'll live.)
 - [x] Low impact (It's really a tiny thing that I could live with.)

#### Urgency

 - [ ] High urgency (I can't continue development without it.)
 - [ ] Medium urgency (I found a workaround, but I'd love to have it fixed.)
 - [x] Low urgency (It can wait. I just wanted you to know about it.)

piyush828-design:
@Paul-Bob I would like to work on this.

Paul-Bob:
Thank you @piyush828-design !

Let me know if you need any help

Paul-Bob:
https://github.com/avo-hq/avo/pull/3756

PR #3756, opened 2025-03-26
Fix: No cancel button showing when creating new user via new review

piyush828-design:
# Description

No cancel button showing when creating new user via new review 
issue link - https://github.com/avo-hq/avo/issues/3604

# Checklist:

## Screenshots & recording
[Screencast from 26-03-25 03:50:30 PM IST.webm](https://github.com/user-attachments/assets/91a63e99-f796-4aa4-a377-fc6429fda522)

## Manual review steps

I've run all the tests to ensure that none of the existing tests breaking due to my changes.


piyush828-design:
@Paul-Bob please help reviewing PR

Paul-Bob (changes requested):
Hi @piyush828-design 👋

Thanks for submitting this PR!

I noticed that the cancel button in your example links to `"admin/resources/users"` and triggers a GET request when clicked.

Let's make sure the button doesn't render as a link, and instead includes:

```erb
data: { action: 'click->modal#close' }
```

This will ensure the modal closes without making any network requests.

Thanks again!

piyush828-design:
Hi @Paul-Bob ,
Thanks for your inputs. Could you please help me to understand where should I add `data: { action: 'click->modal#close' }`?
as I added it in render_back_button method but it is still triggering the endpoint.
Sorry if I've done anything wrong as I didn't work on hotwire that much.
Thanks!

Paul-Bob:
> Could you please help me to understand where should I add data: { action: 'click->modal#close' }?

Yes, that `data` attribute should be added to the button that's rendered as the "Cancel" button, but only when it's inside the **new form modal**. In that context, it should not trigger any redirect or navigate to another page.

Here are a couple of places where similar logic is already in use, for reference:
- https://github.com/avo-hq/avo/blob/main/app/views/avo/associations/new.html.erb#L69
- https://github.com/avo-hq/avo/blob/main/app/views/avo/actions/show.html.erb#L50

> as I added it in render_back_button method but it is still triggering the endpoint.
Sorry if I've done anything wrong as I didn't work on hotwire that much.
Thanks!

No worries at all, appreciate you looking into it! You're right, it should be added in `render_back_button`, but only for this specific case when it's being rendered within the modal form. It shouldn't apply globally in all contexts.

Paul-Bob (approved):
I've pushed a small tweak to preserve the original lines and avoid over-refactoring the link logic.

Thank you for this contribution @piyush828-design!

piyush828-design:
Thank you @Paul-Bob ;)

PR #3758, opened 2025-03-26
Allow style customisation of the profile photo

rctneil:
# Description
A couple of options on the profile photo option to allow users to improve the look of the selected image.

I can use these to make this:
Screenshot 2025-03-26 at 22 01 38

look like this:
Screenshot 2025-03-26 at 22 01 26

Fixes # (issue)

# Checklist:

## Screenshots & recording
Screenshots above

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/app/components/avo/profile_photo_component.html.erb
+++ b/app/components/avo/profile_photo_component.html.erb
@@ -1,6 +1,9 @@
-<%= image_tag helpers.main_app.url_for(@profile_photo.value), class: class_names(
-  "relative rounded-full object-cover aspect-square border-4 border-application",
+<%= image_tag helpers.main_app.url_for(@profile_photo.value),
+  style: "background-color: #{ @profile_photo.options[:background_color] || '#ffffff' }",
+  class: class_names(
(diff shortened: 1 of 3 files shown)


adrianthedev (changes requested):
This looks like we're trying to bring CSS into the Ruby DSL.
Can this be fixed with some custom CSS? 

If so, I think I prefer that method.

rctneil:
@adrianthedev Oh, ok then.

rctneil:
> I think this should work on your end:
> 
> ```css
> .has-profile-photo [data-target=panel-header] > div img {
>   ... CSS rules here
> }
> ```

Yes, this is what i was going to look into this evening.

I'm sorry for submitting the PR.  I understand the reasoning.    Best get started on my next one as I have a small thing that I hope I can improve!

adrianthedev:
No need to apologize. 
Thanks for putting in the time. I'm sorry we can't merge it, but we learned something.

Looking forward to the next one.

Issue #3718, opened 2025-03-05
`reload_records` not working with the Grid view


m7madmagdy:
### Context

In the grid view type, if I use `reload_records` for any action, it's not working correctly and I must refresh the page manually 

### System configuration
**Avo version**: 3.17.9

**Rails version**: 8.0

**Ruby version**: 3.4.2

**License type**

 - [ ] Community
 - [ ] Pro
 - [x] Advanced

**Are you using Avo monkey patches, overriding views or view components?**

 - [x] Yes. If so, please post code samples.
 - [ ] No

Deepanshu0211:
# Avo `reload_records` Not Working in Grid View

##  **Issue Description**
In the **grid view**, calling `reload_records` does not refresh the records correctly. A manual page refresh is required.

##  **System Configuration**
- **Avo version:** 3.17.9  
- **Rails version:** 8.0  
- **Ruby version:** 3.4.2  
- **License type:** Community / Pro / Advanced  
- **Using monkey patches or custom views?** No  

---

## **Debugging Steps**
### ✅ 1. Check Browser Console  
- Open DevTools (`F12` → Console)  
- Look for **JavaScript errors** or **network errors** when `reload_records` is triggered  

### ✅ 2. Verify Grid View Setup  
- Are you calling `reload_records` inside a **custom action** or **controller**?  
- Check if the method is being executed properly.  

### ✅ 3. Check Avo Documentation & Source Code  
- Avo might have **changed** how `reload_records` works.  
- Check [Avo GitHub Issues](https://github.com/avo-hq/avo/issues) and [Avo Docs](https://docs.avohq.io) for updates.  

### ✅ 4. Force Turbo to Work (If Using Hotwire)  
Try manually forcing Turbo to refresh:  
```js
Turbo.visit(window.location.href, { action: "replace" })

```
or, soft trigger reload 
  ```Avo.reload()```

Paul-Bob:
Hi @m7madmagdy,  

The `reload_records` action response is not expected to work in grid view. As stated in the [docs](https://docs.avohq.io/3.0/actions.html#reload_records), _this option leverages Turbo Stream to refresh specific **table rows** in response to an action._ 

Thanks for reaching out! We'll discuss the possibility of extending this feature to work with other view types, such as grid

Paul-Bob:
We'd love to merge a PR if you're up for contributing! I'd be happy to guide you through the process.

m7madmagdy:
[...]

PR #3780, opened 2025-04-04
enhancement: extend `reload_records` to support grid view

m7madmagdy:
# Description

Fixes https://github.com/avo-hq/avo/issues/3718

Support `reload_records` with grid and map views after running an action.

# Checklist:

- [x] I have made corresponding changes to the [documentation](https://github.com/avo-hq/docs.avohq.io/pull/382)

## Screenshots & recording

[Screencast from 04-06-2025 09:20:49 PM.webm](https://github.com/user-attachments/assets/3b363198-3e93-4ca8-a138-95e822761442)

## Manual review steps

1. Step 1
1. Step 2


lib/avo/base_action.rb
```diff
@@ -8,6 +8,16 @@ class BaseAction
...
 
+    VIEW_ITEM_NAME_BY_TYPE = {
+      table: "avo/index/table_row_component",
+      grid: "avo/index/grid_item_component"
+    }.freeze
+
+    COMPONENT_ROW_TYPES = {
+      table: "Avo::Index::TableRowComponent",
+      grid: "Avo::Index::GridItemComponent"
+    }.freeze
```
adrianthedev, line 19: I skimmed the PR, but I wasn wondering if we can have just one constant and get the other from it?

Like turn `Avo::Index::TableRowComponent` into `avo/index/table_row_component`?

m7madmagdy: Good catch! Yeah, we can generate the path from the class name with `.underscore`. I'll update it to use a single constant.

lib/avo/base_action.rb
```diff
@@ -293,40 +299,39 @@ def reload
       self
     end
 
-    def reload_record(records)
+    def reload_record(records, view_type)
```
Paul-Bob, line 302: This is a public API, developers shouldn't need to specify the `view_type` manually, Avo should infer it implicitly behind the scenes.

The `reload_records(query)` method should refresh the table rows when in table view, and the grid items when in grid view.

Paul-Bob (changes requested):
Thanks for looking into this @m7madmagdy

The view type should be inferred by Avo.

lib/avo/base_action.rb
m7madmagdy: Okay @Paul-Bob 
How can I get the `view_type` implicitly?

Paul-Bob (commented):
Hi @m7madmagdy, I noticed that you requested a review and then marked the PR as a draft.

Feel free to ping me once it's ready for a full review.

Paul-Bob (commented):
Hi @m7madmagdy, thanks for looking into this!

We're getting close to merging it. I've made [this refactor](https://github.com/avo-hq/avo/pull/3814) that moves `view_type` from `@index_params` into the `resource` object. This way, we don't need to pass `view_type` through all the components like before. Also, for people using `link_arguments`, there is no need to manually include `view_type` as an argument anymore.

I also separated the reload row and reload grid functions. The grid item reload is much simpler and doesn't need to compute header fields or run other "heavy" logic.

Let me know what you think about these changes! Also, could you please add a test for this feature and a short paragraph in the docs mentioning that this feature is available for grid items starting from Avo 3.next_minor_version?

Paul-Bob:
@m7madmagdy , please re-request a review once the tests and docs are up for a final review.

image

m7madmagdy:
@Paul-Bob, I requested a review
I updated the avo docs, but sorry, I don't have time to write tests

Paul-Bob (approved):
LGTM

Thanks @m7madmagdy and @a7madM for your contribution on this PR

PR #3801, opened 2025-04-17
fix: missing icon button

icaroryan:
# Description
Icon buttons (`style: :icon`) don't show up on the UI because it doesn't have a size property.

I ran into this problem when trying to customize the `actions_list` in the row control -> `actions_list style: :icon`.
The icon was still present in the HTML, but it only showed up after adding `h-6`as a class.
Something else I noticed is that `stroke-width` also didn't match with the other icons, but I couldn't find out a way to change that in the code

# Checklist:

### Before
image

### After
image

## Manual review steps

Since I don't have access to avo_advanced, @Paul-Bob helped me reproduce this by adding a custom button in `admin/custom_tool`
```
<%= a_link 'javascript:void(0);',
           icon: 'plus',
           size: :sm,
           # style: :primary,
           style: :icon,
           color: :stale do %>
  Test
<% end %>
```

1. When using `style: :icon`, the button should be visible on the UI
2. The icons should have a similar style from the others on the page -> can use `row_controls` to test this

### DIFF
--- a/app/components/avo/button_component.rb
+++ b/app/components/avo/button_component.rb
@@ -168,7 +168,7 @@ class Avo::ButtonComponent < Avo::BaseComponent
   end
 
   def icon_size_classes
-    icon_classes = ""
+    icon_classes = " h-6"
     return icon_classes if is_icon?
 
     case @size


Paul-Bob:
Hi @icaroryan thanks for submitting this PR

I had a look and it seems that `actions_list style: :icon` was not working as expected because the component was not getting the `icon_class` correclty.

Just pushed a commit that fixes it.

PR #3804, opened 2025-04-21
fix: apply `scope` on `has_many` fields while searching on the association table

### LINKED ISSUE #1859: Searchable scope do not apply the field scope
### Describe the bug

Searchable do not respect the field scope.

### Steps to Reproduce
1. Inside team resource add this field configuration:
```ruby
  field :memberships,
    as: :has_many,
    searchable: true,
    attach_scope: -> do
      query.where.not(user_id: parent.id).or(query.where(user_id: nil))
    end,
    scope: -> { query.where(id: [45, 44, 43]) }
```
2. Go to team show page and see only 3 records on table
3. Click to search on table and notice that more than 3 records appear on search

The field scope should be applied on the search scope

### System configuration
**Avo version**:
2.37.0

**Rails version**:

**Ruby version**:

**License type**:

 - [ ] Community
 - [x] Pro

 - [x] No

image

#### Impact

#### Urgency

piyush828-design:
# Description

Fixes #1859 

# Checklist:

## Screenshots & recording
image


Paul-Bob:
> I'm not getting which scope needs to be checked?

For example this field:

```ruby
  field :memberships,
    as: :has_many,
    searchable: true,
    scope: -> { query.where(id: [45, 44, 43]) }
```

When search is pressed on it `.where(id: [45, 44, 43])` should be applied on the search query.

---

For this field: 

```ruby
  field :admins,
    as: :has_many,
    searchable: true,
    scope: -> { query.admins }
```

When search is pressed on it `.admins` should be applied on the search query.

piyush828-design:
When we click on `search`, below condition never get true. Due to that apply_scope doesn't get executed.
image

I'm assuming these methods are responsible for to apply scope.  
image

Do I need to dig more deeper? Right now, I'm not able to think of a solution. Maybe I need more understanding of the codebase.

Paul-Bob:
You were definitely on the right track @piyush828-design!

> Do I need to dig deeper?

Yes, that's correct, digging deeper usually helps in these cases.

> Right now, I'm not able to think of a solution. Maybe I need more understanding of the codebase.

Totally understandable. I actually took a closer look at this PR to be able to guide you better, and found that the solution is to apply the field scope inside `apply_has_many_scope`.

I've pushed the necessary changes to the branch. Thanks a lot for all your effort on this PR, you were on the right track by pinpointing the area in the code where the issue was.

piyush828-design:
Thanks for the appreciation @Paul-Bob . I'll spend more time on the codebase.

piyush828-design:
but still it is showing more records.
image

Paul-Bob:
> but still it is showing more records. image

The image you're referencing appears to be from the index view and seems unrelated to the issue being addressed in this PR. However, good catch, it's happening because the default filters aren't being applied to the search query.

The `Team` resource uses the `Avo::Filters::MembersFilter`, which defaults to `has_members: true`, but this filter doesn't seem to be applied during searches. If you're interested, feel free to explore this further in a separate PR. It would involve digging into the code to ensure filters from the index query also apply to search.

About this PR, while looking into it to guide you, I ended up finding a solution for the current issue. I'm just making final tweaks for the tests to pass, and then it should be good to merge.

Thanks again for your contribution!

Issue #1902, opened 2023-08-12
[RFC] `format_(edit|new|show|index|forms|display)_using` field options


adrianthedev:
### Context

We usually need to format values in different ways and most of the time on `Index` and `Show` views.
One example of that is currency. On the edit view we'll want to display the raw value but on the `display` views (Index, Show) you need the embellished view.

Example:

```ruby
# Before
field :amount, as: number, format_using: -> do
  if view == :index || view == :show
    number_to_currency value
  else
    value
  end
end

# After
field :amount, as: number, format_index_using: -> { number_to_currency value }

# or using display (index and show views) after we merge it
field :amount, as: number, format_display_using: -> { number_to_currency value }
```

Another example are computed links. For example have an id in the database represented as a link on `display` views.

```ruby
# Before
field :stripe_id, as: :text, format_using: -> do
  if view == :index || view == :show
    link_to id, "https://dashboard.stripe.com/plans/#{value}"
  else
    value
  end
end

# After
field :stripe_id, as: :text, format_display_using: -> { link_to id, "https://dashboard.stripe.com/plans/#{value}" }
```

The `format_using` option will still be available and it will take precedence over all others.

```ruby
# this will yield "general" on all views
field :stripe_id, as: :text, format_using: -> { "general" }, format_index_using: -> { "index" }
```

**Avo version**: 3 and up

This will deprecate the `decorate` option which supported only display views.
We are doing this change so we can support a more granular per-view formatting.

zhephyn:
@Paul-Bob Let me give this a shot

zhephyn:
@Paul-Bob I think this should be closed as its PR was merged.

Paul-Bob:
> @Paul-Bob I think this should be closed as its PR was merged. 

Right, missed that, thanks!

PR #3821, opened 2025-04-30
Adds a format_display_using option for the number field

zhephyn:
This PR adds a format_display_using option for the number field such that instead of this:
```ruby
field :amount, as: number, format_using: -> do
  if view == :index || view == :show
    number_to_currency value
  else
    value
  end
end
```
We can do this instead which is more cleaner and allows us to apply appropriate formatting on a value for both the index and show views at once :
```ruby
field :amount, as: number, format_display_using: -> { number_to_currency value }
```
Solves #1902

# Checklist:

### DIFF
--- a/lib/avo/fields/number_field.rb
+++ b/lib/avo/fields/number_field.rb
@@ -1,9 +1,11 @@
 module Avo
   module Fields
     class NumberField < TextField
+      include ActionView::Helpers::NumberHelper
       attr_reader :min
       attr_reader :max
       attr_reader :step
+      attr_reader :format_display_using
 
       def initialize(id, **args, &block)
         super(id, **args, &block)
@@ -11,7 +13,21 @@ module Avo
         @min = args[:min].present? ? args[:min].to_f : nil
         @max = args[:max].present? ? args[:max].to_f : nil
         @step = args[:step].present? ? args[:step].to_f : nil
+        @format_display_using = args[:format_display_using]
       end
+
+      def value(property = nil)
+        raw_value = super
+        if @format_display_using && view == :show || view == :index
+          return raw_value if @formatting
+          @formatting = true
+            formatted_value = instance_exec(&@format_display_using)
+          @formatting = false
+          return formatted_value
+        elsif view == :edit
+          raw_value
+        end
+      end 
     end
   end
 end


zhephyn:
@Paul-Bob . I have a working solution so far. Before i proceed with writing tests for this particular feature and solving the failed CI checks, i wanted check if this is going in the expected direction. If yes, I'll proceed to write the tests and then update the docs.

Also, i opted for implementing just the format_display_using option alone since it allows us to do formatting for both the index and show views at once, instead of implementing individual options named format_index_using and format_show_using for the index and show page respectively. 
In case these had to be implemented as well, kindly let me know and I'll make the necessary changes.

Paul-Bob (changes requested):
> @Paul-Bob . I have a working solution so far. Before i proceed with writing tests for this particular feature and solving the failed CI checks, i wanted check if this is going in the expected direction. If yes, I'll proceed to write the tests and then update the docs.
> 
> Also, i opted for implementing just the format_display_using option alone since it allows us to do formatting for both the index and show views at once, instead of implementing individual options named format_index_using and format_show_using for the index and show page respectively. In case these had to be implemented as well, kindly let me know and I'll make the necessary changes.

Thanks for the work on this

A few clarifications before moving forward: this change should apply to all field types, not just number fields. Starting with `format_display_using` is a good starting point, but the goal outlined in the issue is to build a DSL that supports formatting for specific views like `format_index_using`, `format_show_using`, and so on, in addition to `format_display_using` for shared display logic.

Let's start by making these changes first, then please re-request a review before starting on tests and docs.

zhephyn:
@Paul-Bob. New changes are up. Kindly take a look when you get some time.

lib/avo/fields/base_field.rb
```diff
@@ -36,6 +36,9 @@ class BaseField
       attr_reader :nullable
       attr_reader :null_values
       attr_reader :format_using
+      attr_reader :format_display_using
+      attr_reader :format_index_using
+      attr_reader :format_show_using
```
Paul-Bob, line 41: Let's also include `edit` `new` and `forms`

```suggestion
      attr_reader :format_display_using
      attr_reader :format_index_using
      attr_reader :format_show_using
      attr_reader :format_edit_using
      attr_reader :format_new_using
      attr_reader :format_form_using
```

lib/avo/fields/base_field.rb
```diff
@@ -175,6 +181,18 @@ def value(property = nil)
...
+        end
+
+        if @format_index_using.present? && @view.index?
+          final_value = execute_context(@format_index_using, value: final_value)
+        end
+
+        if @format_show_using.present? && @view.show?
+          final_value = execute_context(@format_show_using, value: final_value)
+        end
+
```
Paul-Bob, line 195: This logic needs a bit of refinement.

For instance, when the view is `index`, the code should follow this order of precedence:
1. Check if `@format_index_using` is defined — if so, use it.
2. If not, fall back to `@format_display_using`, which offers less granularity.
3. If neither is present, default to `@format_using`.

I've implemented a similar pattern for fetching fields from a resource, like `show_fields`, `display_fields`, etc.  
You can take a look here:  
https://github.com/avo-hq/avo/blob/enhancement/assign_action_record_from_index_page/lib/avo/resources/base.rb#L322

The idea is that a granular option is applied it takes precedence over the less granular. That should be applied for all the views

Paul-Bob (changes requested):
Thank you for your patience on this PR, @zhephyn.

We're definitely on the right track now and heading in the right direction.  
Let's continue the discussion within the individual review comments.

lib/avo/fields/base_field.rb
```diff
@@ -71,6 +74,9 @@ def initialize(id, **args, &block)
         @nullable = args[:nullable] || false
         @null_values = args[:null_values] || [nil, ""]
         @format_using = args[:format_using]
+        @format_display_using = args[:format_display_using]
+        @format_index_using = args[:format_index_using]
+        @format_show_using = args[:format_show_using]
```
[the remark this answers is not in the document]
zhephyn: Some clarification on this, should the comment state that "@decorate was deprecated in favor of format_display_using which should be used instead"?

zhephyn:
@Paul-Bob could you help me out with the failing tests? Tried to narrow down what's causing them with no success

Paul-Bob:
> @Paul-Bob could you help me out with the failing tests? Tried to narrow down what's causing them with no success

Hi @zhephyn, of course!

I've recorded a Loom where I walk through the issue in one of the failing feature tests.

This is my usual approach for broken CI tests: I imagine a PR that breaks, say, 40 tests. Fixing just one often resolves several others due to shared causes. So, I fix one, re-run the full test suite, then address the next, and so on.

Thanks for all the effort you're putting into this!

https://www.loom.com/share/62de28586c4f457296e06f7dbd1f6019?sid=340c5f42-6e1b-4ccd-bf33-ccddad97bff3

zhephyn:
@Paul-Bob thanks so much for the debugging tip. I figured out what the issue was. The value was being formatted twice with the format_using formatter. On solving that, all the failing tests passed.

lib/avo/fields/base_field.rb
```diff
@@ -172,10 +184,8 @@ def value(property = nil)
...
 
-        # Run the value through resolver if present
-        if @format_using.present?
-          final_value = execute_context(@format_using, value: final_value)
-        end
+        # Format value based on available formatter
+        final_value = format_value(final_value)
 
         if @decorate.present? && @view.display?
           final_value = execute_context(@decorate, value: final_value)
```
Paul-Bob, line 196: We can safely remove the `@decorate` application since we're deprecating it in favor of `@format_display_using` and that is already being applied on the `format_value` method

lib/avo/fields/base_field.rb
```diff
@@ -71,6 +77,12 @@ def initialize(id, **args, &block)
         @nullable = args[:nullable] || false
         @null_values = args[:null_values] || [nil, ""]
         @format_using = args[:format_using]
+        @format_display_using = args[:format_display_using] || args[:decorate]
```
Paul-Bob, line 79:
```suggestion
        @format_display_using = args[:format_display_using] || args[:decorate]
        
        unless Rails.env.production?
          if args[:decorate].present?
            puts "[Avo DEPRECATION WARNING]: The `decorate` field configuration option is no longer supported and will be removed in future versions. Please discontinue its use and solely utilize the `format_display_using` instead."
          end
        end
        
```

lib/avo/fields/base_field.rb
```diff
@@ -317,6 +327,29 @@ def get_resource_by_model_class(model_class)
...
         resource || (raise Avo::MissingResourceError.new(model_class, self))
       end
+
+      def format_value(value)
+        final_value = value
+
+        formatters_by_view = {
+          index: [:format_index_using, :format_display_using, :format_using],
+          show: [:format_show_using, :format_display_using, :format_using],
+          edit: [:format_edit_using, :format_using],
```
Paul-Bob:
```suggestion
          edit: [:format_edit_using, :format_form_using, :format_using],
```

zhephyn:
@Paul-Bob . I've updated the PR to reflect the changes you suggested earlier. And now that the CI tests are passing, i think we can proceed to the tests and the documentation. 

I already have a draft of the documentation which I'm gonna push shortly to the docs repository after which it'll be available for you to look at and suggest any changes necessary.

zhephyn:
I tried to merge main into this branch however, one test is now failing yet before i tried to merge main, all of them were passing. Don't know if this is part of the scope of the PR or not

Paul-Bob:
Thanks, @zhephyn. Let's focus on the docs and add some tests for this.

> I tried to merge main into this branch however, one test is now failing yet before i tried to merge main, all of them were passing. Don't know if this is part of the scope of the PR or not

I re-ran the failing tests, probably just some flakiness.
If the tests continue to fail on this PR but are passing on main, then it's within the scope of the PR

zhephyn:
@Paul-Bob  I've pushed tests for the changes of this PR. 

Also, one thing to note is that, for the ```format_show_using```, ```format_index_using``` and ```format_display_using```, we expect these to format a value and prettify it while for the ```format_form_using``` and ```format_edit_using```, these should display the actual raw value which was saved in the database.
 
This is the kind of thinking i used when writing the tests as its what had been mentioned as per the details displayed on the issue page. In case this out to be changed, kindly lmk and ill adjust.

Paul-Bob:
I understand what you mean, formatting on forms should be used carefully, but there are indeed cases where a developer might format the value on the forms in addition to the value stored in the database.

Any `format_#{view}_using` should display the formatted value regardless of the view.

Typically on forms, `format_edit_using` is used alongside `update_using`, which formats the value before it reaches the database during an update.

In the tests, let's ensure that the formatter is indeed formatting the value across all views.

zhephyn:
@Paul-Bob New changes for this are up as well. Kindly take a look when you get some time.

Paul-Bob (approved):
LGTM, thanks @zhephyn!!!

zhephyn:
@Paul-Bob I was waiting for your review of the tests and then start on the docs. I've noticed that you did the docs though. It's cool either way. Glad the PR has been merged. Thanks for your guidance throughout. Learnt a ton of stuff🙏

Paul-Bob:
I was reviewing which PRs to merge for tomorrow's release, so I went ahead and handled the docs since we need them right away. Thanks for all your contributions, I'm really glad you're learning through them!

PR #3856, opened 2025-05-15
refactor: allow progress field to submit nil values

### LINKED ISSUE #3846: refactor: allow progress field to submit `nil` values
I think I’ve got it now. I believe we should look into the progress field and see why it isn’t allowing the submission of a `nil` value. I did some tests on the project creation form (not the bulk update), and when visiting the create form, `@record.progress` is `nil`, but upon submission, it sends `50`. Let's do some research around this. I feel that the correct approach would be to fix the progress field to allow it to submit `nil` values.

Let’s please work on a separate PR branched from `main` for this, and put the current PR on hold until we fix the progress field. I’ll create an issue to track it.

_Originally posted by @Paul-Bob in https://github.com/avo-hq/avo/issues/3695#issuecomment-2872865854_

Nevelito:
# Description
Fix progress bar field to allow a nil value during creation and update if the bar was untouched and the default field value is nil.

Fixes #3846

# Checklist:

## Screenshots & recording

[Screencast from 15.05.2025 17:22:41.webm](https://github.com/user-attachments/assets/dae790a6-c806-4480-8b4e-5f32ad77c9ca)

### DIFF
--- a/app/components/avo/fields/progress_bar_field/edit_component.html.erb
+++ b/app/components/avo/fields/progress_bar_field/edit_component.html.erb
@@ -5,14 +5,22 @@
         <span data-progress-bar-field-target="label"><%= @field.value %></span><%= @field.value_suffix if @field.value_suffix.present? %>
       </div>
     <% end %>
-    <%= @form.range_field @field.id,
+    <%= @form.text_field @field.id,
+      value: @field.value,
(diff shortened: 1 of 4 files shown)


Nevelito:
@Paul-Bob I finished, all tested and works as we want

Paul-Bob (approved):
LGTM!

Thanks for this contribution @Nevelito!

PR #3885, opened 2025-05-26
enhancement: grouped options for select

adrianthedev:
# Description

Add `grouped_options` for select, check [docs](https://docs.avohq.io/3.0/fields/select.html#grouped-options) for more info.

image

# Checklist:


adrienpoly:
Thanks @adrianthedev for drafting this — this would be a great addition to the Select field!

If I understand the proposed API correctly, it requires explicitly specifying grouped_options instead of the current options. That definitely makes the API very clear and explicit.

That said, I was wondering if it might be more idiomatic to continue using the existing options parameter and add a bit of logic under the hood to automatically detect whether the provided options are grouped. IMHO this would make the API simpler and possibly more intuitive.

What do you think?

adrienpoly:
I agree with your points and it is a good call to point that Rails is keeping this as separate methods

> Let me know if I'm missing a strong case where this abstraction would be genuinely beneficial.

The only thing I can think of is that is was the first thing I tried (pass a nested hash to options) without looking into the documentation (RTFM 😄 ).

PR #3890, opened 2025-05-27
fix: make country field work with country attributes

manuelmeurer:
# Description

In my app, I use `composed_of` to convert the database value of a `country` attribute to an `ISO3166::Country` like this:

```ruby
composed_of :country,
  class_name: ISO3166::Country.to_s,
  mapping:    %w[country alpha2],
  converter:  ISO3166::Country.method(:new)
```

This doesn't work with the `country` field though, since it will compare the value from `select_options` (string) with the attribute (ISO3166::Country) and thus not mark the selected country as the selected option in the dropdown.

To fix this, the `value` method must return the `alpha2` short code if it finds a ISO3166::Country.

# Checklist:

I haven't added any tests or updated the docs yet, since this is my first PR, I wanted to get some quick feedback if it looks ok and will be accepted if tests are added and docs are updated. 😀

## Manual review steps

1. Opened a model with a country attribute in edit view
2. Verified that the correct country is selected in the dropdown

### DIFF
--- a/lib/avo/fields/country_field.rb
+++ b/lib/avo/fields/country_field.rb
@@ -34,6 +34,13 @@ module Avo
       def options_for_filter
         select_options
       end
+
+      def value
+        value = super
+        value.is_a?(ISO3166::Country) ?
+          value.alpha2 :
+          value
+      end
     end
   end
 end


Paul-Bob (commented):
Thanks for submitting this contribution, @manuelmeurer!

I've never used `ISO3166::Country` myself, is this a common use case?

I wonder if the country field should support this out of the box, or if it's an edge case that should be handled individually.

manuelmeurer:
It's the `countries` gem that [Avo uses as well](https://docs.avohq.io/3.0/fields/country.html), so IMO it makes sense to support it.

Paul-Bob:
> It's the `countries` gem that [Avo uses as well](https://docs.avohq.io/3.0/fields/country.html), so IMO it makes sense to support it.

My question is more about forcing the `alpha2` identifier, why not `alpha3`? Do you see where I'm going?

I see this as a field option: `field :country, as: :country, code: :alpha2`, where each developer can configure the code instead of hardcoding it to `alpha2`.

manuelmeurer:
> My question is more about forcing the alpha2 identifier, why not alpha3? Do you see where I'm going?

Sorry, I don't understand. Avo uses the countries' `alpha2` identifiers [here](https://github.com/avo-hq/avo/blob/main/lib/avo/fields/country_field.rb#L15) as the values of the select field options and as the value that is saved in the database. So if we load a value from the database, and check if it's a `ISO3166::Country` object, we need to return `alpha2` from the `value` method, so it matches the value in the select option list and the correct option is selected.

manuelmeurer:
Ah yeah, maybe I should have explained the `composed_of` part a bit better... 😄
It's built into Rails (`ActiveRecord::Aggregations::ClassMethods`) but I haven't seen it used in many places...
In fact, I stopped using it as well, because it's not really obvious what it does, and it can easily be replaced with this:

```ruby
def country=(value)
  if value.is_a?(ISO3166::Country)
    value = value.alpha2
  end
  super value
end

def country
  ISO3166::Country.new(super)
end
```

So it just converts a string into a value object (`ISO3166::Country` in this case) and back.

> Should Avo support this use case because it's a commonly accepted pattern, or would it be just to accommodate a specific implementation?

Well, even if `composed_of` is not that common, I think value objects are a common pattern, right?
I use them in my apps as much as I can...
But whether it's so common that Avo should support it, I can't answer. 😀
For me personally, it would of course be fine to use `field :country, as: :country, format_using: -> { value.alpha2 }`, if that works.

Thanks for taking the time to dive into this and discuss with me! 🙏

manuelmeurer:
One last comment about `alpha3` - I don't think it will work.
Avo uses the 2 digit country code (i.e., `alpha2`) for the `value` attributes in the `<select>` (returned by `ISO3166::Country.translations` used [here](https://github.com/avo-hq/avo/blob/main/lib/avo/fields/country_field.rb#L15)), so using the 3 digit code via `alpha3` will not select the correct country.
Not trying to prolong the discussion unnecessarily, just wanted to mention this. 😄

PR #3897, opened 2025-05-29
DRAFT feature: nested resources v2

### LINKED ISSUE #3894: Namespaced resources
### Feature
Support defining resources within a module.

e.g.

```ruby
module Avo
  module Resources
    module Authorization
      class Actor < Avo::BaseResource
        # This should not be necessary
        # self.model_class = ::Authorization::Actor

        def fields
          field :id, as: :id
          # ...
        end
      end
    end
  end
end
```

### Current workarounds

https://docs.avohq.io/3.0/resources.html#namespaced-resources

### Additional context

Our application organizes its models under namespaces. Given that Avo doesn't yet support this, this adds cognitive burden to find the right resource.

I'm guessing the routing piece would be the most challenging aspect of this.

Paul-Bob:
DRAFT that fixes #3894

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -74,8 +74,8 @@ GEM
       addressable
     active_median (0.6.0)
       activesupport (>= 7.1)
-    activejob (8.0.2)
-      activesupport (= 8.0.2)
+    activejob (8.0.1)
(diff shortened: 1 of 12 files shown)


Paul-Bob:
#3922

PR #3901, opened 2025-05-30
fix: add URLHelpers inflection

manuelmeurer:
# Description

I added the "URL" acronym in my app and afterwards got this error:

```
NameError: uninitialized constant Avo::UrlHelpers (NameError)

    include Avo::UrlHelpers
               ^^^^^^^^^^^^
Did you mean?  Avo::URLHelpers
```

I found the PR #2518 that addressed this but was abandoned.

Are you still open to adding this?
If so, I can add documentation and tests.

# Checklist:

## Manual review steps

1. Add "URL" acronym in host app
2. Run with Avo 3.20.x -> NameError: uninitialized constant Avo::UrlHelpers (NameError)
3. Run with this PR -> no error

### DIFF
--- a/app/components/avo/tab_switcher_component.rb
+++ b/app/components/avo/tab_switcher_component.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class Avo::TabSwitcherComponent < Avo::BaseComponent
-  include Avo::UrlHelpers
+  include Avo::URLHelpers
   include Avo::ApplicationHelper
 
   delegate :white_panel_classes, to: :helpers
--- a/app/controllers/avo/base_application_controller.rb
+++ b/app/controllers/avo/base_application_controller.rb
@@ -9,7 +9,7 @@ module Avo
     include Avo::InitializesAvo
     include Avo::CommonController
     include Avo::ApplicationHelper
-    include Avo::UrlHelpers
+    include Avo::URLHelpers
     include Avo::Concerns::Breadcrumbs
     include Avo::Concerns::FindAssociationField
 
--- a/app/helpers/avo/url_helpers.rb
+++ b/app/helpers/avo/url_helpers.rb
@@ -1,5 +1,5 @@
 module Avo
-  module UrlHelpers
+  module URLHelpers
     def resources_path(resource:, keep_query_params: false, **args)
       return if resource.nil?
 
--- a/lib/avo.rb
+++ b/lib/avo.rb
@@ -6,6 +6,7 @@ require_relative "avo/engine" if defined?(Rails)
 loader = Zeitwerk::Loader.for_gem
 loader.inflector.inflect(
   "html" => "HTML",
+  "url_helpers" => "URLHelpers",
   "uri_service" => "URIService",
   "has_html_attributes" => "HasHTMLAttributes"
 )


Paul-Bob:
> Are you still open to adding this?

Hi @manuelmeurer, yes, we are still open to adding this.

If we can introduce an inflection that only affects `Avo::UrlHelpers` and ensure that it does not interfere with the parent application, I'd be happy to merge that PR.

These are the guidelines the PR should follow:

- [ ] Ensure the added inflection only affects the Avo namespace.
- [ ] Add a class with the same name (without the inflection) in the dummy app to confirm it doesn’t break, and test against it.
- [ ] Document this as a breaking change, as someone might be using that module elsewhere in their application.

Thanks for looking into this!

manuelmeurer:
Alright, I'm a bit busy at work right now, but I'll try to get it done soon! 🤞

Paul-Bob:
Manuel, this PR was automatically closed due to inactivity. Feel free to reopen it whenever you're ready to resume work on it. Thanks!

manuelmeurer:
Sorry @Paul-Bob, I didn't get around to working on this yet... :(
Hopefully soon. If anybody else wants to grab it, please go ahead!

Paul-Bob:
All right, no need to apologize

manuelmeurer:
Hmm, seems like I cannot reopen this PR.

So I did some more digging (with the help of Claude) to figure out what's going on.

The way I understand it, the `"url_helpers" => "URLHelpers"`inflection in the Zeitwerk loader (`lib/avo.rb`), which I added in this PR's commit, doesn't work because - in contrast to the other three inflections configured there - `url_helpers.rb` is not locaed in `lib/avo` but in `app/helpers/avo`. This means that it's not really part of the code of the gem (which the Zeitwerk loader and inflections handle) but it's loaded as part of the host app, which the autoloader of the host app handles.
To be able to use `URLHelpers` instead of `UrlHelpers`, we'd have to add

```ruby
Rails.autoloaders.main.inflector.inflect(
  "url_helpers" => "URLHelpers"
)
```

e.g. in the `avo.autoload` initializer in `engine.rb` (I added this in a commit in this PR's branch, but it doesn't show up here in the PR, probably because the PR is closed).

This is not a viable option IMO, since it changes the host app's inflections, which is unexpected and will possibly break some apps that happen to use the `UrlHelpers` constant.

Sooo.... I'm a bit at a loss as to what to do here. 😄 I really need the Avo helper to be `URLHelpers` in the apps I use Avo in, since they all have the `URL` acronym configured and therefore complain when they encounter `Avo::UrlHelpers`.
Maybe we could rename that helper to not include URL/Url at all?

But also, maybe I'm overlooking some obvious way to solve this? 😅

Paul-Bob:
Here’s how I see it: Avo’s inflections shouldn’t affect the host app, and the host app’s inflections shouldn’t affect Avo. For example, if your app uses the "URL" acronym, there should be a way to isolate it from Avo, just like if Avo uses another acronym, it shouldn’t affect your app.

The problem right now is that acronyms from the host app are affecting Avo components. What you said about `app/helpers/avo` being loaded by the parent app kind of makes sense. Maybe we can specify that directory to be autoloaded by Avo’s loader instead of the host app?

I played around with Cursor trying to figure this out, and I ended up with this setup. It’s not mergeable, but it fixes the isolation issue. Host app inflections don’t affect Avo anymore in this commit:  
https://github.com/avo-hq/avo/commit/02a76c0b2edfdc92df369c3e28fa843302e3cfa1

The next step would be to find what exactly causes this in that commit and make a much more simplified PR.

manuelmeurer:
FYI: I eventually "solved" this by adding `Rails.autoloaders.main.inflector.inflect("url_helpers" => "UrlHelpers")` to my own app instead of changing the module name in Avo. If I ever need that module in my app, I need to call it `UrlHelpers` as well, but I think this is an acceptable compromise. 😄

Paul-Bob:
I also found something related on a stack overflow question

https://stackoverflow.com/questions/78239116/rails-7-1-resolve-conflicting-zeitwerk-inflection-rules/78254934#78254934

hmk:
Also came across this issue. I was able to fix temporarily by moving my app level inflections into the app (rather than global).

```ruby
 Rails.autoloaders.each do |autoloader|
      autoloader.inflector.inflect(
        "url" => "URL",
        # other inflections here
      )
 end
 ```

But :+1: to this comment from @Paul-Bob:

>If we can introduce an inflection that only affects Avo::UrlHelpers and ensure that it does not interfere with the parent application, I'd be happy to merge that PR.

If I have time, will ship a PR. Still an active issue.

hmk:
@Paul-Bob - just opened a PR to fix this issue. 

I chose to special case URL/Url in a way that other acronyms could easily be addressed in the future - but that doesn't monkeypatch inflections or build a new inflector.

I worry that monkeypatching inflections (or building an entirely new inflector) as discussed in that StackOverflow may collide with Zeitwerk in the future in a way that's harder to detect.

If the only reports of collisions so far are in URL/Url, this #4204 feels like a good approach, but I'm open to alternative thoughts.

hmk:
> I chose to special case URL/Url in a way that other acronyms could easily be addressed in the future - but that doesn't monkeypatch inflections or build a new inflector.

Actually... on second thought I like the approach I took even less.

Going to explore how other gems deal with inflections / see if inspiration hits.

Paul-Bob:
Hi @hmk, I saw the PR. I will respond here to keep all the context in one place.

First of all, thanks for looking into this and taking the time to open a contribution. That PR does fix the `URL` edge case, but it leaves the issue open for other inflections.

I did not test whether the solution from the PR works when it is applied in a parent app, but I think it should work whenever someone is facing this specific inflection.

I agree with you that the double definition is not ideal.

> I worry that monkey patching inflections or building an entirely new inflector, as discussed in that Stack Overflow post, may collide with Zeitwerk in the future in a way that is harder to detect.

In my opinion, it would be great if this could be added as `ActiveSupport::Inflector` native support, such as inflection isolation per engine. I am not sure if that is easy to achieve.

PR #3935, opened 2025-06-11
feature: Add `timeout` option for feedback notifications toasts

### LINKED ISSUE #3934: Prevent automatic dismissal for specific feedback toasts
### Feature
This feature would introduce an option to keep feedback notifications (toasts) open until the user manually closes them.

Currently, the dismiss time is a global configuration (config.alert_dismiss_time). The proposed solution is to add a per-call option, like `keep_open: true`, to the notification helpers (inform, success, error, warning).

Example:

`inform "The action has completed and the report is ready.", keep_open: true`

This would give developers more granular control, allowing specific important messages to persist while others follow the default auto-dismiss behaviour.

### Current workarounds
The only current workaround is to set a very high value for `config.alert_dismiss_time` in the Avo initializer. This is not ideal because it changes the behaviour for all notifications throughout the application, whereas the goal is to control it only for specific instances.

### Additional context
This feature is for cases where an action returns an important result that the user needs to acknowledge or read carefully. Forcing it to close after a set time can lead to a poor user experience, as the user might miss the information. Letting the user decide when to close the toast for these specific notifications would be a significant improvement.

icaroryan:
# Description

This PR introduces a flexible `timeout` option for flash messages, allowing developers to control how long notifications remain visible or if they should stay open indefinitely.

Previously, all flash messages would auto-dismiss based on the global `config.alert_dismiss_time` setting. This new feature provides granular control for customizing notification display times or keeping important messages visible until manually closed.

**Usage Examples**
```ruby
# Keep message open indefinitely (won't auto-dismiss)
succeed 'This will not close automatically', timeout: :forever

# Custom timeout in milliseconds
succeed 'Closes after 5 seconds', timeout: 5000

# Default behavior (uses global configuration)
succeed 'Uses default timeout'
```

Fixes #3934

# Checklist:

### TLDR
ezgif-68c899075a222f

### DEMO
https://www.loom.com/share/d7f9658358a94821a770182e106418a9

## Manual review steps

1. Test different timeout options in any feedback notification:
  * `timeout: :forever` - Should only close when user clicks the 'x' button
  * `timeout: 3000` - Should auto-dismiss after 3 seconds
  * No timeout specified - Should use the default from configuration (`alert_dismiss_time`)
2. Ensure there are no regressions with other toasts

### DIFF
--- a/app/components/avo/alert_component.rb
+++ b/app/components/avo/alert_component.rb
@@ -60,6 +60,6 @@ class Avo::AlertComponent < Avo::BaseComponent
   end
 
   def keep_open?
-    @timeout&.try(:to_sym) == :forever
+    @timeout.try(:to_sym) == :forever
   end
 end
--- a/spec/components/avo/flash_alerts_component_spec.rb
+++ b/spec/components/avo/flash_alerts_component_spec.rb
@@ -18,7 +18,7 @@ RSpec.describe Avo::FlashAlertsComponent, type: :component do
       # Should auto-dismiss with default timeout
       alert_with_default_timeout = page.all("div[data-controller='alert']")
       alert_with_default_timeout.each do |alert|
-        expect(alert['data-alert-dismiss-after-value']).to eq Avo.configuration.alert_dismiss_time.to_s
+        expect(alert["data-alert-dismiss-after-value"]).to eq Avo.configuration.alert_dismiss_time.to_s
       end
     end
   end
@@ -38,7 +38,7 @@ RSpec.describe Avo::FlashAlertsComponent, type: :component do
 
       # Should auto-dismiss with custom timeout
       alert_with_custom_timeout = page.all("div[data-controller='alert']").find { |alert| alert.text.include?("Message with custom timeout") }
-      expect(alert_with_custom_timeout['data-alert-dismiss-after-value']).to eq "10000"
+      expect(alert_with_custom_timeout["data-alert-dismiss-after-value"]).to eq "10000"
     end
   end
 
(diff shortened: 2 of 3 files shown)


icaroryan:
TODO: Update documentation with new option

Paul-Bob (approved):
Awesome @icaroryan!!

Paul-Bob (changes requested):
Instead of `keep_open`, let's switch to a `timeout` option, which directly translates to `alert_dismiss_time` when an integer is provided.

`timeout: :forever` => no `alert_dismiss_time`
`timeout: 9000` => `alert_dismiss_time = 9000`

This adds the enhancement that each developer can fine tune the alert dismiss time directly from the flash message.

icaroryan:
@Paul-Bob I'm all in with your suggestion!! It kills two birds with one stone
Thank you for the feedback!

app/components/avo/flash_alerts_component.rb
```diff
@@ -3,5 +3,18 @@
...
+  private
+
+  def parse(data)
+    case data
+    when Hash
+      {body: data[:body], timeout: data[:timeout]}
+    else
+      {body: data, timeout: nil}
+    end
+  end
```
icaroryan, line 19: I wonder if there's a better way to do this

app/components/avo/alert_component.rb
```diff
@@ -52,4 +53,13 @@ def is_warning?
   def is_empty?
     @message.nil?
   end
+
+  def timeout
+    return @timeout if @timeout.is_a?(Numeric)
+    Avo.configuration.alert_dismiss_time
```
icaroryan, line 59: Right now, if the value is invalid, we silently default it to `alert_dismiss_time`. Is this fine, or would an actual validation be more appropriate?

app/components/avo/flash_alerts_component.rb
Paul-Bob: I wondered the same first time I saw it, but not sure if we can simplify it

app/components/avo/alert_component.rb
Paul-Bob: It's all right, no need for strict validations here

Paul-Bob (approved):
LGTM

icaroryan:
AWESOME!

I also just updated the documentation with this new option

https://github.com/avo-hq/docs.avohq.io/pull/401

PR #3939, opened 2025-06-12
Fix disabled and readonly on trix field

zhephyn:
# Description
The Trix field has no straightforward way to render disabled input. To bypass this, the suggested approach was to render the Trix show component within the edit form if either ```disabled``` or ```readonly``` is set to true. 

Show page
image

Edit page
image

The body of the event is shown instead of the Trix field editor on the Edit page when ```disabled``` or ```readonly``` are set to true like below:

```ruby
field :body, as: :trix, disabled: true
```
Fixes # (3888)

# Checklist:

### DIFF
--- a/app/components/avo/fields/trix_field/edit_component.html.erb
+++ b/app/components/avo/fields/trix_field/edit_component.html.erb
@@ -1,25 +1,48 @@
-<%= field_wrapper(**field_wrapper_args) do %>
-  <%= content_tag :div,
-    class: class_names("relative block overflow-x-auto max-w-4xl", @input_id),
-    data: do %>
-    <%= content_tag 'trix-editor',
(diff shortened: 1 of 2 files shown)


zhephyn:
@Paul-Bob I've laid out an initial approach to fix the Trix field readonly bug

However currently the ```edit_component.html.erb``` file could use some refactoring, specifically the show component duplication. I was thinking of extracting the show component into a partial but wanted to run it by you first and see if you think its a sustainable approach. 

Lmk what you think. Thanks

zhephyn:
I've added tests for this as well.

zhephyn:
@Paul-Bob New changes are up. Kindly take a look whenever its convenient for you.

PR #3942, opened 2025-06-16
Fix non-ASCII string in tab name

### LINKED ISSUE #1908: non-ASCII string in Tab name

zhephyn:
# Description
Allows turbo frames of different ids to be generated for non-ASCII strings such as those including emojis. 

If we have the tabs below, we expect each of them to have a unique turbo frame id:
```ruby
  tabs do
    field :stores, as: :has_many, name: '📖 store'
    field :addresses, as: :has_many, name: 'store 🧭'
    field :others, as: :has_many, name: '其他store '
  end
```
Currently, the turbo frame id generated for all the 3 will be the same, therefore when we switch between the tabs, content rendering will be messed up. 

Fixes #1908

# Checklist:

### DIFF
--- a/lib/avo/resources/items/tab.rb
+++ b/lib/avo/resources/items/tab.rb
@@ -34,8 +34,8 @@ class Avo::Resources::Items::Tab
   alias_method :to_param, :id
 
   def turbo_frame_id(parent: nil)
-    id = "#{Avo::Resources::Items::Tab.to_s.parameterize} #{name}".parameterize
-
+    digest_name = Digest::MD5.hexdigest(name)
+    id = "#{Avo::Resources::Items::Tab.to_s.parameterize} #{digest_name}".parameterize
     return id if parent.nil?
 
     "#{parent.turbo_frame_id} #{id}".parameterize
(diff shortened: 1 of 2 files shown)


zhephyn:
Hey @Paul-Bob I could use your review on this one as well whenever its convenient for you. Thanks!

zhephyn:
@Paul-Bob I've added a test to confirm that switching between tabs renders the correct turbo frame content. 

With regards to the failing tests, surprisingly, all of them pass when i run them locally on my machine and even when i use the Github CLI. Could you take a look when you get some time? Thanks!

In case you need screenshots or a video to show you that the tests pass for me, please lmk and I'll have it by you ASAP.

zhephyn:
@Paul-Bob Thanks for the fix. I can't believe I forgot the restore_items_from_backup. 

Is there any other change I should make to the PR apart from the test or it looks good so far?

Paul-Bob:
> @Paul-Bob Thanks for the fix. I can't believe I forgot the restore_items_from_backup.
> 
> Is there any other change I should make to the PR apart from the test or it looks good so far?

Without the `restore_items_from_backup` it was leaking state to next tests. It happened to me some times that's why I quickly identified it 😄 

No, no other changes required, thank you for this contribution!

PR #3943, opened 2025-06-16
Fix mobile field label text wrapping with responsive width classes

### LINKED ISSUE #2588: Poor use of horizontal/vertical space at smaller breakpoints.
### Describe the bug

For field names, the default width is `w-48`, with a breakpoint rule of `md:w-64`.

Unfortunately, for long field names, this presents a problem as certain field labels are relatively long and will get broken onto two lines on mobile.

You might consider this instead: `w-full md:w64`  since the field name and value aren't on the same row when rendering that small anyway.

### Steps to Reproduce
1. Add this configuration '...'
2. Go to '...'
3. Click on '....'
4. Scroll down to '....'
5. See error

### System configuration
**Avo version**: 3.4.4

**Rails version**: 7.1.3.2

**Ruby version**: 3.2.2

**License type**:

 - [ ] Community
 - [ ] Pro
 - [x] Advanced

 - [x] No

#### Impact

 - [x] Low impact (It's really a tiny thing that I could live with.)

#### Urgency

AhmedNadar:
### Problem
Long field names were wrapping to multiple lines on mobile due to the fixed `w-48` are applied to field labels at all screen sizes. This created poor readability and inconsistent spacing on mobile.

### Solution
Implemented a mobile-first responsive approach:

- **Mobile**: We use `w-full` for all field labels since they're stacked anyway
- **Desktop**: Apply contextually appropriate fixed widths:
  - Stacked layouts: Continue using `w-full` 
  - Inline + Compact: Use `md:w-48`
  - Inline + Regular: Use `md:w-64`

Fixes #2588

### DIFF
--- a/app/components/avo/field_wrapper_component.html.erb
+++ b/app/components/avo/field_wrapper_component.html.erb
@@ -1,6 +1,6 @@
 <%= content_tag :div,
   class: classes,
-  style: style,  
+  style: style,
   data: data do %>
   <%= content_tag :div, class: class_names("pt-4 flex self-start items-center flex-shrink-0 w-full px-6 uppercase font-semibold text-gray-500 text-sm", @field.get_html(:classes, view: @view, element: :label), {
     "md:pt-4": stacked?,


app/components/avo/field_wrapper_component.html.erb
```diff
@@ -1,14 +1,14 @@
 <%= content_tag :div,
   class: classes,
-  style: style,
+  style: style,
```
Paul-Bob:
```suggestion
  style: style,
```

Paul-Bob (approved):
LGTM!

PR #3961, opened 2025-07-08
Add ``` hide_filter_button ``` for associations

### LINKED ISSUE #2884: Add `hide_filter_button` for associations
Add `hide_filter_button` for `has_many` and `has_and_belongs_to_many` associations, similar to how the [`hide_search_input`](https://docs.avohq.io/3.0/associations/has_many.html#hide_search_input) works. I'll try to do it (some day).

zhephyn:
# Description
This PR allows us to hide the ```filter``` button similar to how we are able to hide the``` search ``` input.

Fixes #2884 

# Checklist:

### DIFF
--- a/app/components/avo/filters_component.html.erb
+++ b/app/components/avo/filters_component.html.erb
@@ -1,17 +1,19 @@
 <div data-controller="toggle" data-component-name="<%= self.class.to_s.underscore %>">
   <div class="relative w-full flex justify-between">
-    <%= a_button class: 'focus:outline-none',
-      color: :primary,
-      size: :sm,
-      icon: "avo/filter",
-      title: t('avo.click_to_reveal_filters'),
-      'data-button': 'resource-filters',
-      'data-action': 'click->toggle#togglePanel',
-      'data-tippy': 'tooltip' do %>
-      <%= t 'avo.filters' %>
-      <% if @applied_filters.present? %>
-        <span class="ml-1">(<%= "#{@applied_filters.count} #{t('avo.applied', count: @applied_filters.count)}" %>)</span>
-      <% end %>
+    <% unless @field&.hide_filter_button %>
+      <%= a_button class: 'focus:outline-none',
+        color: :primary,
+        size: :sm,
+        icon: "avo/filter",
+        title: t('avo.click_to_reveal_filters'),
+        'data-button': 'resource-filters',
+        'data-action': 'click->toggle#togglePanel',
+        'data-tippy': 'tooltip' do %>
+       <%= t 'avo.filters' %>
+        <% if @applied_filters.present? %>
+          <span class="ml-1">(<%= "#{@applied_filters.count} #{t('avo.applied', count: @applied_filters.count)}" %>)</span>
+        <% end %>
+       <% end %>
     <% end %>
     <div
       class="absolute block inset-auto sm:right-0 top-full bg-white min-w-[300px] mt-2 z-40 shadow-modal rounded divide-y divide-gray-300 <%= 'hidden' unless params[:keep_filters_panel_open] == '1' %>"
--- a/app/components/avo/views/resource_index_component.rb
+++ b/app/components/avo/views/resource_index_component.rb
@@ -132,6 +132,7 @@ class Avo::Views::ResourceIndexComponent < Avo::ResourceComponent
   def render_dynamic_filters_button
     return unless Avo.avo_dynamic_filters_installed?
     return unless @resource.has_filters?
+    return unless show_filters_button?
     return if Avo::DynamicFilters.configuration.always_expanded
 
     a_button size: :sm,
@@ -144,6 +145,15 @@ class Avo::Views::ResourceIndexComponent < Avo::ResourceComponent
       } do
       Avo::DynamicFilters.configuration.button_label
     end
+  end
+
+  def show_filters_button?
+    return false if field&.hide_filter_button
+    return false unless Avo.avo_dynamic_filters_installed?
+    return false unless @resource.has_filters?
+    return false if Avo::DynamicFilters.configuration.always_expanded
+
+    true
   end
 
   def scopes_list
--- a/lib/avo/fields/many_frame_base_field.rb
+++ b/lib/avo/fields/many_frame_base_field.rb
@@ -6,6 +6,7 @@ module Avo
 
       attr_reader :scope,
         :hide_search_input,
+        :hide_filter_button,
         :discreet_pagination
 
       def initialize(id, **args, &block)
@@ -28,6 +29,7 @@ module Avo
         @searchable = args[:searchable]
         @scope = args[:scope]
         @hide_search_input = args[:hide_search_input]
+        @hide_filter_button = args[:hide_filter_button]
         @discreet_pagination = args[:discreet_pagination]
       end
     end


app/components/avo/views/resource_index_component.rb
```diff
@@ -146,6 +146,12 @@ def render_dynamic_filters_button
     end
   end
 
+  def show_filters_button?
+    return false if field&.hide_filter_button
+
+    true
+  end
```
Paul-Bob: Where is this method being used?

Paul-Bob (commented):
Seems all right, not sure why the new `show_filters_button?` method is needed and this PR needs docs + tests

Thank you @zhephyn for looking into this!

app/components/avo/views/resource_index_component.rb
zhephyn: Oh it was being called inside the render_dynamic_filters_button method, however I removed that line as the cognitive complexity for the method exceeded 5. 

Think we can eliminate the ```show_filters_button?```method entirely, though not sure where we can relocate this line for the functionality to work ``` return false if field&.hide_filter_button```

zhephyn: Gonna push the tests shortly

zhephyn:
I'm gonna push the tests shortly. Thereafter, I'll work on the docs

zhephyn:
@Paul-Bob Tests are now up. Also, in addition to the test for the ```hide_filter_button```, I added a test for the ```hide_search_input``` option after going through the code base and not finding a corresponding test for it. So i pushed a total of 2 tests. 

Also, the documentation PR is up as well on the documentation repository. 

Could use a review whenever you are free. Thanks again!

app/components/avo/views/resource_index_component.rb
Paul-Bob: Got it, inside `render_dynamic_filters_button` we can add this line to the guards, `return if field&.hide_filter_button` and we can remove this `show_filters_button?`method since is not used anywhere.

app/components/avo/filters_component.html.erb
```diff
@@ -1,17 +1,19 @@
...
+        icon: "avo/filter",
+        title: t('avo.click_to_reveal_filters'),
+        'data-button': 'resource-filters',
+        'data-action': 'click->toggle#togglePanel',
+        'data-tippy': 'tooltip' do %>
+       <%= t 'avo.filters' %>
+        <% if @applied_filters.present? %>
+          <span class="ml-1">(<%= "#{@applied_filters.count} #{t('avo.applied', count: @applied_filters.count)}" %>)</span>
+        <% end %>
+       <% end %>
```
Paul-Bob: Rendering the component consumes computing resources. If you observe closely, the component still renders all filters in a hidden state and simply omits rendering the button that would reveal the filters. A more efficient approach would be to prevent the filters component from rendering entirely when the field has `hide_filter_button` configured. This can be achieved by returning `false` from the `render?` method in the `Avo::FiltersComponent` class.

spec/features/avo/has_many_field_spec.rb
```diff
@@ -170,4 +170,34 @@
...
+          field :patrons, as: :has_many, through: :patronships, hide_search_input: true
+        end
+
+        visit "/admin/resources/stores/#{store.id}"
+        expect(page).not_to have_css('[data-search-target="autocomplete"]')
+
+        Avo::Resources::Store.restore_items_from_backup
+      end
+    end
+  end
```
Paul-Bob: The tests are well-written, but after manual testing, it appears the behavior is not consistently working as expected. Let's update the tests to ensure that when `hide_filter_button` and `hide_search_input` are *not* configured, the components are rendered. When those options are explicitly set to `true`, the components should not be present (this is already tested). Otherwise, we risk encountering false positives in the test results.

app/components/avo/filters_component.html.erb
Paul-Bob: Instead, you could perform a conditional check [here](https://github.com/avo-hq/avo/blob/main/app/components/avo/views/resource_index_component.html.erb#L49). Note that  there, you should have access to the `@field`. However, inside the filters component, `@field` is always `nil` unless you explicitly add it as a prop and pass it during component initialization. So, I suspect this isn't actually working as intended currently. I tested this on the Project resource, specifically on the Users field, and the filters are still being rendered.

Paul-Bob (changes requested):
After some manual testing, it seems this isn't currently working as expected. I've left detailed comments on how to get it working properly. Thanks, @zhephyn!

zhephyn:
@Paul-Bob I've pushed some tweaks based on the changes you requested though unfortunately, the functionality is still yet to work. Think i could use your eyes on what is potentially the issue. Thanks!

zhephyn:
Also, when i add the line below:
```ruby
return if @field&.hide_filter_button
```
to the ```render_dynamic_filters_button``` method, i get a code climate flag telling me that the complexity of the method exceeds the limit(5)

Paul-Bob:
Thanks for all the effort here @zhephyn !

When testing locally this was still not working, the new option was taking no effect.

Since the back and forward on the review was getting bigger I went ahead and made the necessary changes to make it work and finish the PR.

Tweaks I made to wrap up the PR:

- [Reverted wrong indentation modification](https://github.com/avo-hq/avo/pull/3961/commits/d3378c04f74083686b4dc92a47eed6232f651a0b)
- [Simplified the view component render conditionals](https://github.com/avo-hq/avo/pull/3961/commits/3f025c7478b906bae03f955f497bf9e9e3c64e6b)
- [Removed duplicated and redundant conditional to render the filters component](https://github.com/avo-hq/avo/pull/3961/commits/0845b9ad4b544d5a6eb5b98c9bb61c537a7dde0a)
- [Improved test to address my last comment about tests](https://github.com/avo-hq/avo/pull/3961/commits/54c201cfa4f238d7c97a6626b1eca054059cbfdc)
- [Fixed the tests to work on the correct view](https://github.com/avo-hq/avo/pull/3961/commits/6d26cde2ed659e4fa9c70a22ba5cc871eddea092)
- [Used `field` instead `@field` on conditional, `@field` was always `nil`, this was causing the option to take no effect](https://github.com/avo-hq/avo/pull/3961/commits/7609d75c07499d9768417a457762c4094f7afe3f)

>  i get a code climate flag telling me that the complexity of the method exceeds the limit(5)

That's ok, the method is still readable without any effort.

Thanks once again for all your work in this PR, docs are on point!

Paul-Bob (approved):
LGTM

zhephyn:
Hey Paul. Thanks so much for picking up from where I left off. I had wanted to add the other tests you had requested after confirming that the feature now works. Anyway, I think my biggest problem now is "how to debug" especially when I'm in a view or a component. I kept on trying but I couldn't figure out whether to use field or @field and my attempt at debugging didn't help either. Thanks again!

Paul-Bob:
> Hey Paul. Thanks so much for picking up from where I left off. I had wanted to add the other tests you had requested after confirming that the feature now works. Anyway, I think my biggest problem now is "how to debug" especially when I'm in a view or a component. I kept on trying but I couldn't figure out whether to use field or @field and my attempt at debugging didn't help either. Thanks again!


Got it. How I usually debug is by first navigating to the page I want to test, in this case "/admin/resources/stores/#{store.id}/patrons". Once I'm there, I added a `<% raise %>` just before the component render, where you used `@field`, and I noticed that `@field` is `nil`. This was unexpected, since rendering the component from the link above should include the association field.

I noticed that `hide_search_input` works as expected, so I went ahead and looked into how `hide_search_input` is implemented. I saw that it uses `field&.hide_search_input`, and I found out that this `field` is actually a method on the base component:
https://github.com/avo-hq/avo/blob/main/app/components/avo/base_component.rb#L22

TBH It's not very intuitive to refer to the association field as `field`, especially since it only renders in association views but is defined in the base component. That method definition dates back about three years, and it could use a refactor to make it more understandable. But I guess changing the name or location of that method now could potentially open a Pandora's box, so it's probably best to leave it for now.

zhephyn:
Thanks for the debugging tip Paul. Gonna try it out for my next issue🙏

PR #3963, opened 2025-07-09
enhancement: introduce q for searching and strip it

adrianthedev:
# Description

Fixes the issue of running searches with `LIKE` and untripped values (` something` vs `something`).

cc: @mensfeld

# Checklist:

- [-] I have made corresponding changes to the [documentation](https://github.com/avo-hq/avodocs)

## Manual review steps

1. Try a search in course for ` name`. it should find the record
1. Try the same thing in course link and it shouldn't find the record


Paul-Bob:
Tweaks done before merge:

1. Removed the safe navigator on `params.strip`, reason: `q` should always be present on search requests. If it's not, we don't want silent errors, we want errors to be raised and easily identified. I even found some tests that were not passing the `q` param as the app always does, so I've already identified some tests that behave differently from what the app in the browser does, already a win. Even if that doesn't make a difference in this situation, the test request should ALWAYS match the requests that the app makes in real scenarios.

2. Modified test to check if `q` is stripped in the query, that is what we want to test. There is no reason to complicate tests with involved queries. Reason: minimize test time as much as possible.

mensfeld:
@adrianthedev :saluting_face: :handshake: 
image

PR #3969, opened 2025-07-10
feature: Add indeterminate state for boolean field

### LINKED ISSUE #3873: Boolean field to show indeterminate state
### Context

True or false is good, but showing an undetermined state might also be very useful. Maybe a grey color with a minus-circle or exclamation-circle heroicon?
Not sure what value should the record output to show that, but we'll figure it out.

AmiasYaska:
This change introduces a new nil_as_indeterminate option to the BooleanField to provide a more distinct visual state for nil values.

When this option is set to true, the field will display a grey minus-circle heroicon instead of the default em dash (—).

Fixes #3873

I have used city.rb as an example since it has is_capital which is boolean. When nil_as_indeterminate is made true, the icon is shown under is_capital: 

field :is_capital, as: :boolean, filterable: true, nil_as_indeterminate: true

This is how it appears:

image

### DIFF
--- a/app/components/avo/fields/boolean_field/index_component.html.erb
+++ b/app/components/avo/fields/boolean_field/index_component.html.erb
@@ -1,3 +1,7 @@
 <%= index_field_wrapper(**field_wrapper_args, flush: true) do %>
-  <%= render Avo::Fields::Common::BooleanCheckComponent.new checked: @field.value %>
+  <% if @field.value == :indeterminate %>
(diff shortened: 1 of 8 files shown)


AmiasYaska:
Thanks. Let me make those changes

Paul-Bob:
Hi @AmiasYaska  

Are you planning to make any progress on this PR in the near future?  
No pressure, just cleaning up the queue and came across it.

Paul-Bob:
Closing to keep the queue clean

PR #3971, opened 2025-07-10
refactor: remove sprockets and use propshaft

adrianthedev:
# Description

Removes sprockets in favour of propshaft.

Changes made:

- removed sprockets
- added propshaft
- moved `avo.base.css` to `avo/application.css`
- moved `avo.base.js` to `app/javascript/application.js`
- exporting the CSS to `app/builds/avo/*`
- exporting the JS to `app/builds/avo/*`
- 👆this means that we can reference the files as `avo/application.js` and `avo/application.css`
- moved fonts and the rest of the assets to `app/avo/images`
- removed the `public/avo-assets` middleware and directory
- we are relying on Rails' default pipeline and conventions
- removed rails 6.1 from appraisals as it doesn't support propshaft
- removed `prod` build tasks as they are the same with development

### To-do:
- [ ] check the `contributing.md` doc for misinformation
- [ ] update the docs on how we treat the asset pipeline (mostly path updates)
- [ ] mention in the docs that from now on Avo's assets are passing through the asset pipeline so we have fingerprinting and CDN support for those who need it.
- [x] fix the tests


adrianthedev:
@Paul-Bob can you please have a look at the tests?
not sure if they are related to this PR (probably not) or if we are green on that branch.
if everything is ok, it's good to merge.

Thank you ❤️

PR #4005, opened 2025-08-02
Fix: Country field copies wrong clipboard value when ```copyable: true```

### LINKED ISSUE #3938: Fix `country` fields wrong clipboard value when `copyable: true`
### Context

`field :country, as: :country, copyable: true` 

For the country field, when the `copyable` option is enabled, the value copied to the clipboard is the country code (the database value) instead of the country label.

For example, `PL` is copied instead of `Poland`.

zhephyn:
Fixes #3938 

# Checklist:

### DIFF
--- a/app/components/avo/field_wrapper_component.html.erb
+++ b/app/components/avo/field_wrapper_component.html.erb
@@ -33,7 +33,14 @@
             <%= content %>
           <% end %>
           <% if @field.copyable %>
-            <%= render Avo::ClipboardComponent.new(value: @field.value) %>
+            <% 
+              copy_value = if @field.is_a?(Avo::Fields::CountryField) && !@field.display_code
+                @field.countries[@field.value]
+              else
+                @field.value
+              end
+            %>
+            <%= render Avo::ClipboardComponent.new(value: copy_value) %>
           <% end %>
         </div>
       <% elsif on_edit? %>
--- a/app/components/avo/index/field_wrapper_component.html.erb
+++ b/app/components/avo/index/field_wrapper_component.html.erb
@@ -17,7 +17,14 @@
       <%= content %>
     <% end %>
     <% if @field.copyable %>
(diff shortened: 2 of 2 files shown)


zhephyn:
I had done some extensive digging around on this. Gonna provide a more detailed version for all the potential solutions I had considered. Thanks again for the review

zhephyn:
Some more context.
I've refactored the approach such that we override the ```value``` method in the ```country_field.rb``` file. 
```ruby
      def value
        stored_value = super

        if @display_code
          stored_value
        else
          countries[stored_value]
        end
      end
```
As for the cause of the bug, when i dug around to see why the other fields which utilized the ```copyable``` option were working as expected yet this wasn't the case for the country field, i discovered that a similarity between all these fields was that when ```copyable``` is set to ```true```, the value we copy is the one saved in the database. For the rest of the fields such as the ```text field```, there was no problem with this. 

For the country field however, since we save the country code in the database and not the country name, as per the current implementation of the copyable option, we would always copy the country code and not the country name since this is what was saved in the database. 

By this logic, another approach to solving the bug would be to change the value stored in the database for the country field to be the country name instead of the country code by changing the ```select_options``` method from this:
```ruby
      def select_options
        if @display_code
          countries.map do |code, name|
            [code, code]
          end
        else
          countries.map do |code, name|
            [name, code]
          end
        end
      end
```
to this:

```ruby
      def select_options
        if @display_code
          countries.map do |code, name|
            [code, name]
          end
        else
          countries.map do |code, name|
            [name, name]
          end
        end
      end
```
That allows us to copy the country name without making any other changes in the code base after all, its what is saved in the database. 

Though the issue with this approach is that while country codes are standardized regardless of the language, that is not the case for country names.

zhephyn:
new changes for this are up!

Paul-Bob (approved):
Thanks, @zhephyn! Great explanation and refactor! I probably would have taken the same strategy and approach

PR #4033, opened 2025-08-13
feature: expose parent in default blocks

### LINKED ISSUE #4027: Expose `parent` record to the `default` field option context
### Feature
When we're creating a record from an association, some of the fields can be auto-populated by default by inferring values from the parent record. However, we don't have the `parent` variable exposed to the `default` option block.
It'd be the same as we have with `attach_scope`, where we know who the parent record is by using the variable `parent` from inside the block

### Current workarounds
Created my own `parent_record` in BaseResource

```rb
module Avo
  class BaseResource < Avo::Resources::Base
    def parent_record
      @parent_record ||=
        if params[:via_relation_class].present? && params[:via_record_id].present?
          via_resource = Avo.resource_manager.get_resource_by_model_class(params[:via_relation_class])
          via_resource.find_record params[:via_record_id], params: params
        end
    end
  end
end
```

icaroryan:
# Description

This PR will expose the `parent` record in default blocks when creating records through associations.

```
field :body, as: :textarea, default: -> { parent&.name }
```

Fixes #4027

# Checklist:

## Screenshots & recording

```
field :body, as: :textarea, default: -> { "#{parent.first_name}'s comment" if parent.is_a?(User) }
```

### User Resource
image

### New Comment from User
image

### New Comment (without parent)
image

## Manual review steps

1. Try to create a record from a parent record
2. Check if `parent` is exposed inside the `default` block


icaroryan:
Since we have access to the `parent` in `base_fields`, we could also expose this variable in any other block. So this should also set up some foundation for future work

lib/avo/fields/base_field.rb
```diff
@@ -290,8 +290,24 @@ def meta
...
+        return nil if params.blank?
+
+        via_record_id = params[:via_record_id]
+        via_relation_class = params[:via_relation_class]
+
+        return nil if via_record_id.blank? || via_relation_class.blank?
+        resource = get_resource_by_model_class(via_relation_class)
+        resource.find_record(via_record_id, params: params)
+      end
+
```
Paul-Bob, line 325: I'm wondering if we could execute this logic here, or add an accessor for the parent in the `# Hydrated payload` section, and hydrate the record with the parent when the record is hydrated.

Paul-Bob (commented):
Thanks for this contribution @icaroryan

We need to investigate whether we can hydrate the field with the parent, using the same approach we use on the record

Paul-Bob (approved):
Let's ship this approach, hydrating would be necessary in too many parts.

PR #4042, opened 2025-08-20
Drop disabled_features and add navigation section toggle

Paul-Bob:
# Description

`disabled_features` has been around for more than 4 years, and we only used it for `global_search`. Recently, Adrian added uses for it with `focus_sidebar_menu_item` and `preserve_sidebar_scroll`. I just noticed it's not even documented, and the behavior these options control is actually a good UX improvement, so I'm not sure why anyone would want to turn them off.  

For that reason, we're dropping support for `disabled_features`. This comes after I added a new global search configuration and realized that global search should have a dedicated configuration that wraps both the new option, the enable toggle, and any future options.

```ruby
Avo.configure do |config|
  config.global_search = {
    enabled: true,
    navigation_section: true,
  }
end
```

### DIFF
--- a/app/javascript/js/controllers/sidebar_controller.js
+++ b/app/javascript/js/controllers/sidebar_controller.js
@@ -56,7 +56,7 @@ export default class extends Controller {
     this.attachScrollVisibilityAnchor()
 
     // Restore sidebar scroll position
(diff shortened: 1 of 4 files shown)


adrianthedev:
Yes, let' move he global search disabling to its own configuration item.
thanks for the proactivity

Paul-Bob:
You're welcome

PR #4052, opened 2025-08-29
Join paths resiliently

schmijos:
# Description

I was irritated by avo-generated templates: What if there is no slash in the root path. So I propose to join paths a bit more verbosely.

# Checklist:


adrianthedev:
Love it!
Thanks for taking the time to push a PR

PR #4063, opened 2025-09-12
remove `preload: true`, since it's the default

manuelmeurer:
The `preload` param can be removed, since "true" is the default value.

# Checklist:


manuelmeurer:
From the [docs](https://github.com/rails/importmap-rails#preloading-pinned-modules):

> To avoid the waterfall effect where the browser has to load one file after another before it can get to the deepest nested import, importmap-rails uses [modulepreload links](https://developers.google.com/web/updates/2017/12/modulepreload) by default. If you don't want to preload a dependency, because you want to load it on-demand for efficiency, append preload: false to the pin.

PR #4069, opened 2025-09-17
Resolve herb crash

adrienpoly:
# Description

While trying to use [ReactionView at RubyEvents](https://github.com/rubyevents/rubyevents/pull/992) I encountered some parsing error with Avo. 
Some a questionable and should probably be resolved at some point in Herb one was legitimate.
This PR updates 2 components ERB so that ReactionView compiles successfully in production environment

# Checklist:

## Manual review steps

1. create a project with AVO and ReactionView
1. Start it in Production Environement

### DIFF
--- a/app/components/avo/panel_component.html.erb
+++ b/app/components/avo/panel_component.html.erb
@@ -20,9 +20,9 @@
     <% end %>
   <% end %>
   <% if body? %>
-    <div data-target="panel-body" class="flex flex-col sm:flex-row space-y-4 sm:space-y-0 sm:gap-4 w-full <% if sidebar? %> has-sidebar <% end %>">
-      <div class="relative flex-1 w-full <% if sidebar? %> sm:w-2/3 <% end %>">
-        <% # The body is wrapped inside another div in order to avoid long & tall panels next to sidebars when the sidebar taller. %>
+    <div data-target="panel-body" class="<%= class_names("flex flex-col sm:flex-row space-y-4 sm:space-y-0 sm:gap-4 w-full", "has-sidebar": sidebar?) %>">
+      <div class="<%= class_names("relative flex-1 w-full", "sm:w-2/3": sidebar?) %>">
+        <%# The body is wrapped inside another div in order to avoid long & tall panels next to sidebars when the sidebar taller. %>
         <div class="relative <%= white_panel_classes %> <%= @body_classes %>">
           <%= body %>
         </div>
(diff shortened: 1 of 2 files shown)


app/components/avo/panel_component.html.erb
```diff
@@ -20,9 +20,9 @@
     <% end %>
   <% end %>
   <% if body? %>
-    <div data-target="panel-body" class="flex flex-col sm:flex-row space-y-4 sm:space-y-0 sm:gap-4 w-full <% if sidebar? %> has-sidebar <% end %>">
-      <div class="relative flex-1 w-full <% if sidebar? %> sm:w-2/3 <% end %>">
-        <% # The body is wrapped inside another div in order to avoid long & tall panels next to sidebars when the sidebar taller. %>
+    <div data-target="panel-body" class="<%= class_names("flex flex-col sm:flex-row space-y-4 sm:space-y-0 sm:gap-4 w-full", "has-sidebar": sidebar?) %>">
+      <div class="<%= class_names("relative flex-1 w-full", "sm:w-2/3": sidebar?) %>">
+        <%# The body is wrapped inside another div in order to avoid long & tall panels next to sidebars when the sidebar taller. %>
```
adrienpoly, line 25: here we had a prasing error because of the extra space `<% #` rather than  `<%`

while on this partial I updated the `class` attribute to use `class_names` helper rather than an `<% if %> .... <% end %>` clause inside a string

app/components/avo/turbo_frame_wrapper_component.html.erb
```diff
@@ -1,10 +1,13 @@
...
+    <% if helpers.turbo_frame_request? %>
+      <%= helpers.turbo_stream_action_tag :append, target: "alerts", template: render(Avo::FlashAlertsComponent.new(flashes: helpers.flash)) %>
+    <% end %>
+    <%= content %>
+  </turbo-frame>
+<% else %>
+  <% if helpers.turbo_frame_request? %>
+    <%= helpers.turbo_stream_action_tag :append, target: "alerts", template: render(Avo::FlashAlertsComponent.new(flashes: helpers.flash)) %>
+  <% end %>
+  <%= content %>
```
adrienpoly, line 12: It is more verbose but it seems the first syntax is not currently supported by Herb.

Paul-Bob: We can go with this duplication for now, I think the previous implementation was a bit hacky

Paul-Bob (approved):
LGTM!

Thanks for looking into this and for the contribution @adrienpoly!

PR #4071, opened 2025-09-17
docs: Update local setup docs

divagueame:
# Description
This PR updates the instructions to run the dummy app locally. It was based on my own experience setting up this project on my machine.

This PR also removes the attribute version from `docker-compose.yml`  to avoid this warning when running the docker compose services:

```
==========================
Creating the Docker volume
==========================

avo-db-data

=====================================
Creating and running the Docker image
=====================================

WARN[0000] /Users/divagueame/avo/docker-compose.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion
[+] Running 1/1
 ✔ Container avo-db-1  Running                                                                                                                                                         0.0s

======================
Preparing the database
======================
```

Fixes # (issue)

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/CONTRIBUTING.MD
+++ b/CONTRIBUTING.MD
@@ -50,11 +50,7 @@ Once you pull the code down to your machine, modify `spec/dummy/config/database.
 
 ## Running the dummy app
 
-1. Make sure you have [overmind](https://github.com/DarthSim/overmind#installation) or [foreman](https://github.com/ddollar/foreman#installation) installed on your system.
-2. Run ```bin/init``` to install ruby and js dependencies, and to setup the database.
-3. If you chose to use Docker Compose for the previous step, start the services: ```docker compose up```
-4. Run `bin/dev` from the root directory, which will start an overmind (similar to foreman) process for the rails server, jsbundling and cssbundling.
-5. Navigate to `localhost:3030` and enjoy the app.
+You can run `bin/dev` from the root directory, which will start an overmind (similar to foreman) process for the rails server, jsbundling and cssbundling. Then, navigate to `localhost:3030` and enjoy the app.
 
 ## Seeding the database for local development
 


CONTRIBUTING.MD
```diff
@@ -50,7 +50,11 @@ Once you pull the code down to your machine, modify `spec/dummy/config/database.
 
 ## Running the dummy app
 
-You can run `bin/dev` from the root directory, which will start an overmind (similar to foreman) process for the rails server, jsbundling and cssbundling. Then, navigate to `localhost:3030` and enjoy the app.
+1. Make sure you have [overmind](https://github.com/DarthSim/overmind#installation) or [foreman](https://github.com/ddollar/foreman#installation) installed on your system.
+2. Run ```bin/init``` to install ruby and js dependencies, and to setup the database.
```
Paul-Bob: This step (2) is mentioned just above on the "## Getting your local environment set up" section

CONTRIBUTING.MD
```diff
@@ -50,7 +50,11 @@ Once you pull the code down to your machine, modify `spec/dummy/config/database.
...
+3. If you chose to use Docker Compose for the previous step, start the services: ```docker compose up```
```
Paul-Bob: I'm using local Postgres instance, is this `docker compose up` cmd necessary every time you want to run the dummy app or just once after booting your machine?

Paul-Bob (commented):
Thanks for this contribution @divagueame!

I have some questions

CONTRIBUTING.MD
divagueame: Correct. As a newcomer, I would prefer to have all necessary steps together to get the app running, instead of jumping around and guessing what applies to my case. Maybe would it make sense to update that section instead?

CONTRIBUTING.MD
divagueame: As long as it's not being shut down, it's not necessary each time. However, running it won't hurt.
I see on 'bin/init':

```
  if use_docker == 'y'
    header 'Stopping the Docker image'
    run! 'docker-compose stop'
  end
```

I guess that's there for a reason, but that's also stopping 'bin/dev' to work when choosing 'use_docker' on 'bin/init' without bringing the services up. As a user, I guess it'd be ideal to just do 'bin/init' and 'bin/dev' and have dependences installed and started but that wasn't my experience with the current setup.

CONTRIBUTING.MD
Paul-Bob: I understand, I didn’t write the CONTRIBUTING.md, but I get the idea of having one section for “Getting your local environment set up” (which you only need to run once during setup), and another section for “Running the dummy app” (which are the exact same steps you run every time you want to start the app after setup).

I see where your confusion comes from, and thanks for looking into this. With the current PR, we have `bin/init` listed redundantly in both the setup and running sections.

I think there’s value in merging those, but we should clearly mark which steps are “run only once when setting up” and which are “run every time you start the application.”

We just need to refine it a bit so it’s clearer and make sure the steps aren’t redundant or duplicated.

Paul-Bob:
@divagueame I reverted the changes to CONTRIBUTING.md and started working on some improvements around it here: https://github.com/avo-hq/avo/pull/4114

Thanks for this contribution! I’ll merge the obsolete version removal.

PR #4100, opened 2025-10-16
fix: ensure consistent cache store reuse in avo

Paul-Bob:
# Description

Fixes an internal cache issue.

`ActiveSupport::Cache.lookup_store` always returns a new cache store instance. Since the block returned by `computed_cache_store` is being executed every time the Avo cache store is requested, a new instance was created each time — effectively not using any caching unless the developer had explicitly configured one.

This commit ensures Avo consistently reuses a single cache instance. If no cache is configured, Avo will now default to:
- `:memory_store` in development
- `:file_store` in production

### DIFF
--- a/lib/avo/configuration.rb
+++ b/lib/avo/configuration.rb
@@ -235,17 +235,20 @@ module Avo
     # If it's one of rejected cache stores, we'll use the FileStore.
     # We decided against the MemoryStore in production because it will not be shared between multiple processes (when using Puma).
     def computed_cache_store
+      memory_store_instance = ActiveSupport::Cache.lookup_store(:memory_store)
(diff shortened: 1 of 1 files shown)


lib/avo/configuration.rb
```diff
@@ -235,17 +235,20 @@ def cache_store
     # If it's one of rejected cache stores, we'll use the FileStore.
     # We decided against the MemoryStore in production because it will not be shared between multiple processes (when using Puma).
     def computed_cache_store
+      memory_store_instance = ActiveSupport::Cache.lookup_store(:memory_store)
```
adrianthedev, line 238: I wonder when is this being computed.
Would there be any side-effects if `computed_cache_store` is ran too early in the boot process?
And it tries to set the `memory_store_instance` and `file_store_instance` and has a problem.

Just asking to ensure we are safe.

PR #4144, opened 2025-12-02
add badge as a ui component

Mihai-Munteanu:
# Description

Fixes # (issue)

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

---

> [!NOTE]
> Introduces a reusable UI badge and migrates existing usages to it.
> 
> - Adds `Avo::UI::BadgeComponent` with `style` (subtle|solid), `color` (semantic and palette), optional `icon`, and new CSS in `css/components/ui/badge.css`; imports in `application.css`
> - Replaces legacy `Avo::Fields::Common::BadgeViewerComponent` in badge field index/show with `BadgeComponent`; removes old component
> - Enhances `Avo::Fields::BadgeField` to support explicit `color`, `style`, `icon` (values or procs) and automatic value-to-color mapping via `options`
> - Updates grid card badge API to a `badge` hash (label/color/style/title/icon) and renders via `BadgeComponent`
> - Adds comprehensive specs and Lookbook previews for the new component and updates feature tests accordingly
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8a1a23cf404ab49f5898455c97d48e03431a4635. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/avo/fields/badge_field.rb
+++ b/lib/avo/fields/badge_field.rb
@@ -37,9 +37,10 @@ module Avo
       def badge_color_for_value
         return "neutral" if value.blank?
 
-        @options.find { |_, configured_values|
-          Array(configured_values).map { |v| v.to_s }.include?(value.to_s)
-        }&.first&.to_s || "neutral"
+        values = @options.find do |_, configured_values|
+          Array.wrap(configured_values).map { |v| v.to_s }.include?(value.to_s)
+        end
+        values&.first&.to_s || "neutral"
       end
     end
   end


app/assets/stylesheets/css/components/ui/badge.css
```diff
@@ -0,0 +1,149 @@
...
+/* Icon-only modifier (circular badges) */
+.badge--icon-only {
+  @apply w-6 h-6 p-0 rounded-full flex items-center justify-center;
+  min-width: unset !important;
+}
+
+/* Secondary - Subtle (default) */
+.badge--subtle.badge--secondary {
+  --badge-bg: var(--color-surface-secondary, var(--color-avo-neutral-50, #f6f6f6));
+  --badge-text: var(--color-text-primary, var(--color-avo-neutral-950, #171717));
```
adrianthedev: Let's use the same variable as text if the icon color is the same always...

spec/features/avo/badge_spec.rb
adrianthedev, line 23: Maybe we can refactor this test to not redeclare the `project` every time.
Maybe even without the `!`.

app/components/avo/fields/badge_field/show_component.html.erb
```diff
@@ -1,3 +1,10 @@
 <%= field_wrapper(**field_wrapper_args) do %>
-  <%= render Avo::Fields::Common::BadgeViewerComponent.new value: @field.value, options: @field.options %>
+  <%= render Avo::UI::BadgeComponent.new(
+    label: @field.value,
+    color: @field.color,
+    style: @field.style,
+    icon: @field.icon,
+    icon_only: @field.icon_only,
+    classes: 'min-w-28'
```
adrianthedev: On the show page we don't need to add a min-width.
We do that on index because there are others on top and below with different widths and it would look bad. Here, usually, we don't have that problem.
```suggestion
```

app/components/avo/fields/badge_field/show_component.html.erb
```diff
@@ -1,3 +1,10 @@
 <%= field_wrapper(**field_wrapper_args) do %>
-  <%= render Avo::Fields::Common::BadgeViewerComponent.new value: @field.value, options: @field.options %>
+  <%= render Avo::UI::BadgeComponent.new(
+    label: @field.value,
+    color: @field.color,
+    style: @field.style,
+    icon: @field.icon,
+    icon_only: @field.icon_only,
```
adrianthedev: This option (`icon_only`) will not exist anymore.
Let's nuke it from everywhere.

app/components/avo/fields/badge_field/index_component.html.erb
```diff
@@ -1,3 +1,10 @@
 <%= index_field_wrapper(**field_wrapper_args, flush: true) do %>
-  <%= render Avo::Fields::Common::BadgeViewerComponent.new value: @field.value, options: @field.options %>
+  <%= render Avo::UI::BadgeComponent.new(
+    label: @field.value,
+    color: @field.color,
+    style: @field.style,
+    icon: @field.icon,
+    icon_only: @field.icon_only,
+    classes: 'min-w-28'
```
adrianthedev: Let's make it smaller. I think it looks a bit cleaner.

```suggestion
    classes: 'min-w-24'
```

CleanShot 2026-01-05 at 16 57 43@2x

app/components/avo/u_i/badge_component.html.erb
```diff
@@ -0,0 +1,12 @@
+<%= content_tag :span, class: helpers.class_names("badge", {
+      :"badge--#{@style}" => @style.present?,
```
adrianthedev: Not sure if this doesn't break tailwind with [dynamic class names](https://tailwindcss.com/docs/detecting-classes-in-source-files#dynamic-class-names).
Also it looks fugly and weird with those `=>`.

Maybe it would be better to compute those two particular classes in the `.rb` file and then just add it here?

Regarding the colors, you can still use interpolation but use [inlining](https://tailwindcss.com/docs/detecting-classes-in-source-files#safelisting-variants).
Something like `@source inline("badge--{secondary,success,informative,warning,error,orange,yellow,green,teal,blue,purple}");` in the `badge.css` file.

app/components/avo/u_i/badge_component.html.erb
```diff
@@ -0,0 +1,12 @@
...
+      :"badge--#{@color}" => @color.present?,
+      :"badge--icon-only" => icon_only?,
+      @classes => @classes.present?,
```
adrianthedev: You don't need to check if it's present. That can be `nil` and `class_names` won't mind.

```suggestion
      @classes
```

app/components/avo/u_i/badge_component.html.erb
```diff
@@ -0,0 +1,12 @@
+<%= content_tag :span, class: helpers.class_names("badge", {
```
adrianthedev: You don't have to use `helpers` here. It inhrits from `Avo::BaseComponent` and it has the required module for this.

app/components/avo/u_i/badge_component.html.erb
```diff
@@ -0,0 +1,12 @@
...
+      :"badge--#{@style}" => @style.present?,
+      :"badge--#{@color}" => @color.present?,
+      :"badge--icon-only" => icon_only?,
+      @classes => @classes.present?,
+}) do %>
+  <%= helpers.svg(@icon, class: "badge__icon") if @icon.present? %>
```
adrianthedev: You don't have to use `helpers` here.

app/components/avo/u_i/badge_component.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+class Avo::UI::BadgeComponent < Avo::BaseComponent
+  unless defined?(VALID_STYLES)
+    VALID_STYLES = %w[solid subtle].freeze
+  end
```
adrianthedev: We usually use it like so in these cases.

```suggestion
  VALID_STYLES = %w[solid subtle].freeze unless defined?(VALID_STYLES)
```

app/components/avo/u_i/badge_component.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  unless defined?(VALID_COLORS)
```
adrianthedev: same here.

app/components/avo/u_i/badge_component.rb
```diff
@@ -0,0 +1,46 @@
...
+    VALID_COLORS = %w[
```
adrianthedev: Using `i` will convert those to symbols and you wouldn't have to cast it `to_s` below.

```suggestion
    VALID_COLORS = %i[
```

app/components/avo/u_i/badge_component.rb
```diff
@@ -0,0 +1,46 @@
...
+    VALID_COLORS.include?(value) ? value : "secondary"
+  end
+
+  prop :style do |value|
+    value = value.to_s
+
+    VALID_STYLES.include?(value) ? value : "subtle"
+  end
+
+  prop :label, default: ""
```
adrianthedev: I think it's fine if we let it default to `nil`.

```suggestion
  prop :label
```

lib/avo/fields/badge_field.rb
```diff
@@ -4,17 +4,87 @@ class BadgeField < BaseField
...
+        else
+          "secondary"
+        end
+      end
+
+      private
+
+      def evaluate_dynamic_value(value)
+        return nil if value.nil?
+        return value unless value.respond_to?(:call)
```
adrianthedev: We don't need this chceck for `nil` and `value.call`.
The execution context [does it automatically](https://github.com/avo-hq/avo/blob/avo-864/feature-add-badges-ui-component/lib/avo/execution_context.rb#L69).

I think you can remove this whole method (`evaluate_dynamic_value`) in favor of `execute_context`.

lib/avo/fields/badge_field.rb
```diff
@@ -4,17 +4,87 @@ class BadgeField < BaseField
...
+        when :warning
+          "warning"
+        when :neutral
+          "secondary"
+        else
+          "secondary"
+        end
+      end
+
+      private
```
adrianthedev: Second time `private` is used.

lib/avo/fields/badge_field.rb
```diff
@@ -4,17 +4,87 @@ class BadgeField < BaseField
...
 
         hide_on [:edit, :new]
 
         default_options = {info: :info, success: :success, danger: :danger, warning: :warning, neutral: :neutral}
         @options = args[:options].present? ? default_options.merge(args[:options]) : default_options
+
+        @color_raw = args[:color]
+        @style_raw = args[:style]
+        @icon_raw = args[:icon]
+        @icon_only_raw = args[:icon_only]
```
adrianthedev: Can we drop the `_raw` suffix?

lib/avo/fields/badge_field.rb
```diff
@@ -4,17 +4,87 @@ class BadgeField < BaseField
...
+
+      def icon
+        evaluate_dynamic_value(@icon_raw)
+      end
+
+      def icon_only
+        evaluate_dynamic_value(@icon_only_raw) || false
+      end
+
+      def badge_color_for_value(value = nil)
```
adrianthedev: This never takes `value`.

lib/avo/fields/badge_field.rb
```diff
@@ -4,17 +4,87 @@ class BadgeField < BaseField
...
+        when :info
+          "informative"
+        when :success
+          "success"
+        when :danger
+          "error"
+        when :warning
+          "warning"
+        when :neutral
+          "secondary"
```
adrianthedev: Wondering why we don't use `neutral` here as well?
I like that name much better than `secondary`.

lib/avo/fields/badge_field.rb
```diff
@@ -4,17 +4,87 @@ class BadgeField < BaseField
...
+          return type.to_sym if values_array.include?(value_str)
+        end
+
+        nil
+      end
+
+      def map_color_type_to_badge_color(color_type)
+        case color_type
+        when :info
+          "informative"
```
adrianthedev: let's use `info` here as well.
I'll talk to Alex to change it in Figma too.

lib/avo/fields/badge_field.rb
```diff
@@ -4,17 +4,87 @@ class BadgeField < BaseField
...
+        evaluate_dynamic_value(@icon_raw)
+      end
+
+      def icon_only
+        evaluate_dynamic_value(@icon_only_raw) || false
+      end
+
+      def badge_color_for_value(value = nil)
+        value ||= self.value
+        return "secondary" if value.blank?
```
adrianthedev:
```suggestion
        return "neutral" if value.blank?
```

lib/avo/fields/badge_field.rb
```diff
@@ -3,18 +3,70 @@ module Fields
     class BadgeField < BaseField
       attr_reader :options
 
+      # Identity mappings: field values like "success" automatically use the success color
+      DEFAULT_OPTIONS = {info: :info, success: :success, danger: :danger, warning: :warning, neutral: :neutral}.freeze unless defined?(DEFAULT_OPTIONS)
```
adrianthedev: Why do we need this?

adrianthedev: Nobody will have this.

lib/avo/fields/badge_field.rb
```diff
@@ -3,18 +3,70 @@ module Fields
...
+        @options.each do |color_type, configured_values|
+          return color_type.to_s if normalize_list(configured_values).include?(normalized_value)
+        end
+
+        nil
+      end
+
+      # Normalizes values for case-insensitive comparison
+      # :Done -> ["done"], [:Foo, "Bar"] -> ["foo", "bar"]
+      def normalize_list(values)
```
adrianthedev: we don't need to normalize. let's let the user explicitly say what they want. a string or symbol.

lib/avo/fields/badge_field.rb
```diff
@@ -4,17 +4,43 @@ class BadgeField < BaseField
...
+      end
+
+      # Maps field value to a color based on @options configuration
+      # Example: "Done" -> "success" if options = { success: [:done, :complete] }
+      def badge_color_for_value
+        return "neutral" if value.blank?
+
+        @options.find { |_, configured_values|
+          Array(configured_values).map { |v| v.to_s }.include?(value.to_s)
+        }&.first&.to_s || "neutral"
```
adrianthedev:
```suggestion
        values = @options.find do |_, configured_values|
          Array.wrap(configured_values).map { |v| v.to_s }.include?(value.to_s)
        end
        values&.first&.to_s || "neutral"
```

There's a difference in `Array()` and `Array.wrap`.

Let's use `do end` when the content is more complex.

adrianthedev (approved):

Issue #4146, opened 2025-12-03
Field has_many with an STI child


bruno-costanzo:
When using STI (Single Table Inheritance) combined with a polymorphic `has_many` association defined in the parent class, accessing the show page of an STI subclass resource raises:

   ```
   undefined method 'new' for nil
   ```

   The issue occurs in Avo's reflection handling:

   - `@reflection.active_record.name` returns the **parent STI class** (`"User"`) where the association is defined
   - Avo tries to find a resource for `User`, but only `Admin` and `Viewer` resources exist
   - `get_resource_by_model_class("User")` returns `nil`
   - Calling `.new` on `nil` raises the error

   ### Reproduction repository for the bug

   https://github.com/bruno-costanzo/avo-sti-reproduction-repo

   ### Steps to use in the reproduction repository

   1. Clone the repo and run `bin/setup`
   2. Navigate to `http://localhost:3000/avo`
   3. Click on "Admins" in the sidebar
   4. Click on any Admin record to view its show page
   5. See error: `undefined method 'new' for nil`

   ### Expected behavior & Actual behavior

   **Expected:** Avo should resolve the resource using the actual model class (`Admin`) rather than the class where the association is defined (`User`).

   **Actual:** Avo looks up the resource using the parent STI class name from `@reflection.active_record.name`, which returns `nil` when no resource exists for the parent class.

   ### System configuration

   **Avo version**: >= 3.17

   **Rails version**: 8.1.1

   **Ruby version**: 3.4.5

   **License type**:

    - [x] Community
    - [ ] Pro
    - [ ] Advanced

   **Are you using Avo monkey patches, overriding views or view components?**

    - [ ] Yes
    - [x] No"

PR #4147, opened 2025-12-03
refactor: clean up obsolete code and fix STI resource lookup

bruno-costanzo:
# Description

  When using STI with a polymorphic `has_many` association defined in the parent class, `@reflection.active_record.name` returns the parent class name instead of the actual record's class. This causes `get_resource_by_model_class` to return `nil` when no resource exists for the parent class.
  
  The solution is removing the method `has_reflection_and_is_read_only` and all the calls to it, since it always returns false.

  Fixes [4146](https://github.com/avo-hq/avo/issues/4146)

  # Checklist:

  ## Manual review steps

  1. Clone https://github.com/bruno-costanzo/avo-sti-reproduction-repo
  2. Run `bin/setup`
  3. Navigate to `http://localhost:3000/avo`
  4. Click on "Admins" → Click on any Admin record
  5. Before fix: `undefined method 'new' for nil` / After fix: page loads correctly

### DIFF
--- a/app/components/avo/resource_component.rb
+++ b/app/components/avo/resource_component.rb
@@ -59,7 +59,7 @@ class Avo::ResourceComponent < Avo::BaseComponent
   def can_see_the_actions_button?
     return authorize_association_for(:act_on) if @reflection.present?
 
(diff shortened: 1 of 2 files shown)


adrianthedev:
Thank you for your time, @bruno-costanzo!

PR #4149, opened 2025-12-05
No needed anymore UI4: pagination component

Mihai-Munteanu:
# Description

Fixes # (issue)

# Checklist:

## Manual review steps

1. Step 1
1. Step 2

### DIFF
--- a/spec/dummy/app/controllers/pagination_test_controller.rb
+++ b/spec/dummy/app/controllers/pagination_test_controller.rb
@@ -1,7 +1,0 @@
-class PaginationTestController < ApplicationController
-  include Pagy::Backend
-
-  def index
-    @pagy, @users = pagy(User.order(:id), limit: params[:per_page] || 24)
-  end
-end


spec/dummy/test/components/previews/u_i/pagination_component_preview.rb
```diff
@@ -0,0 +1,94 @@
+# Pagination Component Preview
+#
+# This preview demonstrates the Avo::UI::PaginationComponent
+#
+# To view in Lookbook:
+# 1. Start the development server: `bin/dev`
+# 2. Visit: http://localhost:3030/avo/lookbook
```
Paul-Bob, line 7:
```suggestion
# 2. Visit: http://localhost:3030/admin/lookbook
```

app/components/avo/u_i/pagination_component.rb
```diff
@@ -0,0 +1,57 @@
...
+  def can_go_previous?
+    @current_page > 1
+  end
+
+  def can_go_next?
+    @current_page < total_pages
+  end
+
+  def per_page_display_text
+    content_tag(:span, @items_per_page, class: "font-bold") + " per page"
```
Paul-Bob, line 54: Let's apply localization on all UI strings

```suggestion
    content_tag(:span, @items_per_page, class: "font-bold") + t('avo.per_page').downcase
```

Paul-Bob (changes requested):
Checked the component on Lookbook and it looks good!

There are a few things I noticed:

1. We should not use hardcoded strings in the UI. We should use localization for them. I left an example on the "per page" string.
2. Lets fix the linter errors before requesting review.

---

I think we should put this PR on hold until we get detailed states for the pagination on Figma.

adrianthedev:
We're waiting to finish up the button component first to use in pagination.

PR #4154, opened 2025-12-09
(AVO-877) tabs component implementation

Mihai-Munteanu:
# Description

Replace old tab component with a new tab component and add JavaScript style updates

- Replace old component with new tab component in tab_group_component
- Add tab_button_data_attributes helper method for Stimulus data attributes
- Add tabButton target and updateActiveTabButtons method to tabs controller
- Add toggleTabClasses helper for dynamic class toggling
- Enable instant visual feedback when switching tabs via JavaScript"

# Checklist:

## Screenshots & recording
Screenshot 2026-01-27 at 17 27 53

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -111,6 +111,7 @@
   @import "./css/components/ui/description_list.css";
   @import "./css/components/ui/tabs.css";
   @import "./css/components/ui/badge.css";
+  @import "./css/components/ui/dropdown.css";
   @import "./css/components/ui/checkbox.css";
   @import "./css/components/ui/radio_button.css";
 
--- a/app/assets/stylesheets/css/components/ui/dropdown.css
+++ b/app/assets/stylesheets/css/components/ui/dropdown.css
@@ -1,0 +1,56 @@
+.dropdown {
+  @apply relative inline-flex;
+}
+
+
+/* Closed state - hidden by default */
+.dropdown-popover {
+  @apply absolute z-10 rounded-lg bg-primary top-full end-0 bottom-auto start-auto;
+  box-shadow: var(--box-shadow-card);
+}
+
+/* Backdrop - transparent for dropdown (no overlay) */
+.dropdown-popover::backdrop {
+  background-color: transparent;
+}
+
+/* Backdrop remains transparent when open */
+.dropdown-popover:open::backdrop {
+  background-color: transparent;
+}
+
+.dropdown-menu {
+  @apply flex flex-col w-full rounded-lg border border-tertiary bg-primary overflow-hidden;
+}
+
+/* ==========================================================================
+   List & Items
+   ========================================================================== */
+
+.dropdown-menu__group {
+  @apply flex flex-col items-start self-stretch py-1;
+}
+.dropdown-menu__list {
+  @apply flex flex-col items-start self-stretch px-1;
+}
+
+.dropdown-menu__item,
+.dropdown-menu__list > :is(a, button) {
+  @apply flex items-center w-full justify-start focus:outline-none gap-2 px-2 py-1.5 min-h-8 rounded-md border-none bg-transparent overflow-hidden text-ellipsis whitespace-nowrap text-content font-medium text-sm cursor-pointer leading-5;
+}
+
+.dropdown-menu__item:hover,
+.dropdown-menu__list > :is(a, button):hover {
+  @apply bg-secondary;
+}
+
+.dropdown-menu__item:focus-visible,
+.dropdown-menu__list > :is(a, button):focus-visible {
+  @apply ring-2 ring-inset ring-content bg-primary outline-none;
+}
+
+.dropdown-menu__icon,
+.dropdown-menu__list > :is(a, button) svg {
+  @apply shrink-0 size-4 text-content;
+}
+
--- a/app/assets/stylesheets/css/css-animations.css
+++ b/app/assets/stylesheets/css/css-animations.css
@@ -1,5 +1,6 @@
 /* Actions Dropdown Animations */
 
+.css-animate-dialog-slide-down[open],
 .css-animate-slide-down {
   opacity: 1;
   transform: translateY(0);
@@ -17,6 +18,7 @@
 }
 
 /* Hidden state - CSS handles the leave animation */
+.css-animate-dialog-slide-down,
 .css-animate-slide-down[hidden] {
   opacity: 0;
   transform: translateY(-0.25rem);
--- a/app/assets/stylesheets/css/variables.css
+++ b/app/assets/stylesheets/css/variables.css
@@ -25,6 +25,7 @@
 
   --shadow-content: 0 8px 16px 0 rgba(0, 0, 0, 0.08);
 
+  --box-shadow-card: 0px 8px 16px 0px rgba(0, 0, 0, 0.04);
   --box-shadow: 0 1px 2px 0 rgba(0, 0, 0, 0.04);
 
   /* foundations - Light Mode */
--- a/app/components/avo/u_i/dropdown_component.html.erb
+++ b/app/components/avo/u_i/dropdown_component.html.erb
@@ -1,0 +1,11 @@
+<%= content_tag :div, class: class_names("dropdown"), data: {controller: "dropdown-menu"} do %>
+  <%= trigger %>
+
+  <%= render Avo::UI::DropdownMenuComponent.new(
+    data: {dropdown_menu_target: "menu"}
+  ) do %>
+    <% if items.present? %>
+      <%= items %>
+    <% end %>
+  <% end %>
+<% end %>
--- a/app/components/avo/u_i/dropdown_component.rb
+++ b/app/components/avo/u_i/dropdown_component.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class Avo::UI::DropdownComponent < Avo::BaseComponent
+  renders_one :trigger
+  renders_one :items
+
+  # this is used to trigger the dropdown menu from trigger element
+  # data: {action: component.action} => click->dropdown-menu#toggle
+  def action
+    @action ||= "click->dropdown-menu#toggle"
+  end
+end
--- a/app/components/avo/u_i/dropdown_menu_component.html.erb
+++ b/app/components/avo/u_i/dropdown_menu_component.html.erb
@@ -1,0 +1,13 @@
+<%= content_tag :dialog, class: class_names(
+  "dropdown-popover",
+  "css-animate-dialog-slide-down",
+  @classes
+), closedby: "any", data: @data do %>
+  <div class="dropdown-menu">
+    <div class="dropdown-menu__group">
+      <div class="dropdown-menu__list">
+        <%= content %>
+      </div>
+    </div>
+  </div>
+<% end %>
--- a/app/components/avo/u_i/dropdown_menu_component.rb
+++ b/app/components/avo/u_i/dropdown_menu_component.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+
+class Avo::UI::DropdownMenuComponent < Avo::BaseComponent
+  prop :data, default: {}.freeze
+  prop :classes
+end
--- a/app/components/avo/u_i/icon_button_component.html.erb
+++ b/app/components/avo/u_i/icon_button_component.html.erb
@@ -1,0 +1,7 @@
+<%= tag.button(
+  type: @type,
+  class: button_classes,
+  data: data
+) do %>
+  <%= svg(@icon) %>
+<% end %>
--- a/app/components/avo/u_i/icon_button_component.rb
+++ b/app/components/avo/u_i/icon_button_component.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+class Avo::UI::IconButtonComponent < Avo::BaseComponent
+  prop :icon
+  prop :action
+  prop :classes
+  prop :type, default: :button
+  prop :data, default: {}.freeze
+
+  def data
+    {
+      action: @action
+    }.merge(@data)
+  end
+
+  def button_classes
+    class_names(
+      "inline-flex items-center justify-center rounded-md p-1 cursor-pointer",
+      @classes
+    )
+  end
+end
--- a/app/javascript/js/controllers.js
+++ b/app/javascript/js/controllers.js
@@ -54,6 +54,7 @@ import TextFilterController from './controllers/text_filter_controller'
 import TippyController from './controllers/tippy_controller'
 import TiptapFieldController from './controllers/fields/tiptap_field_controller'
 import ToggleController from './controllers/toggle_controller'
+import DropdownController from './controllers/dropdown_menu_controller'
 import TrixBodyController from './controllers/trix_body_controller'
 import TrixFieldController from './controllers/fields/trix_field_controller'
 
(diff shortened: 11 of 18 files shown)


app/javascript/js/controllers/tabs_controller.js
```diff
@@ -1,73 +1,219 @@
-import { Controller } from '@hotwired/stimulus'
+import { Controller } from "@hotwired/stimulus";
 
 export default class extends Controller {
-  static targets = ['tabPanel']
+  static targets = ["tabPanel", "tabButton"];
```
adrianthedev: You have an extension which overrides the formatting we use.
Check if it's prettier or something else (es-lint, ts-lint).
You should be able to disable it per workspace if you need it in other projects.

app/javascript/js/controllers/tabs_controller.js
```diff
@@ -16,7 +16,11 @@ export default class extends Controller {
 
   selectCurrentTab() {
     const params = {}
-    Array.from(new URL(window.location).searchParams.entries()).forEach(([key, value]) => { params[key] = value })
+    Array.from(new URL(window.location).searchParams.entries()).forEach(
```
adrianthedev: You have an extension which overrides the formatting we use.
Check if it's prettier or something else (es-lint, ts-lint).
You should be able to disable it per workspace if you need it in other projects.

I'm going to let you revert the rest of the non-functional changes back.

app/components/avo/u_i/tabs_component.html.erb
```diff
@@ -0,0 +1,8 @@
+<div
+  class="<%= classes %>"
+  role="tablist"
+  id="<%= tablist_id %>"
+  aria-label="<%= tablist_aria_label %>"
+>
+  <%= safe_join(items.map { |item| render(item) }) %>
```
adrianthedev: I don't think we need the `safe_join` here. We'll just output `<%= content %>`.
```suggestion
  <%= content %>
```

app/components/avo/u_i/tabs_component.html.erb
```diff
@@ -0,0 +1,8 @@
+<div
+  class="<%= classes %>"
+  role="tablist"
+  id="<%= tablist_id %>"
+  aria-label="<%= tablist_aria_label %>"
+>
```
adrianthedev: When we have so much `erb` in `html` we'll jsut switch it up and write `erb`.

```erb
<%= content_tag :div,
  class: classes,
  role: :tablist,
  id: id,
  aria_label: aria_label  do %>
```

app/components/avo/u_i/tabs_component.rb
```diff
@@ -0,0 +1,36 @@
...
+      prop :variant, default: :group # :group (no border) or :scope (with border)
+      prop :aria_label
+      prop :id
+
+      # Ensure TabItems inherit parent variant unless explicitly overridden
+      alias_method :with_item_without_variant, :with_item
+      def with_item(**attrs, &block)
+        attrs[:variant] ||= @variant
+        with_item_without_variant(**attrs, &block)
+      end
```
adrianthedev: I don't think this is needed.

app/components/avo/u_i/tabs_component.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module Avo
+  module UI
+    class TabsComponent < Avo::BaseComponent
+      VARIANTS = %i[group scope].freeze
+
+      renders_many :items, Avo::UI::TabComponent
```
adrianthedev: We won't need this if we use the `content` approach.

app/assets/stylesheets/css/components/ui/tabs.css
```diff
@@ -0,0 +1,126 @@
...
+    border-bottom-width: 1px;
+    border-bottom-color: var(--color-border-primary, #e7e7e7);
+  }
+
+  /* Scope variant: no background (scope variant never has background) */
+  /* This applies to tabs inside a .tabs--scope container */
+  .tabs--scope .tabs__item,
+  .tabs--scope .tabs__item--active,
+  .tabs__item-wrapper--scope .tabs__item,
+  .tabs__item-wrapper--scope .tabs__item--active {
```
adrianthedev: Let's use native CSS nesting where we can.

```css
.tabs--scope {
  &.tabs__item{
    ...
  }
}
```

adrianthedev: When we see a elements (`.tabs__item`) needing modifiers (`.tabs__item--active`), let's consider it a smell and extract that into a separate `.tab` component of itself.
The fact that we have separate component files (view components) for them, should be an indicator.

adrianthedev: We might even add these instructions to the `agents.md` file.

spec/dummy/db/schema.rb
```diff
@@ -174,9 +174,9 @@
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.integer "price_cents", default: 0, null: false
-    t.string "price_currency", default: "'USD'::character varying", null: false
-    t.integer "rating", default: 0, null: false
+    t.string "price_currency", default: "USD", null: false
```
adrianthedev: Please revert this change and re run `rails db:fresh` or `rails db:setup:fresh` on your end to reimplement the right schema.

It might have been commited by mistake from my end.

app/javascript/js/controllers/tabs_controller.js
```diff
@@ -70,4 +79,141 @@ export default class extends Controller {
...
+
+  focusTab(tabButton) {
+    // Remove focus class from all tabs first
+    this.removeKeyboardFocusClass()
+
+    // Add keyboard focus class and focus the tab
+    tabButton.classList.add('tabs__item--focused')
+    tabButton.setAttribute('tabindex', '0')
+    tabButton.focus()
+  }
```
adrianthedev: I think this method does too much.
I mean, when you hit tab it already cycles through  those tabs automatically focusing the right one.
I feel it tries to reinvent what the browser already does, right?

app/javascript/js/controllers/tabs_controller.js
```diff
@@ -70,4 +79,141 @@ export default class extends Controller {
...
+    // Remove keyboard focus class from all tabs
+    this.tabButtonTargets.forEach((button) => {
+      button.classList.remove('tabs__item--focused')
+    })
+  }
+
+  // Handle blur event - remove focus class when tab loses focus
+  handleBlur(e) {
+    e.currentTarget.classList.remove('tabs__item--focused')
+  }
```
adrianthedev: Same here.
Instead of having a class (or just a class) `--focused`, we can target in CSS the `:focused` state.
Isn't that what we actually need?

app/javascript/js/controllers/tabs_controller.js
```diff
@@ -70,4 +79,141 @@ export default class extends Controller {
...
+
+  // Handle mousedown - remove focus class immediately on mouse interaction
+  handleMouseDown(e) {
+    // Remove focus class when mouse is used (prevents focus ring on click)
+    this.removeKeyboardFocusClass()
+  }
```
adrianthedev: Wondering the same about this one... Do we actually need this? Genuinelly asking.

app/javascript/js/controllers/tabs_controller.js
```diff
@@ -70,4 +79,141 @@ export default class extends Controller {
...
+    tabButton.setAttribute('tabindex', '0')
+    tabButton.focus()
+  }
+
+  removeKeyboardFocusClass() {
+    // Remove keyboard focus class from all tabs
+    this.tabButtonTargets.forEach((button) => {
+      button.classList.remove('tabs__item--focused')
+    })
+  }
```
adrianthedev: We can target the `:focused` state.

adrianthedev:
Talked to @Mihai-Munteanu and agreed to split the tab component into two. one for tabs and one for scopes. they will never be interchangable.

maybe extract tab item if needed.

much of the complexity reduced.

app/components/avo/u_i/tabs/tab_component.rb
```diff
@@ -0,0 +1,107 @@
...
+        prop :href, default: nil
+        prop :id
+        prop :disabled, default: false
+        prop :aria_controls
+        prop :data, default: {}.freeze
+        prop :classes, default: nil
+        prop :title, default: nil
+
+        def before_render
+          @data = @data.deep_dup
```
adrianthedev: This feels like a smell. DO we have a reason on why to run this?

app/components/avo/u_i/tabs/tab_component.rb
```diff
@@ -0,0 +1,107 @@
...
+        end
+
+        def link?
+          @href.present? && @href != "#" && !@disabled
```
adrianthedev: Why add `@disabled` in here? Genuinely curious.

app/components/avo/u_i/tabs/tab_component.html.erb
```diff
@@ -0,0 +1,7 @@
+<div class="<%= wrapper_classes %>">
+  <%= if link?
+        link_to(tab_content, @href, shared_attributes)
+      else
+        content_tag(:button, tab_content, button_attributes)
+      end %>
```
adrianthedev: Again, this `if link do something if not do something else`.
Can you explain why we need it like that?
Do we ever use it not as a link?

app/components/avo/u_i/tabs/tab_component.rb
```diff
@@ -0,0 +1,107 @@
...
+          shared_attributes.merge(
+            type: "button",
+            disabled: @disabled,
+            "aria-disabled": @disabled
+          )
+        end
+
+        def tab_content
+          safe_join([render_icon, render_label].compact)
+        end
```
adrianthedev: We don't use this approach too often.
I prefer to keep presentation items in the presentation.

I see how they are extracted away, but it makes it difficult to read the presentation (HTML) and figure out what's happening.

app/components/avo/u_i/tabs/tab_component.rb
```diff
@@ -0,0 +1,107 @@
...
+          return if @icon.blank?
+
+          svg @icon, class: "tabs__item-icon", "aria-hidden": "true"
+        end
+
+        def wrapper_classes
+          class_names(
+            "tabs__item-wrapper",
+            "tabs__item-wrapper--#{@variant}",
+            (ACTIVE_CLASS if scope? && @active)
```
adrianthedev: Using constants with class_names is nasty nomatter how you do it. I think I like this style a bit more.

```suggestion
            "#{ACTIVE_CLASS}": scope? && @active
```

app/components/avo/u_i/tabs/tab_component.rb
```diff
@@ -0,0 +1,107 @@
...
+          }
+        end
+
+        def element_classes
+          class_names(
+            "tabs__item",
+            ("tabs__item--group" if @variant == :group),
+            (ACTIVE_CLASS if @active),
+            ("tabs__item--disabled" if @disabled),
+            @classes
```
adrianthedev:
```suggestion
            @classes,
            "tabs__item",
            "tabs__item--group": @variant == :group,
            "tabs__item--disabled": @disabled,
            "#{ACTIVE_CLASS}": @active
```

app/components/avo/u_i/tabs/tab_component.rb
```diff
@@ -0,0 +1,107 @@
...
+        def shared_attributes
+          {
+            class: element_classes,
+            data: @data,
+            title: @title,
+            **tab_attributes.except(:"aria-disabled")
+          }
+        end
+
+        def button_attributes
```
adrianthedev: If we remove the `button` scenario we might not need this.

app/components/avo/u_i/tabs/tabs_component.rb
```diff
@@ -0,0 +1,21 @@
...
+        prop :aria_label
+        prop :id
+
+        def tablist_id
+          @id || "tabs-#{object_id}"
+        end
+
+        def tablist_aria_label
+          @aria_label || "Tabs navigation"
+        end
```
adrianthedev: We can simplify this by assiging the default in `prop`.

app/components/avo/u_i/tabs/tabs_component.rb
```diff
@@ -0,0 +1,21 @@
...
+  module UI
+    module Tabs
+      class TabsComponent < Avo::BaseComponent
+        prop :variant, default: :scope
+        prop :aria_label
+        prop :id
+
+        def tablist_id
+          @id || "tabs-#{object_id}"
+        end
```
adrianthedev: Same here about the default.

app/assets/stylesheets/css/components/ui/tabs.css
```diff
@@ -0,0 +1,72 @@
...
+  /* Container */
+  .tabs {
+    @apply flex flex-wrap;
+  }
+
+  /* Base Item */
+  .tabs__item {
+    @apply inline-flex items-center mx-1 px-3 py-1 cursor-pointer text-sm font-medium leading-4 bg-transparent gap-2 rounded-md;
+    color: var(--color-content);
+    font-family: var(--font-sans);
```
adrianthedev: Let's make these tailwind classes where possible, please.

app/components/avo/tab_group_component.rb
```diff
@@ -83,4 +84,37 @@ def args(tab)
...
+    }
+
+    if view.in?(%w[edit update])
+      helpers.edit_resource_path(**base_options, record: resource.record)
+    elsif view.in?(%w[new create])
+      helpers.new_resource_path(**base_options)
+    else
+      helpers.resource_path(**base_options, record: resource.record)
+    end
+  end
```
adrianthedev, line 103: This is usually a smell too.
If in a parent component where we have multiple children, we need to compute something for the child (`tab`).

Can we just move this into the child and not have to pass the `tab` around?

app/components/avo/tab_group_component.rb
```diff
@@ -83,4 +84,37 @@ def args(tab)
...
+
+  def tab_data(tab, current_tab)
```
adrianthedev, line 105: Same here about computing in the parent.

app/components/avo/u_i/tabs/tab_component.rb
```diff
@@ -0,0 +1,107 @@
...
+        ACTIVE_CLASS = "tabs__item--active"
+
+        prop :label
+        prop :variant, default: :scope
+        prop :active, default: false
+        prop :icon
+        prop :href, default: nil
+        prop :id
+        prop :disabled, default: false
+        prop :aria_controls
```
adrianthedev: I see that we're never passing `aria-controls` from the outside. Do we really need it here?

app/components/avo/u_i/tabs/tab_component.rb
```diff
@@ -0,0 +1,107 @@
...
+        prop :data, default: {}.freeze
+        prop :classes, default: nil
+        prop :title, default: nil
```
adrianthedev:
```suggestion
        prop :classes
        prop :title
```

They default to `nil` already.

adrianthedev (changes requested):
Great start.
The components look great. I checked them in the preview.

There are a few things to tweak to move this over the finish line:
- add screenshots to the PR with the finished result
- make CSS properties TW classes wehere possible and makes sense
- Let's see if we actually need the link and non-link version. I don't think we are using it as a non-link anywhere in Avo. if not, we would simplify the implementation quite a bit
- let's move some of the inline helpers `render_icon`, `render_label` to the erb part so we have presentation in presentation.
- left a few comments inline

app/components/avo/tab_group_component.rb
adrianthedev: talked about it. not dumb if we move it to the child.

PR #4172, opened 2025-12-19
feat: Add indeterminate state for boolean fields when nil

### LINKED ISSUE #3873: Boolean field to show indeterminate state

icaroryan:
# Description
Adds `nil_as_indeterminate` for boolean fields so nil values can render a grey minus-circle icon instead of the em dash, while keeping default behavior unchanged.

`field :is_capital, as: :boolean, nil_as_indeterminate: true`

Fixes #3873

# Checklist:

- [X] I have made corresponding changes to the [documentation](https://github.com/avo-hq/avodocs) (https://github.com/avo-hq/docs.avohq.io/pull/466)

## Screenshots & recording

image

## Manual review steps

1. In a resource, set `nil_as_indeterminate: true` on a `boolean` field and verify nil shows a grey minus-circle icon instead of the em dash.
2. Ensure the `true` and `false` values still behaves correctly

### DIFF
--- a/app/components/avo/fields/boolean_field/edit_component.html.erb
+++ b/app/components/avo/fields/boolean_field/edit_component.html.erb
@@ -1,6 +1,7 @@
 <%= field_wrapper(**field_wrapper_args) do %>
   <div class="h-8 flex items-center">
-    <%= @form.label @field.id,
+    <%= content_tag :label,
+      for: "#{@form.object_name}_#{@field.id}",
       class: class_names(
         "relative flex items-center",
         "opacity-50": disabled?,


app/components/avo/fields/common/boolean_check_component.rb
```diff
@@ -1,18 +1,39 @@
...
+      color: "text-red-600"
+    },
+    nil => {
+      name: "indeterminate",
+      icon: "heroicons/outline/minus-circle",
+      color: "text-gray-400"
+    }
+  }.freeze
+
+  prop :checked, default: nil
```
icaroryan, line 22: @Paul-Bob Changed the `default` to `nil`, which at the end of the day shouldn't make a difference because `BooleanCheckComponent` is never called without `checked`

Paul-Bob (approved):
LGTM!

Thank you @icaroryan!

PR #4178, opened 2025-12-22
RTL exploration

Paul-Bob:
# Description

RTL support exploration

https://github.com/user-attachments/assets/08570143-ef81-45be-9fc1-b1ace1247acb


adrianthedev:
Awesome that you've started on this.

It would be interesting to see if we could add a testing step to check if we introduced any "left/right" modifiers instead of "start/end".

Paul-Bob:
Closing in favor of https://github.com/avo-hq/avo/pull/4193

PR #4181, opened 2025-12-29
Add hidden description to boolean check component

Timmitry:
# Description

Boolean fields are displayed by default as svg-icons in index and show views. However, since the SVG has `aria-hidden="true"`, no content is shown at all for screen readers. This makes it also difficult to test boolean fields with e.g. Capybara.

This PR introduces a hidden span-Element, which uses `I18n.t(true)` or `I18n.t(false)` to print the global translation of true/false. This will just be the string "True" or "False" if not configured, but can be overwritten in the translation yml files with e.g. `"true": Yes`.

# Checklist:

## Manual review steps

E.g. in firefox, right click on the svg icon and choose "Show accessibility properties". This will show the accessibility tree, where the table cell should have the new translations (without this change, the table cell content will be empty).

### DIFF
--- a/app/components/avo/fields/common/boolean_check_component.html.erb
+++ b/app/components/avo/fields/common/boolean_check_component.html.erb
@@ -1,2 +1,2 @@
-<%= content_tag :span, @checked.nil? ? t("indeterminate") : t(@checked), class: 'sr-only' %>
+<%= content_tag :span, t(@checked), class: 'sr-only' %>
 <%= helpers.svg "#{@icon}.svg", class: classes, data: {checked_state: state} %>
--- a/spec/components/avo/fields/common/boolean_check_component_spec.rb
+++ b/spec/components/avo/fields/common/boolean_check_component_spec.rb
@@ -2,15 +2,13 @@ require "rails_helper"
 
 RSpec.describe Avo::Fields::Common::BooleanCheckComponent, type: :component do
   it "outputs an accessible text" do
-    I18n.backend.store_translations(:en, true: "Yes", false: "No", "indeterminate": "Indeterminate")
-
     render_inline(described_class.new(checked: true))
-    expect(page.text.strip).to eql("Yes")
+    expect(page.text.strip).to eql("True")
 
     render_inline(described_class.new(checked: false))
-    expect(page.text.strip).to eql("No")
+    expect(page.text.strip).to eql("False")
 
     render_inline(described_class.new(checked: nil))
-    expect(page.text.strip).to eql("Indeterminate")
+    expect(page.text.strip).to eql("")
   end
 end


Timmitry:
I'm not sure if the translations should be overridable from the field-definition. Let me know what you think - I can also extend the documentation a bit if required!

Paul-Bob:
Hi @Timmitry, thanks for this contribution. WCAG 2.2 compliance is one of our goals for Avo 4, and this PR brings us closer to that goal.

Since we are also applying a new design in Avo 4, could you please add a test for this use case? This will help ensure that the accessibility of this component is maintained.

Timmitry:
@Paul-Bob Thank you for looking at my PR! I added a test for the accessible text, let me know what you think 🙂

app/components/avo/fields/common/boolean_check_component.html.erb
```diff
@@ -1 +1,2 @@
+<%= content_tag :span, @checked.nil? ? t("indeterminate") : t(@checked), class: 'sr-only' %>
```
Paul-Bob: @Timmitry I think we should not translate "indeterminate". That is a field option that displays an icon instead of an empty value in the UI, but the underlying value is `nil`. In my opinion, the screen reader should read it as `nil` instead of a localized "indeterminate".

```suggestion
<%= content_tag :span, t(@checked), class: 'sr-only' %>
```

Timmitry: I thought about this, too. With your proposed change, it wouldn't print "nil" though, but instead the whole content would be an empty string "". If that is desired, I can make the change. Otherwise we would have to think about how to display "nil".

Paul-Bob: Got it. I am not sure how it is supposed to behave with a screen reader, but maybe an empty string is fine. If you are checking the column and see some values as true and others as false, you would assume that an empty value represents the absence of a value, which is correct. It does not have to say "nil" or "null"

Timmitry: All right, then I fixed this now and simplified the tests!

Paul-Bob: Thank you @Timmitry for this contribution, and congrats on your first merged Avo PR! 🎉

Timmitry: Thanks for reviewing and merging, @Paul-Bob!

adrianthedev:
@Paul-Bob I wonder how is this working when we don't have a translation for `true` or `false`.

Paul-Bob:
> @Paul-Bob I wonder how is this working when we don't have a translation for `true` or `false`.

I added those translations to Avo's locales in the last minor release.

https://github.com/avo-hq/avo/pull/4236

adrianthedev:
Also to `4-dev`?

Paul-Bob:
> Also to `4-dev`?

Probably yes, I merge `main` into `4-dev` often

PS: If is not there yet only need to merge `main` -> `4-dev`

PS PS: https://github.com/avo-hq/avo/commits/4-dev/ Is there

PR #4197, opened 2026-01-09
Remove readable call from search controller.

### LINKED ISSUE #4179: Search breaks when current view has certain query params
### Describe the bug
Avo Pro global search fails when query params have correctly encoded `#` character.

### Reproduction repository for the bug
https://github.com/k1bs/fragment-encoding-bug

### Steps to use in the reproduction repository
1. Run db:setup to get seeds (don't actually think this is required)
2. click on "Bug Demo" in left nav
3. Note query param with correctly encoded `#` character (%23) in url
4. Click into global search

EXPECTED: I expect to be able to use search regardless of which correctly encoded params I use.

ACTUAL: The search errors with `undefined method 'strip' for nil`

### System configuration
**Avo version**:3.24.1

**Rails version**:8.1.1

**Ruby version**:3.4.5

**License type**:
 - [ ] Community
 - [x] Pro
 - [ ] Advanced

 - [x] No

Image

### Additional context
So, after triaging, we figured out that the issue stems from `search_controller.js`'s `searchUrl` method. I believe specifically that calling `.readable()` is allowing the `%23` character to be decoded into `#`, which then becomes treated as a fragment/anchor (like when you link to a div with an id tag) and it makes everything past that character unreadable as query params, including the q param, which then attempts to be stripped later on.

#### Impact

#### Urgency

k1bs:
# Description
Ensures that global search still works when a properly formatted `#` character is in the query params.

Fixes #4179 

# Checklist:

## Screenshots & recording
Screenshot 2026-01-09 at 11 20 17 AM

## Manual review steps
I applied the change and then updated my bug report branch locally to use my fixed version, and the search correctly loads with the query param in the URL. (worth mentioning that this was quite difficult, as I had to precompile assets and force them into a different version of the fork, as my bug report test app was on 3.24.1. I actually did this on another fork, becuase I assume you guys build your assets through a pipeline or something.)

To do from a new bug report app:

1. Instantiate the avo pro bug report app on the latest version
2. point the gemfile to this fork.
3. Bundle (you'll need to precompile to make this work)
4. Go to any AVO route with a query param with a correctly encoded `#` character (you can just type it in, it's `%23`)
5. Successfully use the global search.

This is NOT possible in main.

### DIFF
--- a/app/javascript/js/controllers/search_controller.js
+++ b/app/javascript/js/controllers/search_controller.js
@@ -242,7 +242,7 @@ export default class extends Controller {
 
     return url.segment([window.Avo.configuration.root_path, ...this.searchSegments()])
       .search(this.searchParams(encodeURIComponent(query)))
-      .readable().toString()
+      .toString()
   }
 
   searchSegments() {


k1bs:
There doesn't currently exist a test for search_controller, and I figure starting one is outside of the scope of this work.

Paul-Bob:
Thanks for this contribution @k1bs, we appreciate it!

> There doesn't currently exist a test for search_controller, and I figure starting one is outside of the scope of this work.

The global search tests are in `avo-pro`. We will handle those.

Issue #4097, opened 2025-10-14
Rails 8 Zeitwerk incompatibility: url_helpers.rb defines UrlHelpers instead of URLHelpers


clayreinhardt:
### Describe the bug
The file `app/helpers/avo/url_helpers.rb` defines `module UrlHelpers` but Rails 8's Zeitwerk expects `module URLHelpers` to match the filename convention. This causes `Avo::URLHelpers` to throw an uninitialized constant error.

### Reproduction repository for the bug
N/A - Issue exists in the gem itself, reproducible with any Rails 8.0.3 app.

### Steps to use in the reproduction repository
1. Create new Rails 8 app: `rails new avo_test --skip-everything -d sqlite3`
2. Add Avo 3.25.1 to Gemfile: `echo 'gem "avo", "~> 3.25.1"' >> Gemfile`
3. Run `bundle install`
4. Run `rails runner "puts Avo::URLHelpers.inspect"`
5. See error: `uninitialized constant Avo::URLHelpers`

### Expected behavior & Actual behavior
**Expected:** `Avo::URLHelpers` should be defined (Rails 8 Zeitwerk convention for `url_helpers.rb`)
**Actual:** `Avo::UrlHelpers` is defined instead, causing constant resolution errors

### System configuration
**Avo version**: 3.25.1

**Rails version**: 8.0.3

**Ruby version**: 3.4.5

**License type**:
 - [x] Community
 - [ ] Pro
 - [ ] Advanced

**Are you using Avo monkey patches, overriding views or view components?**
 - [ ] Yes. If so, please post code samples.
 - [x] No

### Screenshots or screen recordings
```bash
# Evidence from gem source:
grep -n "module" $(bundle show avo)/app/helpers/avo/url_helpers.rb
# Output:
# 1:module Avo
# 2:  module UrlHelpers  # <-- Should be URLHelpers
```
### Additional context
Rails 8 has been GA since November 2024. This naming convention incompatibility prevents Rails 8 adoption. The fix is straightforward: rename module UrlHelpers to module URLHelpers in the gem.

#### Impact

 - [x] High impact (It makes my app un-usable.)
 - [ ] Medium impact (I'm annoyed, but I'll live.)
 - [ ] Low impact (It's really a tiny thing that I could live with.)

#### Urgency

 - [x] High urgency (I can't continue development without it.)
 - [ ] Medium urgency (I found a workaround, but I'd love to have it fixed.)
[...]

PR #4204, opened 2026-01-11
fix: url inflection causes zeitwerk error

hmk:
# Description
If a Rails application inflects 'url', Avo installation fails. This fixes it.

Before:
```bash
/rails# bundle exec bin/rails g avo:install
bundler: failed to load command: bin/rails (bin/rails)
/usr/local/bundle/gems/zeitwerk-2.7.4/lib/zeitwerk/loader/callbacks.rb:31:in `on_file_autoloaded': expected file /usr/local/bundle/bundler/gems/avo-81ac4b40002c/lib/avo/acronym_support.rb to define constant Avo::AcronymSupport, but didn't (Zeitwerk::NameError)
```

After:
```bash
/rails# bundle exec bin/rails g avo:install
Starting application configuration...
Avo community 3.28.0
       route  mount_avo
      create  config/initializers/avo.rb
  ActiveRecord::SchemaMigration Load (3.8ms)  SELECT "schema_migrations"."version" FROM "schema_migrations" ORDER BY "schema_migrations"."version" ASC
```

Fixes https://github.com/avo-hq/avo/issues/4097

# Checklist:
- (not relevant) I have made corresponding changes to the [documentation](https://github.com/avo-hq/avodocs) 

## Manual review steps

Create a rails repo that inflects url

Run `bundle install` and then `bundle exec bin/rails g avo:install` - it will fail

Add to Gemfile:
```ruby 
gem "avo", git: "https://github.com/hmk/avo.git", branch: "fix/url-helper-inflection"
```
It should pass

### DIFF
--- a/app/helpers/avo/url_helpers.rb
+++ b/app/helpers/avo/url_helpers.rb
@@ -89,4 +89,7 @@ module Avo
       end
     end
   end
+  # Host apps may configure `inflect.acronym "URL"`, which makes Zeitwerk fail
+  URLHelpers = UrlHelpers unless const_defined?(:URLHelpers, false)
+  UrlHelpers = URLHelpers unless const_defined?(:UrlHelpers, false)
 end
--- a/lib/avo.rb
+++ b/lib/avo.rb
@@ -2,6 +2,7 @@ require "zeitwerk"
 require "net/http"
 require_relative "avo/version"
 require_relative "avo/engine" if defined?(Rails)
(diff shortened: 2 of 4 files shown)


hmk:
after thinking about it for a bit... I don't like this approach either. The dual definition (because of when things get loaded) doesn't feel right.

Will think about it and see if I come up with a better approach.

PR #4205, opened 2026-01-12
remove color option in favor of options on badge field

Mihai-Munteanu:
# Description

Remove color option in favor of options for Badge DSL

Fixes # (issue)

### DIFF
--- a/spec/lib/avo/fields/badge_field_spec.rb
+++ b/spec/lib/avo/fields/badge_field_spec.rb
@@ -13,11 +13,11 @@ RSpec.describe Avo::Fields::BadgeField, type: :model do
     field.instance_variable_set(:@options, options_hash)
   end
 
-  describe "#badge_color_for_value" do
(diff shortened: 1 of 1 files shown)


lib/avo/fields/badge_field.rb
```diff
@@ -19,9 +18,8 @@ def options_for_filter
       end
 
       def color
-        # Priority 1: Use explicit color if provided (via proc/lambda or direct value)
-        # Priority 2: Fall back to automatic color detection based on field value and options mapping
-        execute_context(@color) || badge_color_for_value
+        # Maps field value to a color based on @options configuration
+        badge_color_for_value
```
Paul-Bob: Since `color` returns `badge_color_for_value` and `badge_color_for_value` is not called anywhere else, I suggest removing the `badge_color_for_value` method and moving the entire logic into `color`.

PR #4234, opened 2026-01-30
UI4: inputs

Mihai-Munteanu:
# Description
polish design specifications for text inputs based on figma https://www.figma.com/design/Enmp0p0u8wFG0SuysUDszX/Avo-Design-System?node-id=2934-86442&m=dev

# Checklist:

## Screenshots & recording
Screenshot 2026-02-10 at 12 40 33
Screenshot 2026-02-10 at 12 39 54
Screenshot 2026-02-10 at 12 39 42

### DIFF
--- a/spec/system/avo/code_field_spec.rb
+++ b/spec/system/avo/code_field_spec.rb
@@ -114,6 +114,7 @@ RSpec.describe "CodeField", type: :system do
       expect(JSON.parse(json_text, symbolize_names: true)).to eq(metadata)
 
       click_on "Edit"
+      wait_for_loaded
 
       json_text = page.evaluate_script('document.querySelector(".CodeMirror").CodeMirror.getValue()')
       expect(JSON.parse(json_text, symbolize_names: true)).to eq(metadata)


lib/avo/fields/concerns/is_loading.rb
```diff
@@ -0,0 +1,17 @@
...
+module Avo
+  module Fields
+    module Concerns
+      module IsLoading
+        extend ActiveSupport::Concern
+
+        attr_reader :loading
+
+        def loading?
+          !!Avo::ExecutionContext.new(target: loading, record: record, view: view, resource: resource).handle
```
adrianthedev: Why the `!!`?

```suggestion
          !!Avo::ExecutionContext.new(target: loading, record: record, view: view, resource: resource).handle
```

lib/avo/fields/concerns/is_loading.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Avo
+  module Fields
+    module Concerns
+      module IsLoading
+        extend ActiveSupport::Concern
+
+        attr_reader :loading
```
adrianthedev: do we ever add this loading attribute? I don't recall having that.

Ideally, we should be triggering loading with just a class or HTML attribute, not by reloading the field from the server.

app/components/avo/field_wrapper_component.html.erb
```diff
@@ -34,7 +34,10 @@
           <% end %>
         </div>
       <% elsif on_edit? %>
-        <%= content %>
+        <div class="field-wrapper__input">
+          <%= content %>
+          <%= tag.span class: "input__spinner", "aria-hidden": "true" if @field.loading? %>
```
adrianthedev: Can we add the spinner as an after element and not have to depend on adding an extra element on the page?

adrianthedev (changes requested):
1. the text input size doesn't match figma. can you pease cehck the others too? (don't set the height. see what's wrong with the sizes (padding, line height, font-size))
2. let's try to add the spinner element as an after element.

CleanShot 2026-01-30 at 11 19 04@2x
CleanShot 2026-01-30 at 11 19 13@2x

.github/workflows/danger.yml
```diff
@@ -8,6 +8,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
```
adrianthedev, line 11: What is the reason for this?

app/assets/stylesheets/css/components/input/base.css
```diff
@@ -0,0 +1,38 @@
+/* Input Component - Base */
+/* BEM: .input-field (block), .input-field__* (element), .input--* (modifier) */
+
+/* Replaced input heights (range, color): content + padding + border */
```
adrianthedev:
```suggestion
```

It doesn't seem that these help in any way.

app/assets/stylesheets/css/components/input/base.css
```diff
@@ -0,0 +1,38 @@
...
+:root {
+  --input-height-sm: calc(1rem + 0.5rem + 2px);
+  --input-height-md: calc(1.25rem + 0.75rem + 2px);
+  --input-height-lg: calc(1.25rem + 1.25rem + 2px);
```
adrianthedev: Can we figure these out from actual variables?
I will always wonder where does the 1rem come from. or the 0.5rem or the 2px.
Are these spacings? do they have variables in Tailwind? Can we leave comments on what they are?

app/assets/stylesheets/css/components/input/color.css
```diff
@@ -0,0 +1,23 @@
+/* ==========================================================================
+  Color Input
+  ========================================================================== */
+
+input[type="color"] {
```
adrianthedev: this is still an input. let's move it into `input.css`

app/assets/stylesheets/css/components/input/password.css
```diff
@@ -0,0 +1,83 @@
...
+.input-field ~ .input-field__password-toggle svg,
+input ~ .input-field__password-toggle svg,
+.input--size-md ~ .input-field__password-toggle svg {
+  @apply size-4;
+}
+
+/* Large size */
+.input--size-lg ~ .input-field__password-toggle svg {
+  @apply size-4.5;
+}
```
adrianthedev: can we have one CSS rule to dictate the size of the after element? I mean for password icon, calendar icon, loadign icon, etc.?

Also maybe make it a variable and let CSS apply it?

PR #4247, opened 2026-02-06
enhancement: validate record when destroying attachments

### LINKED ISSUE #4225: Deleting an attachment bypasses validation
### Describe the bug

Using delete attachment button on view or edit page bypasses validation. This can result in an object which has to have an attachment but doesn't.

Attachment validator is using `active_storage_validations` gem and works as expected when creating or editing an object using Avo.

However using the delete button doesn't trigger validation failure.

### Reproduction repository for the bug

Sorry I don't have a repo right now. It should be easy to reproduce with this:

```ruby
# Gemfile
gem 'active_storage_validations'
```

```ruby
# Model
  validates :file, attached: true
```

My real life scenario — I have a document model which should have either URL or file attached:

```ruby
class Document < ApplicationRecord
  has_one_attached :file

  validates :name, presence: true
  validates :url, format: { with: URI::DEFAULT_PARSER.make_regexp }, allow_nil: true
  validates :file, attached: true, unless: -> { url.present? }

  def url_or_blob
    url.presence || file
  end
end
```

Avo fields:
```ruby
  def fields
    field :name, as: :text,
    field :url, as: :text, placeholder: 'Link to the document or form',
      format_index_using: -> { value.present? ? value.truncate(50) : value },
      help: 'Make this empty to use the uploaded file instead of the URL'
    field :file, as: :file, hide_on: :index
  end
```

### Steps to use in the reproduction repository

1. Add this configuration '...'
2. Go to '...'
3. Click on '....'
4. Scroll down to '....'
5. See error

Avo should not be able to detach/delete an attachment/file when model validates `attached: true`.

Currently Avo detaches/removes it which results in an unexpected behavior.

### System configuration
**Avo version**: 3.28.0

**Rails version**: 8.1.2

**Ruby version**:  3.4.8

**License type**:

 - [x] Community
 - [ ] Pro
 - [ ] Advanced

 - [x] No

Image

#### Impact

#### Urgency

Paul-Bob:
# Description

Fixes #4225

Validate the record and roll back the attachment destruction if the record is not valid after destroying the attachment.

### DIFF
--- a/spec/system/avo/attachment_destroy_spec.rb
+++ b/spec/system/avo/attachment_destroy_spec.rb
@@ -1,0 +1,88 @@
+require "rails_helper"
+
+RSpec.describe "Attachment destroy", type: :system do
+  describe "destroying an attachment" do
+    context "when validation fails after destroying attachment" do
+      let(:post_record) do
+        create(:post) do |p|
+          p.cover_photo.attach(
+            io: File.open(Rails.root.join("db", "seed_files", "iphone.jpg")),
+            filename: "iphone.jpg",
+            content_type: "image/jpeg"
+          )
+        end
+      end
+
+      it "rolls back the transaction and keeps the attachment" do
+        attachment_id = post_record.cover_photo.id
+
+        # Make the name blank so validation fails on save
+        # (name has presence: true validation that definitely works)
+        post_record.update_column(:name, nil)
+
+        visit "/admin/resources/posts/#{post_record.to_param}"
+
+        destroy_path = "/admin/resources/posts/#{post_record.to_param}/active_storage_attachments/cover_photo/#{attachment_id}"
+
+        expect {
+          accept_custom_alert do
+            find("a[data-turbo-method='delete'][href='#{destroy_path}']").click
+          end
+          wait_for_loaded
+        }.not_to change { ActiveStorage::Attachment.count }
+
+        # Flash should contain validation error
+        expect(page).to have_text("Name can't be blank")
+
+        # Attachment should still exist
+        expect(ActiveStorage::Attachment.exists?(attachment_id)).to be true
+
+        # Post should still have the cover_photo
+        post_record.reload
+        expect(post_record.cover_photo).to be_attached
+      end
+    end
+
+    context "when destroying an attachment succeeds" do
+      let(:post_record) do
+        create(:post) do |p|
+          p.cover_photo.attach(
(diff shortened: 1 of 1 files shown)


app/controllers/avo/attachments_controller.rb
```diff
@@ -28,10 +28,22 @@ def destroy
       if authorized_to :delete
         attachment = ActiveStorage::Attachment.find(params[:attachment_id])
```
xeron, line 29: Not directly related to the PR and bug in question but I believe this should be scoped to the `@record`, otherwise these 2 requests do exactly the same?
```
DELETE /events/1/active_storage_attachments/photos/123
DELETE /cities/1/active_storage_attachments/photos/123
```

Something like:
```ruby
attachment = @record.send(params[:attachment_name]).attachments.find(params[:attachment_id])
```
and maybe even involving `attachment_name` validation like `create` does here:
```ruby
association_name = BaseResource.valid_attachment_name(@record, params[:attachment_key])
```

xeron:
Thanks, I'll try this in my app. Added a small comment, not sure how important this is.

xeron:
Tested and it works. Thanks!

app/controllers/avo/attachments_controller.rb
Paul-Bob: If the `ActiveStorage::Attachment` is the same object, in your example with id `123`, then even if we scope it to the record, we will still be deleting the same record from events or from cities.

xeron: If you scope to the record then if attachment doesn't belong to it my code example will trigger `ActiveRecord::RecordNotFound`.

Assuming `123` belongs to the event. Current code:

```
DELETE /events/1/active_storage_attachments/photos/123 => Attachment 123 deleted
DELETE /cities/1/active_storage_attachments/photos/123 => Attachment 123 deleted
```

Proposed change:

```
DELETE /events/1/active_storage_attachments/photos/123 => Attachment 123 deleted
DELETE /cities/1/active_storage_attachments/photos/123 => 404
```

Paul-Bob: > Assuming `123` belongs to the event.

If `123` belongs to the event, then Avo will never render this request anywhere in the UI `/cities/1/active_storage_attachments/photos/123`.

But I get the idea. You could edit the link in the DOM or send a custom request. Would you like to give it a try with a contribution?

PR #4249, opened 2026-02-06
fix: add disconnect() cleanup to Stimulus controllers

rickychilcott:
## Summary
- Fixes `TypeError: Cannot read properties of null (reading 'scrollTop')`) and similar errors caused by Stimulus controllers not cleaning up in `disconnect()`
- Adds `disconnect()` to 7 controllers: `tags_field`, `easy_mde`, `tippy`, `preview`, `table_row`, `media_library_attach`, `search`
- Fixes broken `removeEventListener` in `record_selector` (`.bind()` created non-matching refs so removal was a no-op)

## Test plan
- [x] Visit an Avo page with tag fields, navigate away via Turbo, confirm no console errors
- [x] Test EasyMDE markdown fields across Turbo navigations
- [x] Test tooltip (tippy) fields across Turbo navigations
- [x] Test shift-click multi-select on index tables (record_selector) — previously accumulated duplicate listeners on each shift-click
- [x] Test global search keyboard shortcut (Cmd+K / Ctrl+K) works after Turbo navigation
- [x] Test media library file upload after Turbo navigation (no duplicate file inputs)

### DIFF
--- a/app/javascript/js/controllers/fields/easy_mde_controller.js
+++ b/app/javascript/js/controllers/fields/easy_mde_controller.js
@@ -28,9 +28,16 @@ export default class extends Controller {
       options.status = false
     }
 
-    const easyMde = new EasyMDE(options)
+    this.easyMde = new EasyMDE(options)
     if (this.view === 'show') {
-      easyMde.codemirror.options.readOnly = true
+      this.easyMde.codemirror.options.readOnly = true
+    }
+  }
+
+  disconnect() {
+    if (this.easyMde) {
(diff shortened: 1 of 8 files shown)


rickychilcott:
I noticed some of these errors over the years, but finally got motivated to fix them. Claude found some other issues along the way. I've tested all myself manually in my app, except for the EasyMDE fix but it seems pretty benign.

Hope this helps!

Paul-Bob (approved):
LGTM!

Thank you @rickychilcott for this contribution!

Issue #3959, opened 2025-07-05
Add support for Active Record Encryption in fields


adrianthedev:
### Feature

Active Record Encryption enables developers to easily hide specific bits of information behind encryption.

We should enable an easy way to display those fields after the record is displayed.

### Workflow:

- this should be a new field
- The user loads the record in any view
- Avo shows the record with asteriscs (****)
- the user get's a button to reveal the value
- Avo makes a call to the server to get the real value and display it to the user
- we should  take care on forms to not submit the easteriscs and overwrite the values in the DB

### Current workarounds

Custom fields.

piyush-pawar828:
Hey @adrianthedev , I would like to work on this.

Paul-Bob:
Thanks @piyush-pawar828, I've assigned it to you

wencano:
Hi! I’ve opened a PR with an implementation for this: [https://github.com/avo-hq/avo/pull/4445](https://github.com/avo-hq/avo/pull/4445)

@piyush-pawar828 I see you’re assigned here; I didn’t find overlapping work in flight when I started. If you already have a branch or preferred approach, I’m happy to adjust or defer—just let me know. Thanks, and no pressure.

Paul-Bob:
Thank you @wencano!

Paul-Bob:
https://linear.app/avo-hq/issue/AVO-1553/field-level-pii-masking-with-reveal-on-click

Issue #2168, opened 2023-12-18
Add a `reveal` field which can hide some text from the user until they hit the reveal button


adrianthedev:
This field would be rendered only on display fields (index and show).

### My approach

It would show the value as a password and then have a button that would change the input type to text so the value becomes visible.

We need to ensure that the value is not clearly visible if the use inspects the DOM.

jacobgwilliams:
I'd like to take a stab at this. Planning to add a display-only (`index`/`show`) `reveal` field that:

- Renders a masked value with an eye toggle (same UX pattern as password `revealable`)
- Keeps the plaintext out of the initial DOM (Base64-encoded payload decoded only on reveal)

Will open a PR soon.

Paul-Bob:
https://linear.app/avo-hq/issue/AVO-1553/field-level-pii-masking-with-reveal-on-click

PR #4445, opened 2026-04-27
Add encrypted_text field for Active Record Encryption

wencano:
# Description
- Adds a field type for attributes protected with [Active Record Encryption](https://guides.rubyonrails.org/active_record_encryption.html).
- **Index / show:** masked display.
- **Edit:** form does not echo plaintext; leaving the input **blank** does not overwrite stored ciphertext.
- **Reveal (optional):** button loads the real value through an **authenticated JSON** endpoint (not embedded in the initial HTML).

Fixes # [(issue)](https://github.com/avo-hq/avo/issues/3959)

**Assignment**

- Issue is assigned to **@piyush-pawar828**; I did not see an open PR when I started.
- If you already have work in progress, I am happy to **align**, **combine**, or **close** this in favor of your branch—just say the word.

## What changed

- **`Avo::Fields::EncryptedTextField`** — extends the text field; `#fill_field` ignores blank submissions and the mask string so existing encrypted values are preserved unless the user enters a new value.
- **UI** — show/index components display a fixed mask; show includes a reveal control when `revealable: true` (default).
- **`Avo::EncryptedFieldsController#reveal`** — resolves the resource field, authorizes with `show`, and returns `{ value: ... }` as JSON. Runs `detect_fields` before lookup so `get_field_definitions` is populated (same need as other resource screens).
- **Routing** — `GET /avo_api/resources/:resource_name/:id/encrypted_fields/:field_id` (engine scope, name `avo_api_reveal_encrypted_field`).
- **Stimulus** — `encrypted-field` (reveal fetch) and `encrypted-field-edit` registered in the JS entrypoint.
- **Tests** — `spec/features/avo/encrypted_text_field_spec.rb` covers masking, `revealable`, and `fill_field` behavior using a small in-memory record (no dummy schema required for the spec).

# Checklist:

## Screenshots & recording

image

image

image

## How to verify manually

In an app that already uses Active Record Encryption on a column, add to a resource:

```ruby
field :some_encrypted_column, as: :encrypted_text
```

- **List / show:** mask is shown.
- **Edit:** no plaintext prefill; **empty submit** leaves the stored value unchanged.
- **Reveal:** returns decrypted value for **authorized** users.


Paul-Bob (changes requested):
Thanks for looking into this @wencano!

Can we use a turbo solution instead of stimulus?

Paul-Bob:
Thanks for submitting this PR, @wencano.

I'm closing it for now until we untangle the requirements for this field. You worked on https://github.com/avo-hq/avo/issues/3959, but there is a similar issue, https://github.com/avo-hq/avo/issues/2168, which also received a PR. They are kind of the same concept, but with different field names.

I'm closing both for now until we define and merge the requirements and consolidate the related tasks.

PR #4446, opened 2026-04-27
Add DuplicateRecord action for easy record cloning

wencano:
# Description

- Adds a **built-in row action** to duplicate any record with a single click.
- **No confirmation modal** — creates the copy immediately and shows a **toast** notification.
- Configurable via `self.duplicate_config` on the resource:
  - `title_field` — which field gets the " Copy" suffix (auto-detects `name`, `title`, `label`, `subject`).
  - `title_suffix` — customize the suffix (default: `" Copy"`).
  - `exclude` — fields to clear on the duplicate (e.g., `[:slug, :uuid]`).
  - `defaults` — static or callable values for the new record.
  - `copy_attachments` — `true` (all), `[:image]` (specific), or `false` (none).

Fixes # none

## What changed

- **`Avo::Actions::DuplicateRecord`** — new action class in `lib/avo/actions/`; uses `record.dup`, applies config, handles Active Storage attachments.
- **`Avo::Resources::Base`** — added `class_attribute :duplicate_config` so resources can configure duplication behavior.
- **Tests** — `spec/features/avo/duplicate_record_action_spec.rb` covers field exclusion, title suffix, defaults (static + callable), auto-detection, and real `Product` model integration.

# Checklist:

## Screenshots & recording

image

image

## How to verify manually

Add to any resource:

```ruby
self.duplicate_config = {
  title_field: :name,
  title_suffix: " Copy",
  exclude: [:slug],
  copy_attachments: true
}

def actions
  action Avo::Actions::DuplicateRecord
end
```

- **Index:** select a row → click "Duplicate" → new record created with " Copy" suffix.
- **Show:** actions dropdown → "Duplicate" → same result.
- **Attachments:** if `copy_attachments` is enabled, files are cloned to the new record.

### DIFF
--- a/lib/avo/actions/duplicate_record.rb
+++ b/lib/avo/actions/duplicate_record.rb
@@ -1,0 +1,125 @@
+# frozen_string_literal: true
+
+module Avo
+  module Actions
+    class DuplicateRecord < Avo::BaseAction
+      self.name = -> { I18n.t("avo.duplicate_record", default: "Duplicate") }
+      self.no_confirmation = true
+      self.visible = -> { view.show? || view.index? }
+
+      class << self
+        def duplicate_config_for(resource_class)
+          resource_class.try(:duplicate_config) || {}
+        end
+      end
+
+      def handle(records:, **args)
+        record = records.first
+
+        if record.blank?
+          error I18n.t("avo.no_record_to_duplicate", default: "No record to duplicate")
+          return
+        end
+
+        config = self.class.duplicate_config_for(@resource.class)
+        new_record = duplicate_record(record, config)
+
+        if new_record.save
+          succeed I18n.t("avo.record_duplicated", default: "Record duplicated successfully")
+          reload
+        else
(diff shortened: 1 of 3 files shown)


Paul-Bob:
Thanks for this submission, @wencano.

We do not ship built in actions. For this kind of functionality, we usually provide documentation as a guide. For example, a bulk destroy records action is documented here: https://docs.avohq.io/3.0/guides/bulk_destroy_action_using_customizable_controls.html

I am closing this PR since we do not ship built in actions. However, could you tell me more about the origins of this PR? Is there any issue opened on this subject? Is it something that you personally needed and used at some point? I can see the value in it, and we may end up creating a guide for this use case.

adrianthedev:
Just to weigh in for a bit.
It's a great idea and we'll probably add it sooner rather than later.
We just need to figure out the mechanisms around it. How can a user modify or replace it, authorization, etc.

Thanks for sending this in @wencano

PR #4456, opened 2026-05-08
chore: remove invalid `style: 'hidden'` from filter url_redirect links

ursm:
# Description

The `url_redirect` `link_to` calls in the filter partials pass `style: 'hidden'`, which generates an invalid `style="hidden"` attribute (`hidden` is not a valid CSS property/value).

The same elements already use `class: 'hidden'` (Tailwind's `display: none`), so the `style` option is redundant and has no effect on the rendered output. This change just removes the dead option.

Files touched:

- `app/views/avo/base/_boolean_filter.html.erb`
- `app/views/avo/base/_date_time_filter.html.erb`
- `app/views/avo/base/_multiple_select_filter.html.erb`
- `app/views/avo/base/_select_filter.html.erb`
- `app/views/avo/base/_text_filter.html.erb`

# Checklist:

## Manual review steps

1. Open any resource index page that has filters (boolean / date-time / multiple-select / select / text).
1. Open the filters panel and inspect the `<a>` element that has `class="hidden"` inside each filter partial.
1. Confirm there is no longer a `style="hidden"` attribute on the link, and that the link continues to be hidden via the existing `hidden` class.
1. Apply / change a filter and confirm the URL-redirect behavior driven by the corresponding Stimulus controller still works as before.

### DIFF
--- a/app/views/avo/base/_boolean_filter.html.erb
+++ b/app/views/avo/base/_boolean_filter.html.erb
@@ -23,7 +23,7 @@
           <% end %>
         </div>
       <% end %>
-      <%= link_to 'url_redirect', request.url, data: { 'boolean-filter-target': 'urlRedirect', 'turbo-frame': params[:turbo_frame] }, style: 'hidden', class: 'hidden' %>
+      <%= link_to 'url_redirect', request.url, data: { 'boolean-filter-target': 'urlRedirect', 'turbo-frame': params[:turbo_frame] }, class: 'hidden' %>
     </div>
   <% end %>
 </div>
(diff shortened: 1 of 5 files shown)


adrianthedev:
@cursoragent simulate this bug. Is it true?

Paul-Bob (approved):
LGTM! 

Thank you @ursm!

Paul-Bob:
@adrianthedev this is not necessary a bug, the `hidden` class already hides the link. But `style="hidden"` generates invalid CSS (it's ignored by the browser), so is ok to remove the `style` attribute to avoid confusion.

PR #4460, opened 2026-05-10
Fix Trix dark mode code styling

hksharma2004:
## Summary

Fixes Trix field `pre` and `code` styling so code blocks use Avo theme tokens and render correctly in dark mode.

## Why

Trix ships default styles for rendered content, including fixed styling for `pre` blocks. In Avo dark mode, those defaults did not fully adapt to the active theme, which made code blocks look inconsistent inside Trix fields.

## Changes

- Theme Trix `pre` and `code` blocks with Avo token-based colors.
- Keep nested `pre code` content transparent so the parent code block background is used.
- Update the Trix editor and toolbar styling so they render as one connected control.
- Make toolbar button groups compact and horizontally scrollable.
- Keep toolbar spacing RTL-safe with logical margin utilities.
- Keep inactive dark-mode toolbar icons inverted while selected icons remain readable on the active background.

## Verification

- `yarn build:css`
- `git diff --check`


Paul-Bob:
Thank you for this PR submission @hksharma2004!

adrianthedev:
Thanks @hksharma2004 for your work ❤️

PR #4461, opened 2026-05-10
chore: ensure tags spec restores resource items on failure

ursm:
# Description

Move `restore_items_from_backup` calls from the end of test bodies into `ensure` clauses in `spec/system/avo/tags_spec.rb`, so that class-level `temporary_items` state is reliably cleaned up even when an expectation fails mid-test.

## Why

`Avo::Resources::*.with_temporary_items` stores the given block on the class as `temporary_items`. `lib/avo/resources/base.rb#detect_fields` then `instance_eval`s that block on every subsequent `detect_fields` call until `restore_items_from_backup` is called. If a test fails before its trailing `restore_items_from_backup`, the block leaks into following tests, replacing their default field set and producing cascading, hard-to-debug failures (e.g. fields like `:name` disappearing from `Avo::Resources::Project` because a leaked `with_temporary_items` redefined the field list).

This was reproducible with non-cuprite drivers where the `mode: :select` tests are sensitive to layout-dependent Tagify overflow detection — a single failure there would corrupt unrelated tests later in the suite.

Using `ensure` directly on the `do/end` block (Ruby 2.5+) keeps the change minimal and matches the existing call style.

## Notes

- The first test in the `format_using ... with fetch_values_from` describe (`fetches the labels`) is **intentionally** left without `ensure`. The next test (`keep correct tags on validations error and edit`) inherits its `fetch_values_from` / `format_using` setup via the leaked `temporary_items`, and that second test's `ensure` performs the cleanup for both. An inline comment documents this so future readers don't silently break the dependency.
- A more thorough fix would be to make `with_temporary_items` itself scope-bound (yield + `ensure`), but that requires changing the API across ~20 spec files and is out of scope for this change.

# Checklist:

## Screenshots & recording

N/A — test-only change.

## Manual review steps

1. Run \`bundle exec rspec spec/system/avo/tags_spec.rb\` and confirm tests still pass.
2. Optional: deliberately make one of the affected tests fail mid-body (e.g. add \`expect(true).to be false\` after \`with_temporary_items\` in \`mode: :select on edit / update\`). Before this change, subsequent tests in the suite would see leaked \`temporary_items\` on \`Avo::Resources::Project\`. With this change, \`Avo::Resources::Project.temporary_items\` is \`nil\` after the failed test.


spec/system/avo/tags_spec.rb
```diff
@@ -181,6 +181,10 @@
     let(:field_value_slot) { tags_element(find_field_value_element("skills")) }
     let(:tags_input) { field_value_slot.find("span[contenteditable]") }
 
+    # Intentionally no `ensure` to restore items: the next test
+    # (`keep correct tags on validations error and edit`) inherits the
+    # `fetch_values_from` / `format_using` setup from this one's leaked
+    # `temporary_items`. That test's `ensure` performs the cleanup for both.
```
Paul-Bob:
```suggestion
```

Paul-Bob:
Thank you for this contribution @ursm!

PR #4469, opened 2026-05-14
chore: ensure remaining specs restore resource items on failure

ursm:
# Description

Follow-up to #4461. Applies the same `ensure`-based cleanup pattern to every other spec that has inline `restore_items_from_backup` calls inside `it` blocks.

## Why

`Avo::Resources::*.with_temporary_items` stores the given block on the class as `temporary_items`. `lib/avo/resources/base.rb#detect_fields` then `instance_eval`s that block on every subsequent `detect_fields` call until `restore_items_from_backup` is called. If a test fails before its trailing `restore_items_from_backup`, the block leaks into following tests, replacing their default field set and producing cascading, hard-to-debug failures.

## Audit

Every spec file calling `with_temporary_items` / `restore_items_from_backup` was checked. Three categories:

| Category | Action |
|----------|--------|
| **A** — restore is already in `before`/`after`/`around` hooks | No change |
| **B** — `restore_items_from_backup` is the trailing line of an `it` block | Move into `ensure` |
| **C** — intentional inter-test leak (e.g. the `format_using` describe in `tags_spec.rb`) | No change |

Files left untouched (Category A): `default_sort_column_and_direction_spec.rb`, `has_field_discovery_spec.rb`, `date_time_western_spec.rb`, `fields_view_type_spec.rb`, `tabs_panels_and_sidebar_visibility_spec.rb`, `default_field_spec.rb`, `boolean_field_spec.rb`, `time_spec.rb`, `sti_person_spouse_spec.rb`, `stars_field_spec.rb`, `date_time_eastern_spec.rb`, `select_field_spec.rb`, `date_time_utc_spec.rb`, `date_time_spec.rb`, `timezone_spec.rb`.

Files changed (Category B), 5 files / 11 tests:
- `spec/features/avo/has_many_field_spec.rb` — 2 tests
- `spec/features/avo/number_spec.rb` — 5 tests
- `spec/system/avo/create_via_belongs_to_spec.rb` — 1 test
- `spec/system/avo/tabs_spec.rb` — 1 test
- `spec/system/avo/trix_field_spec.rb` — 3 tests

Additionally, one file was missing the `restore_items_from_backup` call entirely (a clear pre-existing bug — the leaked items would persist for the rest of the suite):
- `spec/features/avo/search_has_many_scope_spec.rb` — 1 test, added `ensure restore_items_from_backup`

## Open question: should `with_temporary_items` itself be scope-bound?

The mechanical `ensure` pattern works but duplicates the same boilerplate at every call site. The root cause is that `with_temporary_items` is not actually scope-bound — it just sets class state and forgets. The block it takes is the items definition (later `instance_eval`'d in `detect_fields`), not a scope block.

A cleaner API would let callers write:

```ruby
Avo::Resources::Project.with_temporary_items(-> {
  field :stage, as: :tags, mode: :select
}) do
  visit avo.resources_projects_path
  expect(page).to have_text(\"activity_category2\")
end
# items automatically restored, even on failure
```

Implementation sketch in `lib/avo/concerns/can_replace_items.rb`:

```ruby
def with_temporary_items(items_proc = nil, &block)
  if items_proc.nil?
    # Legacy: block is the items definition, no scope.
    self.temporary_items = block
  else
    # New: positional items proc + scope block.
    self.temporary_items = items_proc
    yield
  end
ensure
  self.temporary_items = nil if items_proc
end
```

Or alternatively, introduce a separate method (e.g. `using_temporary_items`) to avoid the dual-mode API.

This would eliminate ~20 `ensure ... restore_items_from_backup` boilerplate calls across the suite. **Happy to send a follow-up PR doing the migration if you're open to it** — let me know which API shape you prefer (overload on `with_temporary_items`, new method name, or something else).

# Checklist:

## Screenshots & recording

N/A — test-only change.

## Manual review steps

1. Run the affected specs:
   ```
   bundle exec rspec \\
     spec/features/avo/has_many_field_spec.rb \\
     spec/features/avo/number_spec.rb \\
     spec/features/avo/search_has_many_scope_spec.rb \\
     spec/system/avo/create_via_belongs_to_spec.rb \\
     spec/system/avo/tabs_spec.rb \\
     spec/system/avo/trix_field_spec.rb
   ```
2. Optional: deliberately make one of the affected tests fail mid-body (e.g. add `expect(true).to be false` after `with_temporary_items`). Before this change, the class-level `temporary_items` would leak; with this change, it is cleared after the failure.


ursm:
Added a second commit to fix a pre-existing flaky test that surfaced in this PR's CI.

`spec/system/avo/tags_spec.rb:206` (`keep correct tags on validations error and edit` in the `format_using ... with fetch_values_from` describe) was failing because it implicitly depended on the leaked `with_temporary_items` state from the preceding `fetches the labels` test in the same describe. Under randomised test order — or when the leak just didn't land in the right place — the second test ran against the default Course resource, which doesn't have `fetch_values_from`, and failed.

The same failure has been hitting `main` CI (e.g. [system_specs job in run 25716241257](https://github.com/avo-hq/avo/actions/runs/25716241257) on commit 92dc420f).

Fix: moved the `with_temporary_items` setup into a `before` hook and the cleanup into an `after` hook in that describe. Both tests are now self-contained and independent of execution order. This is consistent with the pattern used in many other specs that were already in "Category A" in the audit above.

Paul-Bob:
> Open question: should with_temporary_items itself be scope-bound?

I think yes, your suggestion makes sense. At this point, we'll leave it as is since we're currently developing on both branches `main` for Avo 3 and `4-dev` for Avo 4, and this kind of change would result in conflicts that are not really necessary right now.

It’s a good idea to implement in the future, but for now let’s keep it as is.

Thanks for this submission and for the deep analysis. We really appreciate all the help and well argued opinions!

PR #4470, opened 2026-05-15
add mount_lookbook configuration option

Paul-Bob:
# Description

Avo no longer mounts Lookbook automatically. Lookbook can only be mounted once per app, so this avoids conflicts when the host app already mounts Lookbook.

Opt in via routes:
```ruby
mount_avo mount_lookbook: true
```

Default is off (`mount_lookbook: false`).


config/routes.rb
```diff
@@ -38,6 +38,8 @@
       get "/appearance", to: "private#appearance"
     end
 
-    mount Lookbook::Engine, at: "/lookbook" if defined?(Lookbook)
+    if Avo.configuration.mount_lookbook
+      mount Lookbook::Engine, at: "/lookbook" if defined?(Lookbook)
+    end
```
adrianthedev:
```suggestion
    mount Lookbook::Engine, at: "/lookbook" if Avo.configuration.mount_lookbook && defined?(Lookbook)
```

PR #4497, opened 2026-05-22
Adjust German translations to be gender-neutral

Timmitry:
# Description

The german language has three genders (masculin/feminin/neutral), which makes generic translations a bit harder.

For example, the translation `create_new_item: Neues %{item} erstellen` does not work, as it would have to be:

- Neues Projekt erstellen
- Neue Vorlage erstellen
- Neuen Datensatz erstellen

depending on the gender of the %{item}. I therefore adjusted the translations to be gender neutral.


Paul-Bob (approved):
Thank you @Timmitry! We really appreciate this kind of contributions! 🎉

Timmitry:
Thanks! And you guys are really quick to respond ⚡😃

PR #4534, opened 2026-06-08
Remove humanize and downcase usages

Timmitry:
# Description

Removes some usages of humanize and downcase.

Fixes parts of #4186

While in the english language, downcasing most words works well, in e.g. german, nouns are written with a capital letter at the start, so e.g. writing "New location" in german would need to be `Neuer Standort` and not `Neuer standort`. 

I'd like to point out that I did some experiments locally with rbspy to find out why Avo is not as fast as I'd expect it to be - and in the flamegraph, the `humanize` method was very present, as it uses a Regex-Replace with gsub internally. I'm not sure if this is really a culprit for less-than-ideal performance, but it could probably help to only call humanize when it's absolutely neccessary 🙂

### DIFF
--- a/app/components/avo/fields/preview_field/index_component.rb
+++ b/app/components/avo/fields/preview_field/index_component.rb
@@ -2,7 +2,7 @@
 
 class Avo::Fields::PreviewField::IndexComponent < Avo::Fields::IndexComponent
   def render_preview
-    link_to resource_view_path, title: t("avo.view_item", item: @resource.name).humanize do
+    link_to resource_view_path, title: t("avo.view_item", item: @resource.name) do
       helpers.svg(
         "tabler/outline/zoom-scan",
         class: "block h-6 text-content-secondary",
--- a/app/components/avo/resource_component.rb
+++ b/app/components/avo/resource_component.rb
@@ -250,7 +250,7 @@ class Avo::ResourceComponent < Avo::BaseComponent
         turbo_method: :delete,
         turbo_confirm: "Are you sure you want to detach this #{title}."
       } do
-      control.label || t("avo.detach_item", item: title).humanize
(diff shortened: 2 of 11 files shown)


Paul-Bob (changes requested):
Thanks for the submission @Timmitry.

While this PR might solve some issues, it also introduces others. I noticed that the tests are failing, and after reproducing each failing test locally, I found several areas where this PR is not ready yet.

For example, the breadcrumbs show "edit" instead of "Edit", and "detach" is not capitalized either, as shown in the screenshot:

Image

adrianthedev:
I wonder if this can be handled in the translation file? 
So each locale handles their own.

TBH, something was off for me about Avo's translations for a while now.
We have some which are capitalized in the translation files and some which are not.
I feel that we should have a rule for them.
Is there such a rule that is "universally accepted"?

Timmitry:
Yes, the PR is far from finished, but I'm glad that I started a discussion here 😃😉

I think that in the majority of cases, translations should be used as provided. So when there's a translation for "create", it should be `Create new %{model}` and not `create new %{model}`. The problem in e.g. english with this is that `model` will e.g. be `Location` (capital L), so this will read `Create new Location`. Some may argue that `Create new location` might be more correct, however, I feel like capitalizing the model name here is often done and well accepted, at least for headings...

Paul-Bob:
I agree with all of this. One thing I'm still wondering about: what happens when the same translation is used in different positions within a sentence? (not sure if we have such cases, need to verifiy)

For example, sometimes it's used at the beginning of a phrase: "Create new ..."

and other times in the middle: "... create ..."

Depending on the language, capitalization, grammar, or even the translation itself might differ

Timmitry:
Fixed in #4639 🎉

PR #4545, opened 2026-06-09
UI4 - give the switcher controls a resting transparent outline at the focus

Mihai-Munteanu:
# Description
Fixes the focus-ring color flash on switcher controls and pagination page links, where focusing flashed a gray/black ring that morphed to the focus blue.

Cause
 These controls had no resting outline, so on focus outline-color animated from its default (the control's own gray/black currentColor) to the focus color. With a transition covering outline-color, that gray/black→blue interpolation was visible as a two-color morph.

# Checklist:

## Screenshots & recording

https://github.com/user-attachments/assets/d1876925-0a52-4a26-936d-66f22ac9ab0d
https://github.com/user-attachments/assets/f3b28830-b561-485f-86a8-caa58174fd34

### DIFF
--- a/app/assets/stylesheets/css/components/color_scheme_switcher.css
+++ b/app/assets/stylesheets/css/components/color_scheme_switcher.css
@@ -148,18 +148,6 @@
   }
 }
 
-/* Focus ring starts at the focus color but transparent, so it fades into view
-   rather than starting from the control's (gray) text color. Only the outline
-   color changes between resting and :focus-visible; width/offset stay constant. */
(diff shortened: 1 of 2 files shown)


adrianthedev (changes requested):
Something is off here.
We should not have to override it.
It should just work as the rest of the outlines, right?

here's a recording to show that these outline (pagination included) have a transition applied.

That needs removed.

https://github.com/user-attachments/assets/559f0f85-53ae-4436-96ce-1c0acb9ecfbe

PR #4573, opened 2026-06-15
Fix avo:update (double-run, wrong gem names, over-eager updates) + status page gem names

adrianthedev:
## What

Fixes the `avo:update` rake task and makes the debug **status page** show real gem names.

### 1. The task ran twice
`Rails::Engine#run_tasks_blocks` already auto-loads every `.rake` file under `lib/tasks/`. The engine *also* explicitly `load`ed `avo_tasks.rake` inside its `rake_tasks` block, so the file was loaded twice. Rake doesn't replace a redefined task — it appends the new block as an extra action and runs **all** actions on a single invocation. Result: `bundle update` (and every other task in that file) fired twice.

Removed the redundant `load`; kept the `assets:precompile` enhancement logic.

### 2. Bundler updated unrelated gems
Added `--conservative` so only the listed Avo gems are updated, not their shared dependencies.

### 3. Wrong gem names (`rhino` instead of `avo-rhino_field`)
Plugins register under nicknames — `avo-rhino_field` registers as `:rhino`, `avo-lexxy_field` as `:lexxy`, etc. — and some Avo gems don't even start with `avo-` (e.g. `marksmith`). So neither the nickname nor an `avo-*` bundle scan is reliable.

`PluginManager#register` now records the file that called it (`caller_locations`), and `Plugin#gem_name` matches that path against the bundle's gem paths to find the gem the plugin actually ships in. This resolves every registration style (`:rhino`, `"avo-menu"`, or a `Plugin` class) to the correct gem name. `avo:update` reads the plugin list from the plugin manager, resolves each to its gem name, and adds `avo` core explicitly (it doesn't self-register).

### 4. Status page showed nicknames
`Avo::Debug::StatusComponent` listed `plugin.name` (e.g. `rhino`). It now shows `plugin.gem_name` (e.g. `avo-rhino_field`), falling back to the nickname when the gem can't be resolved.

## Verification
- `caller_locations` correctly resolves the calling file even through a deferred `on_load` block.
- Booting the dummy app: the plugin registered as `:pluggy` resolves to gem name `"pluggy"`, and `avo:update`'s final list is `avo pluggy` — confirming nickname → real gem name resolution (including a non-`avo-` gem) and a single execution.

---

> [!NOTE]
> **Low Risk**
> Rake/plugin metadata changes with no auth or data-path impact; main risk is `avo:update` behaving differently (intended fix).
> 
> **Overview**
> Fixes **`avo:update`** so it runs once, targets the right gems, and doesn’t pull unrelated dependency bumps.
> 
> The engine **stops explicitly loading** `avo_tasks.rake` inside `rake_tasks` (Rails already loads `lib/tasks/*.rake`), which was causing Rake to **append duplicate actions** and run `bundle update` twice. The task now builds its gem list from **`avo`** plus each plugin’s resolved **`gem_name`**, uses **`bundle update --conservative`**, and bails with a message if nothing is found.
> 
> **Plugin registration** records the caller file via `caller_locations`, and **`Plugin#gem_name`** maps that path to the owning Bundler spec (so nicknames like `:rhino` resolve to `avo-rhino_field`). Minor cleanup in `Plugin`’s class helpers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6ff75a9432a7ade1a823c52a7291027919eb5d00. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lib/avo/engine.rb
```diff
@@ -29,8 +29,6 @@ class Engine < ::Rails::Engine
     isolate_namespace Avo
 
     rake_tasks do
-      load File.expand_path("../tasks/avo_tasks.rake", __dir__)
```
adrianthedev, line 32:
```ruby
# Rails::Engine already auto-loads every `.rake` file under `lib/tasks`,
# so we must NOT load avo_tasks.rake here too — doing so defines each task
# twice and Rake then runs every action block twice on a single invocation.
```

lib/avo/plugin_manager.rb
```diff
@@ -16,7 +16,12 @@ def reset
     end
 
     def register(name, priority: 10)
-      @plugins << Plugin.new(name:, priority: priority)
+      # Capture the file that called `register` so the plugin can later resolve
+      # the gem it actually ships in. Plugins register under nicknames
+      # (e.g. `:rhino` for `avo-rhino_field`), so the name alone isn't enough.
+      registered_from = caller_locations(1, 1)&.first&.path
```
adrianthedev, line 22: We should refactor this in the DSL method. So the plugin reports the gem name.

PR #4639, opened 2026-07-19
Trust resolved translations verbatim

adrianthedev:
# Description

Avo applied `.humanize`, `.capitalize`, or `.upcase_first` to strings it had just resolved from a locale file. That silently overrode the casing the developer wrote: `'Payment Intent ID'` rendered as `Payment intent id`, and `'API Products'` rendered as `Api products`. There was no way to opt out, and re-casing the locale entry could not help, because the cast was unconditional.

A resolved **resource or field** translation is now rendered verbatim. Humanization stays exactly where it belongs — on the names Avo generates for you when no translation resolves.

Rebased onto current `main`. #4637 and #4652 replaced `t(translation_key, …)` with `translate_field_name`, which already returns the resolved translation or the supplied default, so at the field layer this reduces to deleting the trailing cast.

## What changed

| Layer | Change |
| --- | --- |
| `lib/avo/fields/{base,array,has_many}_field.rb` | Drop `.humanize` / `.capitalize` from `translated_name` / `translated_plural_name` |
| `lib/avo/resources/base.rb` | `name_from_translation_key` and `navigation_label` no longer humanize; new `sentence_name` for mid-sentence interpolation |
| `lib/avo/resources/controls/save_button.rb` | Drop `.capitalize` from the resource-specific save label |
| `app/controllers/avo/{base,charts}_controller.rb`, `distribution_chart_full.html.erb` | 10 × `plural_name.humanize` → `plural_name` (page titles, breadcrumbs, chart heading) |
| `app/components/**` | has-one controls, and the `view_item` link titles |

Fallback behavior is unchanged, because callers already pass humanized defaults (`@id.to_s.humanize(keep_id_suffix: true)`, `demodulized_class_name.underscore.humanize`).

## Scope boundary

Avo's own chrome strings are **deliberately untouched**. `avo.en.yml` stores those lemmas lowercase on purpose — `new: new`, `edit: edit`, `undo: undo`, `view_item: view %{item}` — so the call-site `.humanize` is load-bearing display casing, not the bug. Removing it would ship lowercase chrome across all ten locales. Where one cast served both purposes (`view_item` interpolating a resource name), `.humanize` became `.upcase_first` plus `sentence_name`, so `view` still capitalizes while `API Product` survives.

Two sites were evaluated and left alone, both noted in the plan as deferred:

- `resource_controls_dropdown_component.rb:78` — feeds a case-insensitive `gsub` matcher, never rendered.
- `default_panel_name`'s `@params[:related_name].capitalize` — a request parameter, not an I18n value.

## ⚠️ Breaking change

Apps whose locale entries are lowercase relied on Avo capitalizing them. Those now render as written. Docs and upgrade guide: **avo-hq/docs.avohq.io#570** — please land the two together.

## Two corrections made during implementation

1. **`sentence_name` must mirror `#name`'s precedence.** A first pass short-circuited on `custom_name?`. But `has_field_discovery.rb:162` assigns a `name:` to *every* auto-discovered association, so `custom_name?` does not mean "developer authored it" — that rendered `Create new Main post` where Avo has always shown `Create new main post`. Caught by `create_has_one_spec`, and now locked in by an explicit guard example.
2. **`detach_item` was scoped back out.** Its `item` is a pre-computed component `title` serving both the panel heading (title case) and the detach label (lower case); one value cannot satisfy both. Left on `.humanize`; doing it properly needs a sentence form threaded through `ResourceShowComponent#title`.

# Checklist:

- [x] I have made corresponding changes to the documentation (avo-hq/docs.avohq.io#570)
- [ ] I tested the design in Light and Dark mode, and all default themes (not applicable — text casing only)

## Screenshots & recording

Not applicable; this changes label casing without touching layout.

## Manual review steps

1. Add `avo.resource_translations.<resource>.one/other` with an acronym (e.g. `'API Products'`) and confirm the index title, breadcrumb, and sidebar all show it verbatim.
2. Remove that translation and confirm the label falls back to the humanized class name.
3. Confirm the "New" and "Edit" breadcrumbs still render capitalized.

## Verification

`spec/features/avo/i18n_spec.rb` grew from 4 to 15 examples. Fixtures carry internal capitals on purpose — `.humanize` would render `Payment Intent ID` as `Payment intent id` and `.capitalize` would render `API People` as `Api people` — so any surface still casting a resolved translation fails. They are stored in-memory via `store_translations`, leaving the `i18n-tasks` lint suite untouched.

Locally: i18n spec 15/15; `spec/components` 215/215; `spec/system/avo/i18n_spec.rb` 3/3; `standardrb` clean. `spec/features spec/requests` was diffed against a `main` baseline on the same machine — **zero new failures**.

The `pt` courses assertion flips from `"Cursos"` to `"cursos"` on purpose. That fixture is lowercase in the locale file; asserting the capitalized form would mask exactly the behavior change being shipped.

Linear Issue: [AVO-1461](https://linear.app/avo-hq/issue/AVO-1461/trust-resolved-translations-verbatim)

### DIFF
--- a/app/components/avo/fields/gravatar_field/index_component.html.erb
+++ b/app/components/avo/fields/gravatar_field/index_component.html.erb
@@ -6,7 +6,7 @@
     rounded: @field.rounded,
     link_to_record: @field.link_to_record,
     link: resource_view_path,
-    title: t('avo.view_item', item: @resource.name).humanize
+    title: t('avo.view_item', item: @resource.name).upcase_first
   )
   %>
 <% end %>
--- a/app/components/avo/fields/gravatar_field/index_component.html.erb
+++ b/app/components/avo/fields/gravatar_field/index_component.html.erb
@@ -6,7 +6,7 @@
     rounded: @field.rounded,
     link_to_record: @field.link_to_record,
     link: resource_view_path,
-    title: t('avo.view_item', item: @resource.name).upcase_first
+    title: t('avo.view_item', item: @resource.sentence_name).upcase_first
   )
   %>
 <% end %>
--- a/app/components/avo/fields/has_one_field/show_component.html.erb
+++ b/app/components/avo/fields/has_one_field/show_component.html.erb
@@ -18,7 +18,7 @@
             turbo_frame: Avo::MODAL_FRAME_ID,
             target: :attach
           } do %>
-          <%= t("avo.attach_item", item: @field.name.humanize(capitalize: false)) %>
+          <%= t("avo.attach_item", item: @field.sentence_name) %>
         <% end %>
       <% end %>
 
@@ -29,7 +29,7 @@
           "data-turbo-frame": "_top",
           style: :primary,
           color: :accent do %>
-          <%= t("avo.create_new_item", item: @field.name.humanize(capitalize: false)) %>
+          <%= t("avo.create_new_item", item: @field.sentence_name) %>
         <% end %>
       <% end %>
     <% end %>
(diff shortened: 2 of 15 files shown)


Timmitry:
Just saw my Dependabot Update PR CI fail with this update - thank you so much @adrianthedev and @Paul-Bob !  🥳

I only found one spot where wrong casing is still applied - on a resource index page, the create new item button still downcases the first word ([relevant code](https://github.com/avo-hq/avo/blob/1bb0c119d097108a52f8271e248286fb085267d3/lib/avo/resources/controls/create_button.rb#L9)). Everything else seems to preserve the casing 💯

grafik

Paul-Bob:
@Timmitry `avo 4.0.22` includes a fix to the create new item button!

PR #4665, opened 2026-07-24
feat(sidebar): resizable sidebar

adrianthedev:
## Summary

The sidebar is a fixed 256px, so long resource names wrap onto two lines and the only escape is toggling the sidebar off entirely. This makes the sidebar/content boundary a drag handle, with the chosen width stored per browser and applied before first paint.

Two things about the sidebar are worth knowing up front:

- **A user who never drags sees exactly today's layout.** The handle draws no line at rest — the divider stays `.main-content`'s existing `border-s`, and the grip only fades in on hover, focus or drag. The idle `col-resize` cursor is the whole discoverability story.
- **`--sidebar-width` is consumed by two paths with different breakpoint behavior**, so this is not "store one number". The raw preference lives on `<html>` and is ungated; the *derivation* into `--sidebar-width` is gated behind `lg`. A desktop-chosen 480px therefore cannot become a 480px overlay on a phone.

```mermaid
flowchart TB
    COOKIE["cookie avo.sidebar.width"]
    CTRL["set_sidebar_width<br/>(after _authenticate!)"]
    SCRIPT["nonce'd pre-paint script"]
    HTML["&lt;html&gt; --sidebar-width-stored"]
    HOST[".main[data-controller~=sidebar]<br/>--sidebar-width (lg-gated clamp)"]
    SB[".avo-sidebar"]
    MAIN[".main padding-inline-start"]
    HANDLE[".sidebar-resize-handle"]
    JS["sidebar-resize controller"]

    COOKIE --> CTRL --> SCRIPT --> HTML --> HOST
    HOST --> SB
    HOST --> MAIN
    HOST --> HANDLE
    JS --> HTML
    JS --> COOKIE
```

## Decisions a reviewer should weigh

**The carrier is a nonce'd pre-paint `<script>`, not an inline `style` attribute.** Avo's layout contains zero inline `style` attributes today. Putting one on `<html>` would make every page of every Avo admin depend on `style-src 'unsafe-inline'` — which CSP Level 3 *ignores* when a nonce-source is present, i.e. the Rails-generated default. The failure would have been silent: 256px on every cold load, correct after the first Turbo visit. A CSSOM write from a nonce'd script has no `style-src` surface. The trade is that with JS disabled the sidebar renders at the default.

**`.container-small` had to change first.** It was a *fixed* `960px`, and a sidebar above ~287px overflows it on every show/edit page at 1280px. It is now `max-width`-based. Note the exact utility list — `xl:px-0 xl:w-full xl:max-w-240 xl:mx-auto` — because per Flexbox §4.2 auto cross-axis margins suppress `align-self: stretch`, so dropping `w-full` collapses the container to its content width on every show page. **This is a behavior change on a documented extension point:** a host's `.container-small { width: 1200px }` override via `avo-overrides` is now silently clamped by an Avo-owned `max-width` they never wrote.

**`pointermove`/`pointerup` bind to `window`, not to the handle.** Pointer capture is supposed to retarget them, and it reports success — `hasPointerCapture()` returns `true` and no `lostpointercapture` fires — but Chrome does not actually retarget moves under CDP-synthesized input, so a capture-dependent drag stops tracking as soon as the cursor leaves the 15px strip. Binding to `window` for the drag's duration makes this independent of whether capture holds anywhere. Capture is still requested; nothing depends on it.

**The grab zone is 15px, and that is a deliberate WCAG trade.** SC 2.5.8 wants 24px, but a 24px strip reaches past what the layout has to spare and covers the start edge of every breadcrumb link, panel body and panel button on a show page. The ceiling is `offset + 13px`, not the `+17px` the border-plus-padding suggests, because the first breadcrumb pulls its own 4px of padding back out for optical alignment. Drag-only resizing already misses SC 2.5.7 (keyboard equivalence does not close it without a pointer-operable control), so AA is unreachable either way — hence `Avo.configuration.sidebar_resizable = false` as a supported off-switch for hosts under a VPAT or public-sector obligation.

**The cookie is a trust boundary.** `avo.sidebar.width` is user-controlled and settable by any sibling subdomain. The parse returns an **Integer or `nil`, never a String**, and `valid_encoding?` comes first: Rack 3.2.6 hands back a UTF-8-tagged String with invalid bytes for `avo.sidebar.width=%C3%28`, on which `blank?`, `presence` and `Regexp#match?` all raise `ArgumentError` — a remotely-settable 500 on every page of the admin. The regex is `\A\d+\z`, not `^\d+$`, which would accept `"200\n<script>alert(1)</script>"`.

**The starting width is configurable; the bounds are not.** `Avo.configuration.sidebar_default_width` sets where the sidebar starts before anyone drags it. It rides the same pre-paint carrier as `--sidebar-width-default`, giving the lg-gated clamp a three-step fallback: the user's stored cookie, then the host's configured default, then Avo's `--spacing(64)`. Three constraints are deliberate. It is **clamped into the same [200, 480]** as the drag, so a configured width can never place the sidebar where the handle cannot drag it back or leave `aria-valuenow` outside its own declared min/max — and an unparseable value falls back to 256 rather than clamping to 200, so a typo does not silently ship a minimum-width sidebar. It applies **only at `>= lg`**, because below that the sidebar is a full-height overlay and a configured 400px would cover a phone screen. And **remove-when-default now means the *configured* default**, so a host on 400px does not accumulate cookies that merely restate their own setting. Nothing is emitted at all when the value is 256, keeping the bytes and the `script-src` surface at zero for everyone who never sets it. This crosses the plan's original "no configuration API for the bounds or the default" boundary, by request — the bounds half of that boundary stands, and `SIDEBAR_WIDTH_MIN`/`MAX` remain `private_constant`.

## Behavior

| Condition | Result |
|---|---|
| No cookie, no config | 256px, no carrier script emitted at all |
| `sidebar_default_width` set | That width at `>= lg`; still 256px below `lg`; a dragged width outranks it |
| `>= lg`, fine pointer, sidebar open | Handle revealed, drag between 200px and `min(480px, 40vw)` |
| Below `lg` | `display: none` — not resizable, not in the tab order, default width regardless of stored value |
| Coarse pointer / no hover | `display: none` |
| Sidebar closed | Handle hidden, no orphaned tab stop |
| No JS | Default width, handle stays `hidden` as rendered |
| Double-click | Resets to the configured default and *removes* the cookie |
| Escape mid-drag | Reverts to the pointerdown width — the only revert path; every other termination commits |

## Scope

Units 1-7 of a nine-unit plan. **Not included:** keyboard operation and dynamic ARIA (Unit 8), the cross-controller interlocks — hotkey suppression during a drag, focus rescue when the sidebar closes (Unit 9), and cross-tab sync (Unit 3). The handle is focusable and exposes static `aria-valuenow`/`min`/`max` (axe-core requires `aria-valuenow` on `role="separator"`), but arrow keys do nothing yet.

## Test plan

Most of this is not visible in the diff:

- **Drag specs use Ferrum's real mouse** (`page.driver.browser.mouse`), not synthetic `PointerEvent`s. Synthetic events dispatched at the element hide the pointer-capture problem above entirely.
- **The permanent invariant:** a spec asserts no interactive element's start edge falls inside the handle's strip on a show page at 1280px with a 480px sidebar. This binds every future Avo field and panel, every paid gem's show/edit UI, and every host's custom fields — it is what caught the 24px version.
- **The parse table** covers `""`, `"abc"`, `"400abc"`, `"1e9"`, `"-50"`, `"+200"`, `"2_0_0"`, `" 200 "`, `"٤٨٠"`, a 5000-digit value, and `"\xC3\x28"`, asserting the return *type* each time.
- **The Armed state** is checked with a `MutationObserver`, because a sub-3px click must leave no trace *at any point*, not merely none afterwards.
- **CSP is a documented manual run**, not a suite scenario: the dummy gates its policy on `DUMMY_CSP_REPORT_ONLY=1`, read at boot. Run `DUMMY_CSP_REPORT_ONLY=1 bundle exec rspec spec/system/avo/group_1/sidebar_resize_spec.rb` and assert no `script-src` violation — checking `script-src` matters more than `style-src`, since a blocked carrier is the failure mode.
- The `os-pc` vs `os-mac` scroll-gutter cell cannot be verified in CI; headless Linux cannot reproduce macOS overlay scrollbars.

144 examples pass across the sidebar, hotkey, request and helper specs.

## Docs

User-facing documentation: avo-hq/docs.avohq.io#577 (guide + API reference on `customization.md`/`customization-api.md`, plus unreleased upgrade notes for the label truncation and `.container-small` changes).

---

[Compound Engineering v2.59.0](https://github.com/EveryInc/compound-engineering-plugin)
🤖 Generated with Claude Opus 5 (1M context, extended thinking) via [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Touches layout, CSP-dependent pre-paint scripts, and user-controlled cookie parsing on every authenticated page; drag/handle behavior is new surface area though bounded and feature-flaggable via `sidebar_resizable`.
> 
> **Overview**
> **Desktop users can drag the main nav sidebar** between 200px and `min(480px, 40vw)` on viewports ≥ `lg`, with the choice stored in `avo.sidebar.width` and applied before paint via a nonce'd head script that sets `--sidebar-width-stored` on `<html>`. Layout derives `--sidebar-width` inside an `lg`-gated `clamp()` on the sidebar host so mobile stays at the default width.
> 
> A fixed **15px resize strip** (`sidebar_resize_controller.js`) updates width during drag (window-bound pointer events, 3px threshold before drag state, cookie write on release, double-click reset, Escape revert). **`Avo.configuration.sidebar_resizable`** and **`sidebar_default_width`** let hosts opt out or change the starting width. Sidebar link/section/group labels **truncate with `title` tooltips** so narrow sidebars stay readable.
> 
> **`.container-small`** switches from fixed `xl:w-240` to `xl:w-full xl:max-w-240` so show/edit pages do not overflow when the sidebar is wide. Server-side **`sidebar_width`** parsing (after auth) rejects malformed cookies; bounds flow to `window.Avo.configuration.sidebar`.
> 
> Per the PR scope note, **Unit 3** (Turbo/cross-tab width sync), **Unit 8** (keyboard + live ARIA), and **Unit 9** (hotkey suppression, toggle interlocks) are **not** in this diff—the handle is focusable with static separator ARIA only.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 29966329dfd04696b068a0aa6397ba606c678efa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -174,6 +174,7 @@
   @import "./css/components/field-wrapper.css";
   @import "./css/components/avatar.css";
   @import "./css/components/alert.css";
+  @import "./css/components/back_to_top.css";
   @import "./css/components/breadcrumbs.css";
   @import "./css/components/button.css";
   @import "./css/components/grid.css";
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -189,6 +189,9 @@
   @import "./css/components/filters.css";
   @import "./css/css-animations.css";
   @import "./css/sidebar.css";
+  /* After sidebar.css: the handle overlays the sidebar's end edge and its
+     focus treatment overrides the shared ring. */
+  @import "./css/components/sidebar_resize.css";
 
   html,
   body {
--- a/app/assets/stylesheets/css/components/back_to_top.css
+++ b/app/assets/stylesheets/css/components/back_to_top.css
@@ -1,0 +1,37 @@
+/* "Back to top" pill, mirroring the Furo docs theme (diataxis.fr): a fixed
+   button centered just below the navbar that reveals when the user scrolls back
+   up and hides when scrolling down. Visibility is driven by the back_to_top
+   Stimulus controller toggling `.back-to-top--visible`. */
+.back-to-top {
+  @apply fixed z-50 flex items-center gap-1 rounded-2xl bg-primary ps-2 pe-3 py-2 text-content no-underline;
+
+  inset-inline-start: 50%;
+  top: calc(var(--top-navbar-height) + 0.5rem);
+  font-size: 0.8125rem;
+  box-shadow: 0 0.2rem 0.5rem rgb(0 0 0 / 0.05), 0 0 1px 0 hsl(220 9% 46% / 0.5);
+
+  /* Hidden by default. Uses opacity/visibility rather than the `hidden`
+     attribute so appearing/disappearing can transition. */
+  opacity: 0;
+  visibility: hidden;
+  pointer-events: none;
+  transform: translateX(-50%) translateY(-0.5rem);
+  transition: opacity 0.2s ease, transform 0.2s ease, visibility 0.2s;
+
+  &:hover {
+    @apply text-content bg-secondary;
+  }
+}
+
+.back-to-top--visible {
+  opacity: 1;
+  visibility: visible;
+  pointer-events: auto;
+  transform: translateX(-50%) translateY(0);
+}
+
+.back-to-top__icon {
+  @apply size-4 shrink-0;
+
+  fill: currentColor;
+}
--- a/app/assets/stylesheets/css/components/sidebar_resize.css
+++ b/app/assets/stylesheets/css/components/sidebar_resize.css
@@ -1,0 +1,162 @@
+/* Drag handle for the desktop sidebar (R1–R3, R8, R18, R22).
+
+   A fixed overlay sibling rather than a border on either side: .avo-sidebar and
+   .main-content are separate layout contexts and neither can host a grab zone
+   that straddles the seam without eating its own clicks.
+
+   Geometry — the strip covers [offset − 2px, offset + 13px]. Only 2px bite into
+   the sidebar, leaving >= 6px of the 8px `body.os-pc` scroll gutter grabbable.
+   The other 13px stop exactly at the earliest interactive box edge in the
+   layout: .main-content's start budget is 17px (1px border + 16px padding, since
+   .container-small zeroes the inner 8px at >= xl), but the first breadcrumb
+   pulls its own 4px of padding back out for optical alignment, so anything
+   past +13px starts covering it. sidebar_resize_spec.rb holds that line.
+
+   Why not the 24px that WCAG SC 2.5.8 wants: it reaches 11px past this and
+   covers the start edge of every breadcrumb link, panel body and panel button on
+   a show page. Drag-only resizing already misses SC 2.5.7 (hence
+   Avo.configuration.sidebar_resizable as the off-switch), so AA is not reachable
+   here either way, and 15px is still comfortably above the 12px minimum this
+   feature needs.
+
+   The at-rest divider is .main-content's own `border-s`; this element draws no
+   second hairline. Its ::before seam is transparent until hover/focus/drag, and
+   the grip is ::after because role="separator" forces role="presentation" onto
+   any real child node.
+
+   `hidden` is owned exclusively by sidebar_resize_controller.js. */
+.sidebar-resize-handle {
+  @apply fixed;
+
+  z-index: 100; /* above .avo-sidebar (50), below .top-navbar (400) */
+  inset-inline-start: calc(var(--sidebar-offset-size) - 2px);
+  top: var(--top-navbar-height);
+  width: 15px;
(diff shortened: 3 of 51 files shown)


adrianthedev:
- [x] Let's also add a default size for the sidebar?

PR #4690, opened 2026-08-02
feature: add reveal field for display views

### LINKED ISSUE #2168: Add a `reveal` field which can hide some text from the user until they hit the reveal button

jacobgwilliams:
## Summary
- Adds a display-only (`index`/`show`) `reveal` field that shows a masked value with an eye toggle
- Keeps plaintext out of the initial DOM by Base64-encoding the payload and decoding only on reveal
- Wired into the Playground resource for easy manual testing

Fixes https://github.com/avo-hq/avo/issues/2168

# Checklist:

- [x] I tested the design in Light and Dark mode, and all default themes

## Manual review steps

1. Visit `/admin/resources/playgrounds` (index) or a playground show page
2. Find the **Password value** column/field — it should render as a masked password input with an eye icon
3. Confirm the plaintext (`super-secret`) is not visible in the page HTML (only a Base64 payload)
4. Click the eye icon — the value becomes visible; click again to hide
5. On edit/new forms, confirm the field is not shown as `reveal` (password field with `revealable` remains for forms)

```bash
bin/test ./spec/lib/avo/fields/reveal_field_spec.rb ./spec/features/avo/reveal_field_spec.rb ./spec/system/avo/group_1/reveal_field_spec.rb
```

### Usage

```ruby
field :api_token, as: :reveal
field :api_token, as: :reveal, mask_length: 12
```

Made with [Cursor](https://cursor.com)

### DIFF
--- a/app/components/avo/fields/common/reveal_component.html.erb
+++ b/app/components/avo/fields/common/reveal_component.html.erb
@@ -1,0 +1,33 @@
+<div
+  class="field-wrapper__input relative inline-flex items-center max-w-full"
+  data-controller="reveal-field"
+  data-reveal-field-payload-value="<%= encoded_value %>"
+  data-reveal-field-mask-value="<%= mask %>"
+  data-component="<%= component_name %>"
+>
+  <input
+    type="password"
+    readonly
+    tabindex="-1"
+    autocomplete="off"
+    spellcheck="false"
+    class="<%= class_names("w-full", "min-w-[8rem]": index?) %>"
+    data-reveal-field-target="input"
+    value="<%= mask %>"
+    aria-label="<%= @field.name %>"
+  />
+
+  <button
+    type="button"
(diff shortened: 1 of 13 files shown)


Paul-Bob:
Thanks for submitting this PR, @jacobgwilliams.

I'm closing it for now until we untangle the requirements for this field. You worked on https://github.com/avo-hq/avo/issues/2168, but there is a similar issue, https://github.com/avo-hq/avo/issues/3959, which also received a PR. They are kind of the same concept, but with different field names.

I'm closing both for now until we define and merge the requirements and consolidate the related tasks.

PR #4703, opened 2026-08-10
Render server-side dates/times in the visitor's browser time zone via config.use_browser_timezone

adrianthedev:
## Description

Adds an opt-in `config.use_browser_timezone = true` that makes every Avo request render server-side dates and times in the visitor's own time zone instead of the app's `Time.zone`.

No HTTP header carries the browser's time zone, so it has to be smuggled out of the JS runtime:

- A new `browser-timezone` Stimulus controller (attached to `<body>` only when the config is on) reads `Intl.DateTimeFormat().resolvedOptions().timeZone` and mirrors it into the `avo.browser_timezone` cookie.
- **Loop-proofing:** the controller writes the cookie and reads it back. If the read-back fails (cookies disabled), it does nothing — the user stays on the app zone forever, no reload loop. If it stuck, it sets a one-shot `avo.timezone_changed` cookie and soft-reloads once via `Turbo.visit(..., { action: "replace" })` (no full browser reload).
- `BaseApplicationController#set_browser_timezone` (around_action) validates the cookie through `ActiveSupport::TimeZone[]` — junk values fall back to the app zone without raising — and wraps the request in `Time.use_zone`.
- On the reloaded request the server deletes the one-shot cookie and shows a flash: "Dates and times are now displayed in your time zone (Europe/Bucharest)." — once per browser, ever; later visits render correctly on first paint with no reload and no alert.

The luxon-based datetime fields already convert client-side to the browser zone, so with this on, fields and server-rendered surfaces (e.g. the avo-calendar grid) agree.

## Checklist

- [x] Request specs: zone applied, alert shown once then cookie deleted, junk-cookie fallback, config off, controller attachment
- [x] Initializer template + `avo.en.yml` locale key (with inline `default:` for apps whose locale file predates the key)
- [x] `avo-admin-config` skill updated; `ws audit-skills` clean
- [x] docs.avohq.io: `4.0/customization.md` + `4.0/customization-api.md` updated (separate docs repo, uncommitted there)

---

> [!NOTE]
> **Medium Risk**
> Changes how every Avo request interprets time when enabled (default on), including cookie-driven `Time.use_zone` and a first-visit Turbo reload; validation limits abuse but display semantics shift app-wide.
> 
> **Overview**
> Adds **`config.use_browser_timezone`** (default **on**) so server-rendered dates and times follow each visitor’s local zone instead of only the app’s `config.timezone`. Set it to **false** to keep everyone on the configured app zone.
> 
> When enabled, a **`browser-timezone`** Stimulus controller on `<body>` writes the IANA zone from `Intl` into a cookie, then triggers a **one-time Turbo replace** reload (with read-back guarding against cookie-blocked reload loops). **`BaseApplicationController#set_browser_timezone`** validates that cookie via `ActiveSupport::TimeZone[]`, wraps the request in `Time.use_zone`, and shows a **one-shot flash** when the `timezone_changed` cookie is present.
> 
> Docs and templates cover the new option (`avo-admin-config`, initializer comment), **`avo.timezone_changed`** is added across locale files, **`agents.md`** documents filling missing translations via `i18n-tasks`, and the dummy app disables the feature in **test** so browser specs stay deterministic. Request specs cover zone application, flash/cookie cleanup, junk cookies, and config-off behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7ad0fb88e16403288354366ab3d60dfe51a17ee3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/agents.md
+++ b/agents.md
@@ -98,6 +98,16 @@ Guards:
 - Assert slug uniqueness within the set at boot; raise and point the author at `self.param` on conflict.
 - During a transition, also accept the legacy value so bookmarked URLs keep working (`scope.param == p || scope.name == p`); drop the fallback a version later.
 
+## Translations
+
+When you touch translations (any key in `lib/generators/avo/templates/locales/*.yml`), fill in the other languages before declaring done:
+
+```sh
+bundle exec i18n-tasks translate-missing --backend=openai
+```
+
+It needs `OPENAI_API_KEY` in the environment; if it is not set, ask the developer to run the command instead of skipping it.
+
 ## Logging in (development & testing)
 
 The dummy app at `spec/dummy` uses Devise for authentication. Avo is mounted at `/admin` and is wrapped in an `authenticate :user, ->(user) { user.is_admin? }` block, so you need an admin user to reach it.
--- a/app/controllers/avo/base_application_controller.rb
+++ b/app/controllers/avo/base_application_controller.rb
@@ -18,6 +18,7 @@ module Avo
     before_action :decode_params
     around_action :set_avo_locale
     around_action :set_force_locale, if: -> { params[:force_locale].present? }
+    around_action :set_browser_timezone, if: -> { Avo.configuration.use_browser_timezone }
     before_action :init_app
     before_action :set_active_storage_current_host
     before_action :set_resource_name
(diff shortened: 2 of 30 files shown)


4ndv:
I've just spent couple of hours debugging an issue caused by usage of `Time.use_zone` in this PR

Overriding application's configured timezone via `Time.use_zone` is NOT the same as "Render server-side dates/times in the visitor's browser time zone"

adrianthedev:
hey @4ndv.
Sorry about that. I'm sure it wasn't pretty.

What do you mean? What did it break?

4ndv:
A call to `Time.zone.local` inside of action handler, which was expected to be in application's configured timezone, but instead was in my timezone, which differs by 5 hours from expected

adrianthedev:
Yes. I see how that's a big issue.
It's in our court and definitely a bug. 
Sorry about that.  
We'll issue a fix right away.

4ndv:
The problem is not really with the change itself, but a combination of:

- It wasn't properly explained in release changelog
- It was made in "patch" release, breaking semver
- It's enabled by default

Issue #4712, opened 2026-08-11
Nested association create returns 500 after record is committed


nrgbistro:
## Description

Creating a record through a parent resource's association can save the child successfully and then return a 500 response while Avo re-hydrates the resource for the `:new` view. In a Turbo form this leaves the submit button spinning indefinitely, even though the database transaction has already committed.

This is especially risky because an operator may retry and create duplicate records.

## Steps to reproduce

1. On Avo v3.32.2, define a normal parent/child relationship:

   ```ruby
   class Parent < ApplicationRecord
     has_many :children
   end
   
   class Child < ApplicationRecord
     belongs_to :parent
   end
   ```
2. Expose both sides in Avo:

   ```ruby
   # ParentResource
   field :children, as: :has_many
   
   # ChildResource
   field :parent, as: :belongs_to
   ```
3. Open a parent record in Avo, use the `children` association to create a new child, fill the required fields, and submit.
4. Observe that the `POST` returns 500 and the UI remains in its loading state, but the child row exists in the database and is associated with the parent.

The failure is specific to the relationship-create path. Creating the same child directly, or editing an existing child, succeeds.

## Expected behavior

After the child is committed, Avo should complete the success response and clear the Turbo loading state.

## Actual behavior

The child is committed, then the request raises:

```text
NoMethodError: undefined method `find_record` for nil
lib/avo/resources/base.rb: hydrate_model_with_default_values
```

## Root cause

`Avo::Resources::Base#hydrate_model_with_default_values` enters the relationship-default branch based on the resource's cached `@params[:via_relation]`, but resolves the resource with the current `params[:via_relation_class]`:

```ruby
resource = Avo.resource_manager.get_resource_by_model_class params[:via_relation_class]
record = resource.find_record @params[:via_record_id], params: params
```

During the post-save `:new` hydration in `Avo::BaseController#create`, the current params can lack `via_relation_class`. The lookup therefore returns `nil`, and `resource.find_record` raises after `save_record` has already committed the row.

The v3.32.2 `3.x` source still dereferences both `resource` and `record` without presence checks.

## Related v4 fix / backport request

[...]

PR #4713, opened 2026-08-11
fix: guard relationship default hydration

nrgbistro:
# Description

Nested association creates can commit the child record and then fail during the post-save `:new` hydration pass. The resource retains cached `via_relation` params, while the current request can lack `via_relation_class`; `get_resource_by_model_class` then returns `nil` and Avo dereferences it with `find_record`.

This backports the defensive relationship-hydration behavior from #4444 to `3.x`:

- require a relationship record ID before reflecting or resolving a parent
- guard missing regular and polymorphic Avo resources
- guard custom resource finders that return no record
- preserve an already assigned parent when the relationship default cannot be resolved

Valid relationship creates keep the same resource lookup and parent query. Missing IDs now skip the lookup entirely; the regression suite asserts that no parent SQL query is made in that case.

Fixes https://github.com/avo-hq/avo/issues/4712

# Checklist

- [x] Comments are not required; the guards follow the existing hydration branches
- [x] Documentation changes are not required for this v3 bug fix

## Screenshots & recording

Not applicable. This fixes a server-side post-save exception and Turbo loading state.

## Manual review steps

1. Run `bin/test ./spec/features/avo/resource_hydration_spec.rb ./spec/features/avo/default_field_spec.rb ./spec/features/avo/create_through_record_spec.rb ./spec/features/avo/has_many_field_spec.rb ./spec/features/avo/belongs_to_polymorphic_spec.rb`.
2. Verify all 37 examples pass, including regular, missing-ID, and missing-polymorphic-resource hydration cases.
3. Run `bundle exec standardrb lib/avo/resources/base.rb spec/features/avo/resource_hydration_spec.rb`.

### DIFF
--- a/app/components/avo/fields/belongs_to_field/edit_component.rb
+++ b/app/components/avo/fields/belongs_to_field/edit_component.rb
@@ -62,6 +62,7 @@ class Avo::Fields::BelongsToField::EditComponent < Avo::Fields::EditComponent
 
     helpers.new_resource_path(**{
       via_relation: @field.id.to_s,
+      via_relation_class: resource.model_class.to_s,
       resource: target_resource || @field.target_resource,
       via_record_id: resource.record.persisted? ? resource.record.to_param : nil,
       via_belongs_to_resource_class: resource.class.name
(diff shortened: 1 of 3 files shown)


Paul-Bob (approved):
Thanks for the contribution @nrgbistro 🙏

I pushed one tweak on top

Merging soon and releasing later today.

PR #4747, opened 2026-08-20
Hover flyout for sidebar items whose sub-items aren't listed inline

adrianthedev:
The core half of a sidebar hover flyout: a stylesheet block and the Stimulus controller it reaches for. Nothing in core renders it yet — `avo-menu`'s `ItemSwitcherComponent` is the consumer, and that change ships separately.

## Why

A sidebar parent with sub-items lists them inline only while that branch is open. Everywhere else they are unreachable without navigating to the parent first. Hovering the parent now shows the same list in a panel beside it.

## The panel is a native popover, and it has to be

The sidebar body is its own scroll container, so an absolutely positioned panel is clipped by it — the top layer is the only place this can be drawn without a portal. `popover` gets it there, and CSS anchor positioning pins it to the item through a per-item `anchor-name` the consuming component writes inline, with `position-try-fallbacks` for a sidebar pinned to either edge.

`transition-behavior: allow-discrete` mirrors `.popover-menu__panel`: the panel is `display: none` while closed, so without it the leave transition never plays.

## Details worth knowing

- **Opens on focus as well as hover**, so the list is reachable by keyboard.
- **The close is delayed ~120ms.** The panel sits a few pixels off the item, so the pointer is over neither element for a frame on the way across; without the delay the panel closes under the cursor.
- `showPopover()` / `hidePopover()` are guarded by `:popover-open`, so a repeated enter or a leave that races the timeout can't throw.

## Files

| File | What |
| --- | --- |
| `app/assets/stylesheets/css/sidebar.css` | The `.sidebar-flyout` block |
| `app/javascript/js/controllers/sidebar_flyout_controller.js` | Show/hide, with the crossing delay |
| `app/javascript/js/controllers.js` | Registers it |
| `app/assets/stylesheets/css/variables.css`, `.../components/ui/card.css` | Tokens the panel reads |

---

> [!NOTE]
> **Low Risk**
> Additive CSS and a small Stimulus controller with no current call sites in this diff; behavior only activates once the menu component lands.
> 
> **Overview**
> Adds the **core building blocks** for a sidebar hover flyout when a parent’s children aren’t shown inline (branch collapsed). Nothing in core wires this up yet; **`avo-menu`’s `ItemSwitcherComponent`** is expected to consume it in a follow-up.
> 
> **Styles (`sidebar.css`):** New `.sidebar-flyout` / `.sidebar-flyout__panel` use a **native `popover`** in the top layer so the panel isn’t clipped by the sidebar’s `overflow-y: auto`. **CSS anchor positioning** (`position-area`, `position-try-fallbacks`) pins the panel beside the item; enter/leave opacity transitions mirror `.popover-menu__panel` with `allow-discrete` on `display`/`overlay`. Flyout-specific overrides strip sub-item indent and connector pseudo-elements inside the panel.
> 
> **JS:** New **`sidebar-flyout`** Stimulus controller opens the panel on hover/focus via `showPopover()` / `hidePopover()` (guarded with `:popover-open`), with a **~120ms close delay** so the cursor can cross the gap between item and panel without flicker.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b5fe594555859c23c5b84ae0f55cd3778ca858f8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/app/assets/stylesheets/css/sidebar.css
+++ b/app/assets/stylesheets/css/sidebar.css
@@ -518,3 +518,65 @@
 .sidebar-status:hover .sidebar-status__hint {
   @apply block;
 }
+
+/* ================================================ */
+/* Sidebar Flyout — ItemSwitcherComponent           */
+/* BEM: .sidebar-flyout (block)                     */
+/* ================================================ */
+
+/* A parent item's sub-items, shown on hover for as long as they aren't already
+   listed inline (that is, whenever the parent's own page and its children are
+   all closed). The panel is a native popover: the sidebar body scrolls, so an
+   absolutely positioned panel would be clipped by it — the top layer is the
+   only place this can be drawn. Position comes from CSS anchor positioning,
+   with the per-item `anchor-name` written inline by the component. */
+.sidebar-flyout {
+  @apply w-full;
+}
+
+.sidebar-flyout__panel {
+  @apply m-0 p-1 rounded-lg min-w-48;
+  background-color: var(--sidebar-background);
+  border: 1px solid var(--sidebar-border);
+  box-shadow: var(--box-shadow-card);
+  position-area: inline-end span-block-end;
+  position-try-fallbacks: flip-inline, flip-block, flip-block flip-inline;
+  margin-inline-start: 0.25rem;
+
+  /* Mirrors .popover-menu__panel: the panel is display:none while closed, so
+     `allow-discrete` keeps it in the top layer through the leave transition. */
+  opacity: 0;
+  transition:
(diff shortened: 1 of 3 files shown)


adrianthedev:
Duplicate — this shipped as #4744 (995191eaf) while I was preparing it. The branch here was the pre-Prettier copy of the same work, so merging it would also revert #4745's reformatting of variables.css and card.css. Closing.

PR #4752, opened 2026-08-25
fix: scope `use_browser_timezone` to rendering only

adrianthedev:
# Description

Follow-up to #4703, from [@4ndv's report](https://github.com/avo-hq/avo/pull/4703#issuecomment-5414482140) that they lost a couple of hours to it.

`use_browser_timezone` applied the visitor's zone with an `around_action`, so `Time.zone` was swapped for the **entire request**. That is a much bigger blast radius than "render dates and times in the visitor's zone": everything running inside the action saw the swapped zone, including app code Avo knows nothing about.

- `Time.zone.local` in an action's `handle` (the reported case, off by 5 hours)
- `Date.current` / `Time.current` in a scope, filter, or query object
- ActiveRecord's time-zone-aware attribute casting on write
- model callbacks

Nothing raises. The value is just silently wrong by the offset between the app zone and whichever browser happened to make the request, which is why it reads as a data bug rather than a config change.

The option is documented as a display setting, so this applies it as one. The zone now wraps `render_to_body`, the common inner path for `render` (explicit and implicit, HTML and Turbo Stream) and `render_to_string`. Templates and view components still render in the visitor's zone; everything else in the request keeps `config.timezone`, so `Time.zone` means the same thing inside Avo as it does everywhere else in Rails.

The one-shot `timezone_changed` flash moves from the `around_action` to a `before_action`, registered after `_authenticate!` so the user-controlled cookie is not parsed for an unauthenticated request (matching the convention already used for `avo.sidebar.width`).

No config change: the option stays on by default. If you would rather it were opt-in as well, say so and I will add that.

Docs side: avo-hq/docs.avohq.io#672

# Checklist:

- [x] I have made corresponding changes to the documentation
- [ ] I tested the design in Light and Dark mode, and all default themes 

## Screenshots & recording

No visual change.

## Manual review steps

1. In an app with `config.timezone` set to something other than your browser's zone, add an action whose `handle` does `Rails.logger.info Time.zone.name` (or builds a timestamp with `Time.zone.local`).
2. Run it from Avo. Before this change the zone logged is your browser's; after it is `config.timezone`.
3. Confirm display is unaffected: server-rendered dates and times still show in your browser's zone.

Automated:

```
bin/rspec spec/requests/avo/browser_timezone_request_spec.rb   # 8 examples, 0 failures
bin/rspec spec/requests                                        # 104 examples, 0 failures
```

The new `leaves the action on the app's configured zone` example fails against the previous implementation and passes against this one (verified by reverting the controller and re-running).

### DIFF
--- a/app/controllers/avo/base_application_controller.rb
+++ b/app/controllers/avo/base_application_controller.rb
@@ -18,7 +18,6 @@ module Avo
     before_action :decode_params
     around_action :set_avo_locale
     around_action :set_force_locale, if: -> { params[:force_locale].present? }
-    around_action :set_browser_timezone, if: -> { Avo.configuration.use_browser_timezone }
     before_action :init_app
     before_action :set_active_storage_current_host
     before_action :set_resource_name
@@ -28,6 +27,7 @@ module Avo
     before_action :set_view
     before_action :set_sidebar_open
     before_action :set_sidebar_width
+    before_action :announce_browser_timezone_change, if: -> { Avo.configuration.use_browser_timezone }
 
     rescue_from Avo::NotAuthorizedError, with: :render_unauthorized
     rescue_from ActiveRecord::RecordInvalid, with: :exception_logger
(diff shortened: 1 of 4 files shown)


4ndv:
I'm not sure if it's actually needs to be removed, because setting the timezone this way can be useful. My issue was that it was released as a breaking change in patch release and enabled by default, not the change itself per se

PR #4753, opened 2026-08-26
Document the container CSS class rename in avo-admin-config

adrianthedev:
# Description

**This PR has been narrowed.** It opened as "Sync shipped skills with docs through 70f1de9 (2026-08-20)", but that same window was reconciled independently by #4760, and #4763 then carried the marker on to `ee0277f` (2026-08-31). Rather than leave a conflicted, superseded PR open, main has been merged in and everything already covered there was dropped.

What survives is the **one gotcha no sync picked up**. The branch now differs from main by a single line.

## The change

When `container_width` moved to Tailwind's scale in #4749, the config values were deprecated *and aliased* — `:large`/`:small` keep working and log a warning. The **CSS classes Avo emits were renamed at the same time and are not aliased**:

| Before                  | After              |
| ----------------------- | ------------------ |
| `.container-large`      | `.container-lg`    |
| `.container-small`      | `.container-md`    |
| `.container-full-width` | `.container-full`  |
| —                       | `.container-sm` (new) |

So an app that styles Avo's container in its own CSS silently stops matching, and nothing warns about it, because the deprecation only covers the config value. `avo-admin-config` documents the config-value rename but not this, which is the half that fails quietly.

Re-verified on current main, not just from the docs:

- `app/assets/stylesheets/css/layout.css:153-176` defines `.container-lg`, `.container-md`, `.container-sm`, `.container-full`.
- `git grep container-large|container-small|container-full-width -- app/ lib/` returns nothing, so the old names are genuinely gone rather than aliased somewhere.

## What was dropped, and why

| Originally in this PR | Why it's gone |
| --- | --- |
| `docs-index.json` → `70f1de9` (2026-08-20) | Would **regress** the marker — main is already at `ee0277f` (2026-08-31) via #4763. Resolved in main's favour. |
| `avo-index-views` map dark-mode note | main carries this as a fuller standalone gotcha plus a cross-reference in the `mapkick_options` bullet. Main's version is better; dropped. |
| `avo-navigation-search` `link_to`-block correction | main carries the same correction. Dropped. |

Merge conflicts in those three files were all resolved in main's favour.

## Worth flagging separately

Two independent runs produced the same sync for the 2026-08-20 window (this PR and #4760). That's duplicated effort rather than a code problem, but whoever owns the skills-sync schedule may want to look at why both fired.

## Verification

```bash
grep -rnE '\[!code|:::' lib/avo/skills/      # plus the JSX-ish docs tags; returns nothing
bundle exec rspec spec/lib/avo/skills        # => 167 examples, 0 failures
git diff origin/main -- lib/avo/skills/      # => 1 file changed, 1 insertion(+)
```

The drift marker is deliberately **not** touched here — docs have since moved to `6e17419` (2026-09-09), but re-indexing is the scheduled sync's job, not this PR's.

https://claude.ai/code/session_01DfAz3nmJREY42UTdMFJMiB


adrianthedev:
`tailwindcss_integration_test` is failing on this PR's head, and it is **not this PR's failure** — this branch changes one line of Markdown in a `SKILL.md`.

**It is red on `main` too**, including this PR's exact base commit:

| main commit | Date | Result |
| --- | --- | --- |
| `f9c8ba2` Bump avo to v4.2.1 | 2026-09-05 | ✅ last green |
| `4abddb2` #4778 | 2026-09-07 | ❌ |
| `c873c8d` #4777 | 2026-09-07 | ❌ |
| `ba8270b` #4779 | 2026-09-07 | ❌ |
| `9365c2c` #4783 | 2026-09-09 | ❌ |
| `c6989d2` Bump avo to v4.2.2 | 2026-09-09 | ❌ |
| `d515470` #4785 | 2026-09-09 | ❌ |
| `372fbf2` Bump avo to v4.2.3 | 2026-09-09 | ❌ |
| `5272140` #4786 | 2026-09-09 | ❌ |
| **`1cd4974` #4787 — this PR's base** | **2026-09-09** | **❌** |

Nine consecutive failures over four days, so it is deterministic rather than flaky. I have **not** spent a re-run on it: re-running a failure that reproduces on every main push would tell us nothing.

### Root cause

```
ArgumentError: unknown keyword: quirks_mode
  json-3.0.2/lib/json/common.rb:382:in 'JSON.generate'
  activesupport-7.1.2/.../json/encoding.rb:92:in 'JSONGemEncoder#stringify'
  propshaft-0.8.0/lib/propshaft/processor.rb:35:in 'write_manifest'
Tasks: TOP => assets:precompile
```

`json` 3.0.2 removed the `quirks_mode` keyword that ActiveSupport 7.1.2's JSON encoder still passes. Propshaft's `write_manifest` calls `to_json`, so `assets:precompile` aborts before the Tailwind assertions ever run.

Nothing in Avo pins it. The job checks out the fixture repo `avo-hq/avo_tailwindcss_test_repo` (ref `avo-1147/tailwindcss-integration`), which is on Rails 7.1.2, then runs `bundle config set frozen false` + `bundle add avo --git ...` + `bundle install`. That re-resolution lets `json` float to its newest release, which is why the break arrived on its own with no commit to blame.

### Proposed patch — not applied here

I'm deliberately not widening a one-line docs PR with a CI fix, and the better fix lives in a repo I can't reach from this session. Either of these should work:

1. **In `avo_tailwindcss_test_repo`** (preferable — fixes it at the source): bump off Rails 7.1.2 to a 7.1.x that is compatible with json 3, or add `gem "json", "< 3"` to the fixture Gemfile.
2. **In this repo's workflow**, if you'd rather not touch the fixture: constrain the gem during the `Add Avo` step, before `bundle install` —
   ```yaml
   bundle add "json" --version "< 3"
   ```

Option 1 is the real fix; option 2 is a one-line unblock that keeps the integration test meaningful in the meantime.

No fix PR exists for this yet as far as I can find, so there is nothing for me to port into this branch. Flagging it here because it silently reds every PR and every main push until someone picks it up.

---

PR #4758, opened 2026-08-29
Skills: static maps are Mapbox-only on the preview view too, not just with `static: true`

adrianthedev:
# Description

**Scope changed since this PR was opened — please re-read.** It was opened as the docs sync through `70f1de9`, then closed. While it was closed, #4760 landed that same sync, with fuller wording than this branch had on all three overlapping edits. This PR was then reopened, which left it conflicting with `main`.

I merged `main` in and resolved **every conflict in favour of `main`** — took its `avo-admin-config`, `avo-index-views` and `avo-navigation-search` wholesale rather than reinstating the shorter phrasings from this branch, so nothing #4760 added is reverted. `docs-index.json` was already identical on both sides.

What is left is the **one fact #4760 did not carry**, and the whole diff is now a single line.

## The change

`avo-fields` tells an agent that `:location` / `:area` need no Mapbox token, and scopes the Mapbox-only static-map requirement to `static: true`. That requirement is broader: **the preview view always renders a static map**, whatever the field is configured with.

```ruby
# lib/avo/fields/location_field.rb:22-24
def static_map?(params)
  params[:action] == "preview" || @args[:static]
end
```

So an app with a plain `field :coordinates, as: :location` and no `MAPBOX_ACCESS_TOKEN` gets working interactive maps everywhere and a blank one on preview — and the skill as written says nothing that would lead an agent to expect it. The line now names both cases.

## Why it was not in #4760

The drift checker never flagged it. `avo-fields` cites the field-type catalog as a *pattern* URL — `.../fields/text.md`, "swap `text` for any type" — so `docs_drift.rb` records only `text.md` as a citation, and reported `fields/area.md` and `fields/location.md` as **uncited by any skill** even though `avo-fields` restates their content. Both pages changed in the synced range (avo-hq/docs.avohq.io#667 added exactly this preview caveat). I found it by reading the diff rather than from the tool's output.

That gap is worth its own follow-up — either the checker resolves pattern citations, or `avo-fields` cites the per-type pages it actually paraphrases. Not attempted here, to keep this diff at one line.

## Also still open from the original sync run

Carried over so they are not lost with the rest of the description:

- **`def chip` / record chips has no owning core skill.** `resources-api.md` now lists it in the resource-options table, but the DSL is `avo-ai`'s and documented as alpha. It does not exist in core — `grep` for `def chip` across `lib/` and `app/` finds nothing — so it was not added to `avo-resources`. Belongs in the `avo-ai` gem's own skill once the DSL settles.
- **Add-on-owned pages that drifted, for their gems' skills:** `ai.md`, `ai-agents-and-tools.md`, `ai-what-you-can-ask.md` (chat context sources, Media Library files and conversations as context, record chips, schema-first answering, finding an action without being told its resource) → **avo-ai**. `forms-and-pages.md`, `forms-and-pages-api.md` (`self.container_width` on pages and forms moving to `:lg` / `:md` / `:sm`) → **avo-forms**.
- **`link_to` taking a block for sub-items** is menu-editor surface and belongs in **avo-menu**'s own skill. Core only carries the `icon:` gotcha, which #4760 corrected.
- **`docs_drift.rb` aborts under a non-UTF-8 default external encoding.** On a `LANG=C` / US-ASCII runner the citation scan at `lib/avo/skills/bin/docs_drift.rb:75` raises `ArgumentError: invalid byte sequence in US-ASCII`, because the skills contain non-ASCII em dashes. Worked around by running with `LANG=C.UTF-8`; an explicit `encoding: "UTF-8"` on that `File.read` would fix it properly.

# Checklist:

- [x] I have made corresponding changes to the [documentation](https://github.com/avo-hq/avodocs) — the docs already say this (avo-hq/docs.avohq.io#667); this PR brings the shipped skill in line with them
- [x] I have added tests that prove my fix is effective or that my feature works — content-only change; covered by the existing `spec/lib/avo/skills` suite
- [ ] I tested the design in Light and Dark mode, and all default themes — n/a, no UI change

## Screenshots & recording

n/a — one line of Markdown.

## Manual review steps

1. Confirm the claim against the source — `static_map?` is true for the preview action regardless of the `static:` arg:

   ```bash
   sed -n '22,24p' lib/avo/fields/location_field.rb
   ```

1. Confirm nothing from #4760 is reverted — the only diff against `main` is the one `avo-fields` line:

   ```bash
   git diff origin/main -- lib/avo/skills/
   ```

1. The shipped-skills suite still passes:

   ```bash
   bundle exec rspec spec/lib/avo/skills
   # => 167 examples, 0 failures
   ```

### DIFF
--- a/lib/avo/skills/avo-admin-config/SKILL.md
+++ b/lib/avo/skills/avo-admin-config/SKILL.md
@@ -29,6 +29,7 @@ A handful of these have **resource-level equivalents** — set them globally her
 - Installing, mounting (`root_path`, `prefix_path`), and the license key → **avo-setup**
 - Logos, favicons, colors, theming (`appearance`) → **avo-branding-appearance**
 - Menus, global search, breadcrumbs, keyboard shortcuts (`main_menu`, `global_search`, `set_initial_breadcrumbs`, `hotkeys`) → **avo-navigation-search**
+- Map provider and the light/dark map style pair (`map_view`) → **avo-index-views**
 - Caching internals, cache store, N+1 tuning → **avo-performance**
 - Authorization client and policy wiring → **avo-authorization**
 
--- a/lib/avo/skills/avo-admin-config/SKILL.md
+++ b/lib/avo/skills/avo-admin-config/SKILL.md
@@ -29,9 +29,9 @@ A handful of these have **resource-level equivalents** — set them globally her
 - Installing, mounting (`root_path`, `prefix_path`), and the license key → **avo-setup**
 - Logos, favicons, colors, theming (`appearance`) → **avo-branding-appearance**
 - Menus, global search, breadcrumbs, keyboard shortcuts (`main_menu`, `global_search`, `set_initial_breadcrumbs`, `hotkeys`) → **avo-navigation-search**
-- Map provider and the light/dark map style pair (`map_view`) → **avo-index-views**
 - Caching internals, cache store, N+1 tuning → **avo-performance**
 - Authorization client and policy wiring → **avo-authorization**
+- Map styles / providers (`map_view`) — the map view and the `location`/`area` fields share one setting → **avo-index-views**
 
 ## Docs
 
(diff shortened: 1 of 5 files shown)


adrianthedev:
Heads up that this PR is not what it was when it was opened — the title and description are rewritten, and the diff is now **one line**.

While it was closed, #4760 landed the same docs sync, with fuller wording than this branch had on all three overlapping edits. Reopening left this conflicting with `main`. I merged `main` in and resolved **every conflict in favour of `main`** — took its `avo-admin-config`, `avo-index-views` and `avo-navigation-search` wholesale, so nothing #4760 added is reverted, and `docs-index.json` was already identical on both sides.

What survives is the one fact #4760 didn't carry: `avo-fields` scopes the Mapbox-only static-map requirement to `static: true`, but `static_map?` is also true for the preview action (`lib/avo/fields/location_field.rb:22-24`), so a plain `field :coordinates, as: :location` with no `MAPBOX_ACCESS_TOKEN` renders a blank map on preview and the skill gives no hint of it.

Happy to close this instead if you'd rather fold that line into another change — the substance is four words in one bullet.

---

PR #4761, opened 2026-08-31
fix(buttons): save in Safari when the click lands on the button text

### LINKED ISSUE #4741: Save button doesn't work when clicking on the text portion in Safari
### Describe the bug
***This bug report is directly related to this closed issue: [https://github.com/avo-hq/avo/issues/4625](https://github.com/avo-hq/avo/issues/4625) Reference that issue for additional details.***

I have determined a precise way to reproduce the issue linked above with 100% consistency in my testing. The ```Save``` button works correctly when clicking on the button surface ***around*** the actual ```Save``` text. However, if you click on the actual ```Save``` text on the button, the request fails to be sent. This appears to be due to the action being triggered on the ```<span>``` element, which is immediately deleted and replaced by the loading spinner, interrupting the request before it can be sent.

#### In detail:

Avo wraps the Save label in a child element — the button isn't <button>Save</button>, it's <button><svg/><span class="button__label">Save</span><kbd>⌘⏎</kbd></button>. So when you click the word "Save", the click event's target is that inner <span>, not the button.

Avo deletes that span during the click. Its loading-button controller runs button.innerHTML = spinner synchronously inside the click handler to swap the label for a loading spinner. That destroys the very element you clicked, while the click event is still being dispatched.

WebKit refuses to submit when its click target vanishes mid-dispatch. It abandons the submit button's activation behaviour entirely: no submit event, no HTTP request, no validation error, no flash. Chrome and Firefox delete the same element and submit anyway, which is why they work fine.

### Steps to Reproduce
- Use Safari (reproduced on Safari 26.5.2 / macOS Tahoe 26.5.2).
- Open any resource → Edit.
- Change a field, click Save ***on the button text*** with the mouse.
- The spinner will hang, no request will be sent, the the change is not saved.

### Expected behaviour & Actual behaviour
Expected: The edited record is saved and you are redirected to the resource page.
Actual: The loading spinner appears and hangs forever.

### System configuration
**Avo version**: 4.0.1

**Rails version**: 8.1.3.1

**Ruby version**: 3.4.8

**License type**:

 - [ ] Community
 - [ ] Pro
 - [X] Advanced

 -  #[X] No

https://github.com/user-attachments/assets/5e725a44-c155-4bb8-b519-1a4a48602696

As you can see, clicking on the edge of the button ***not on the text*** works perfectly, while clicking ***on the "Save" text*** fails.

#### Impact

#### Urgency

 - [X] Medium urgency (I found a workaround, but I'd love to have it fixed.)

Paul-Bob:
# Description

Clicking the word "Save" in Safari did nothing. The record was not saved, no request went out, and the spinner hung forever. Clicking the button's padding a few pixels away worked fine. Save now works wherever you click it.

Avo wraps every button label in a `<span class="button__label">`, so a click on the text targets that span, not the button. The loading controller replaced the button's `innerHTML` with a spinner synchronously inside its own click handler, destroying that span while the click was still dispatching. WebKit responds by abandoning the submit button's activation behaviour entirely: no submit event, no request, no validation error, no flash. Chrome and Firefox remove the same node and submit anyway, which is why this was Safari only and why CI never caught it.

The fix removes the destruction rather than routing around it. The spinner is now overlaid on top of contents that stay in the DOM, using the `.button__spinner` rule that already existed in the stylesheet and that nothing had ever emitted. Because the contents stay put, the inline width/height pinning and the `data-original-content` round trip, which existed only to survive that swap, are gone as well.

One consequence worth a reviewer's eye: the spinner's DOM shape changed. It used to be a direct child of the button, and is now wrapped in `.button__spinner`. A host app with custom CSS targeting the old shape, or a view override reading `data-original-content`, would notice.

Related: AVO-1751

Fixes #4741

# Checklist:

- [ ] I have made corresponding changes to the [documentation](https://github.com/avo-hq/avodocs) (no page covers the button loading state, and nothing user-configurable changed)
- [ ] I tested the design in Light and Dark mode, and all default themes (not verified: the change adds no colours and only toggles `visibility`, but a themed pass is still worth a look)

## Screenshots & recording

No recording attached. The loading state was measured directly in a headless browser instead, on a store edit page, where the confirmation dialog holds the page still long enough to inspect:

| Property | Idle | Loading |
| --- | --- | --- |
| Button box | 84x32 | 84x32 |
| `.button__spinner` count | 0 | 1 |
| `.button__label` present in DOM | yes | yes |
| Label `visibility` | `visible` | `hidden` |
| `aria-busy` | absent | `true` |

The equal box sizes are the interesting row: they are why removing the inline width/height pinning is safe. A document level click listener also confirmed the clicked node was still attached when the dispatch finished, which is the property Safari actually enforces.

## Manual review steps

1. Open any resource, then Edit, **in Safari**.
2. Change a field and click directly on the word "Save", not the padding around it.
3. The record saves and you land on the record page. Before this change the spinner hung and nothing was sent.
4. On a resource with `confirm_on_save` (`Store` in the dummy app), click Save and then dismiss the dialog. The spinner clears, the label returns, and the button stays enabled.

**Please run step 2 in Safari before merging.** The suite is Chrome only through Cuprite, and this change was developed on Linux where WebKit is not available, so the Safari behaviour itself is the one claim here that has not been executed. What the new spec pins instead is the browser agnostic invariant Safari depends on: the clicked element is still attached when the click finishes dispatching. That assertion fails against the old controller in Chrome, so it is a real regression guard rather than a restatement of the fix.

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/app/assets/stylesheets/css/components/button.css
+++ b/app/assets/stylesheets/css/components/button.css
@@ -470,6 +470,14 @@
     @apply absolute inset-0 flex items-center justify-center;
   }
 
+  /* `visibility`, not the `hidden` attribute: the content has to keep occupying space so the button doesn't collapse
+     around the spinner while it loads. */
+  .button--loading-active .button__label,
+  .button--loading-active .button__icon,
+  .button--loading-active .hotkey-badge {
+    @apply invisible;
+  }
+
   .hotkey-badge {
     @apply -mt-0.5;
   }
--- a/app/javascript/js/controllers/loading_button_controller.js
+++ b/app/javascript/js/controllers/loading_button_controller.js
@@ -2,7 +2,7 @@
 import { Controller } from '@hotwired/stimulus'
 
 export default class extends Controller {
-  spinnerMarkup = '<span class="loading-spinner" aria-hidden="true"></span>'
+  spinnerMarkup = '<span class="button__spinner"><span class="loading-spinner" aria-hidden="true"></span></span>'
 
   static values = {
     confirmationMessage: String,
@@ -10,7 +10,6 @@ export default class extends Controller {
   }
 
   connect() {
-    this.button.setAttribute('data-original-content', this.button.innerHTML)
     this.dialog = document.getElementById('turbo-confirm')
     this.dialogCloseHandler = this.handleDialogClose.bind(this)
 
@@ -29,13 +28,20 @@ export default class extends Controller {
     return this.context.scope.element
   }
 
+  get spinner() {
+    return this.button.querySelector('.button__spinner')
+  }
+
   applyLoader() {
     const { button } = this
 
-    button.style.width = `${button.getBoundingClientRect().width}px`
-    button.style.height = `${button.getBoundingClientRect().height}px`
-    button.innerHTML = this.spinnerMarkup
-    button.classList.add('justify-center')
+    if (this.spinner) return
+
+    // Overlay the spinner, never replace the button's contents: WebKit abandons a submit button's activation behaviour
+    // when the element the user clicked is removed while the click is still dispatching (AVO-1751).
+    button.insertAdjacentHTML('afterbegin', this.spinnerMarkup)
+    button.classList.add('button--loading-active')
+    button.setAttribute('aria-busy', 'true')
   }
 
   handleDialogClose() {
@@ -47,7 +53,9 @@ export default class extends Controller {
   resetButton() {
     const { button } = this
 
-    button.innerHTML = button.getAttribute('data-original-content')
+    this.spinner?.remove()
+    button.classList.remove('button--loading-active')
+    button.removeAttribute('aria-busy')
     button.removeAttribute('disabled')
   }
 }
--- a/docs/plans/2026-08-31-1357-fix-safari-save-button-click-plan.md
+++ b/docs/plans/2026-08-31-1357-fix-safari-save-button-click-plan.md
@@ -31,20 +31,20 @@ The loading-button controller replaces the button's inner HTML with a spinner wh
 
 Avo wraps every button label in `<span class="button__label">` (`app/components/avo/button_component.rb`, `render_content`). Clicking the word "Save" makes that span the click event's target. The button carries `data-action="click->loading-button#attemptSubmit"`, and `applyLoader` runs `button.innerHTML = this.spinnerMarkup` synchronously inside that handler (`app/javascript/js/controllers/loading_button_controller.js:35`).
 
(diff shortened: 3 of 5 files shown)


Paul-Bob:
## Residual Review Findings

Automated run report. Everything below was found during this PR's own review passes and was **not** applied here, so it is recorded rather than lost.

### Filed

- **[P2] Loading button loses its accessible name while `aria-busy` is set** (`app/assets/stylesheets/css/components/button.css`, `.button--loading-active`). `visibility: hidden` drops the label out of the accessibility tree, so the button announces that it is busy without naming what is busy. Not a regression (the old code also left the button nameless), but `aria-busy` makes it worth closing. Two remedies with different costs, so it is a real fork rather than a mechanical fix. Filed as [AVO-1766](https://linear.app/avo-hq/issue/AVO-1766/loading-button-loses-its-accessible-name-while-aria-busy-is-set).

### Not filed, recorded here

- **[P3] The plan carries no Alternatives section.** `.button__icon` already sets `pointer-events: none`, which is exactly why clicking the icon never reproduced this bug, and the same one-line treatment on `.button__label` and `.hotkey-badge` would also have fixed Safari. The overlay was chosen because it covers every child including ones added later and removes the destructive swap instead of routing around it, but that reasoning is not written down anywhere a reviewer would find it. Recording it here instead.

### Investigated and cleared

- **Turbo cache restore.** Review raised that a cached snapshot could restore a button stuck mid-load, since nothing clears the loading state on `connect()`. Tested directly: save, then browser-back. The restored button came back with 0 spinners, no `aria-busy`, no state class, and a visible label. Does not reproduce, so no ticket.

### Environment notes for whoever runs the suite locally

- `spec/system/avo/group_1/hotkey_spec.rb:438` fails on this branch **and on its parent commit**. `ViewComponent::ReusedInstanceError` in `Avo::UI::Tabs::TabComponent#render_badge`, unrelated to this change.
- `spec/system/avo/group_2` reported 26 failures locally on this branch and **the same 26 on the parent commit**. The two that differed between runs each pass and fail non-deterministically across repeats. Local group_2 was not a usable signal here; CI is the gate.

### Still outstanding

- **The Safari check in "Manual review steps" has not been run.** This was developed on Linux with no WebKit available, so the one claim in this PR that has not been executed is the Safari behaviour itself. The new spec pins the browser-agnostic invariant Safari depends on, and that assertion does fail against the old controller in Chrome, but it is a proxy rather than the real thing.

<sub>Generated by an automated Compound Engineering run.</sub>

PR #4762, opened 2026-08-31
Memoize visibility detection for performance speedup

rynonl:
# Description
visible_items walks a container's whole item subtree, and every container asks its children whether they have anything visible (IsResourceItem#visible?, VisibleItems#visible?) while the tab and panel components re-derive it as they render. On resources with many fields spread across tabs, the same subtree gets walked thousands of times per request.

This memoizes the result against the context it was built in (the items holder, the view, and the record identity) so it recomputes when any of those change but not otherwise.

Fixes # (issue)

# Checklist:

- [x] I tested the design in Light and Dark mode, and all default themes

## Screenshots & recording
N/A

## Manual review steps
The performance benefit scales with tabs x fields.

Using the dummy app in the repo, the Person show page renders ~20% faster.

A resource page with 120 fields and 15 tabs saw a ~45% boost to page load time (122ms to 68ms) and the rendered html is identical.
```
class Avo::Resources::BenchPerson < Avo::BaseResource
  self.model_class = ::Person
  self.title = :name

  def fields
    field :name, as: :text, link_to_record: true

    tabs do
      15.times do |t|
        tab title: "Tab #{t}" do
          panel { card(title: "Card #{t}") { 8.times { |f| field :"b_#{t}_#{f}", as: :text do "v" end } } }
        end
      end
    end
  end
end
```

### DIFF
--- a/lib/avo/concerns/has_items.rb
+++ b/lib/avo/concerns/has_items.rb
@@ -209,17 +209,35 @@ module Avo
       #
       # It can't be a plain `||=` because the contextual items can change throughout a
       # request lifecycle.
+      #
+      # The context is compared by object identity, so anything *replaced* recomputes.
+      # `items.size` is in there because `Holder#add_item` appends in place: a holder
+      # that gains an item keeps both its own identity and its array's, so the count is
+      # the only cheap signal that the items changed underneath us.
+      #
+      # Request-scoped state (params, context, current_user) is deliberately absent.
+      # `detect_fields` is a `before_action` and assigns a brand-new `Items::Holder`, so
+      # the memo is already cold at the top of every request and none of those can change
+      # within the lifetime of one entry.
       def visible_items
         hydration_resource = is_a?(Avo::Resources::Base) ? self : try(:resource)
-        context = [@items_holder, view.to_s.to_sym, hydration_resource.try(:record)]
+        context = [@items_holder, items.size, view.to_s.to_sym, hydration_resource.try(:record)]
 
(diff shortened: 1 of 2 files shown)


Paul-Bob:
Hi @rynonl

Thank you for the contribution! Nice strategy with real benefits, and thanks for taking the time to include the metrics.

My one question before merging was whether the view and the record are enough to key this on. I checked and they are — I pushed one commit with two small hardenings rather than block on it:

- **`items.size` in the context.** `Holder#add_item` appends in place, so a holder that gains an item keeps both its own identity and its array's. The count is the only cheap signal that the items moved underneath us. Not reachable today (`detect_fields` builds the holder fully, as a `before_action`), but it's the one mutation the key couldn't structurally see.
- **`freeze` on the memoized array.** `visible_items` now hands the same object to every caller. Nothing in core mutates it, but `avo-forms` includes `HasItems` too — the freeze makes a future mutation loud instead of silently corrupting every later read.

Plus a comment recording *why* request-scoped state is deliberately absent from the key: `detect_fields` is a `before_action` and assigns a brand-new `Items::Holder`, so the memo is already cold at the top of every request and `params` / `context` / `current_user` can't change within the lifetime of one entry. That's the load-bearing argument for this being safe and it wasn't visible from the code. `compute_visible_items` is private now as well.

Two specs added for the above, and comparing with `.equal?` was the right call — a different record object always recomputes, so the only blind spot is "same object, mutated in place", and nothing on the show/edit render path does that.

Merging. Thanks again!

adrianthedev:
Thanks @rynonl!

PR #4777, opened 2026-09-06
fix: skip action records deleted after selection 🤖🤖🤖

### LINKED ISSUE #4078: Action dies when selected record is deleted just before it is pressed
### Describe the bug
When you have an action, e.g. csv download, and someone else deletes one of the records you selected the action will just die. This is something that might happen when multiple users are allowed to act on or delete records.

### Reproduction
just use main avo demo, go to projects, create a new project, open another tab with project index page so you have 2 tabs with the same index page open. Then delete the newly created project in one tab and try to download csv where the deleted project is selected in the other tab.

Avo should just remove records that it can't find as it could have been deleted / discarded or something like that by another user at the same time. The action just dies instead which is never a good thing.

**License type**:

 - [ ] Community
 - [ ] Pro
 - [x] Advanced

 - [x] No

#### Impact

#### Urgency

 - [x] It's a nasty edge case that happens rarely.

be-student:
# Description

Bulk actions now keep every selected record that still exists when another record is deleted between selection and submission. The normal bulk lookup stays fast; only a missing-record race falls back to the resource's own per-record lookup, preserving custom and FriendlyId behavior.

Fixes avo-hq/avo#4078 (AVO-806).

# Checklist

- [x] I updated the shipped action guidance
- [x] No visual behavior changed

## Manual review steps

1. Run `bundle exec rspec spec/requests/avo/actions_request_spec.rb`.

Validation: 5 examples passed; StandardRB passed for the changed Ruby files.

AI disclosure: Codex assisted with implementation, tests, and documentation.

---

## Maintainer follow-up

The original spec targeted the `users` resource, whose model uses FriendlyId — the default `find_record_method` takes the `where(slug: ids)` branch, which never raises, so the example passed on `main` too. It has been moved to `Avo::Resources::Fish` (plain ActiveRecord, `query.find(ids)`), plus a case for every selected record being gone. Both fail without the controller change.

Docs follow-up: avo-hq/docs.avohq.io#683

### DIFF
--- a/app/controllers/avo/actions_controller.rb
+++ b/app/controllers/avo/actions_controller.rb
@@ -80,6 +80,12 @@ module Avo
       else
         []
       end
+    rescue ActiveRecord::RecordNotFound
+      ids.filter_map do |id|
+        @resource.find_record(id, params: params)
+      rescue ActiveRecord::RecordNotFound
+        nil
+      end
     end
 
     def set_fields
--- a/app/controllers/avo/actions_controller.rb
+++ b/app/controllers/avo/actions_controller.rb
@@ -80,6 +80,12 @@ module Avo
       else
         []
       end
+    # A record may be deleted between the moment it was checked on the index and
+    # the moment the action is submitted, and the bulk lookup above raises for the
+    # whole batch when that happens. Fall back to the resource's own per-record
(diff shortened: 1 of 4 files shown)


Paul-Bob:
Thanks for this contribution @be-student!

I pushed a few tweaks and made the docs PR

PR #4780, opened 2026-09-07
Sync shipped skills with docs through 55931cc (2026-09-05)

adrianthedev:
# Description

Routine reconciliation of the gem-shipped skills (`lib/avo/skills/`) against `avo-hq/docs.avohq.io`, covering everything under `docs/4.0` between the recorded marker `ee0277f1` (2026-08-31) and docs HEAD `55931cc` (2026-09-05). Moves `docs-index.json` to the new baseline.

Nine pages changed; three skills needed an edit. Only files under `lib/avo/skills/` are touched.

## Changed pages

| Changed page | Skills citing it | What changed |
| --- | --- | --- |
| `theming.md` (M) | `avo-branding-appearance` | **Edited.** New "Change the font" section — a capability the skill never mentioned. |
| `appearance-api.md` (M) | `avo-branding-appearance` | **Edited.** New Fonts table documenting the two font variables and their defaults. |
| `i18n.md` (M) | `avo-associations`, `avo-i18n` | **Edited `avo-i18n`.** The add-on gems table gained a REST API row, and the "ships English only" claim became "most of them". No change for `avo-associations` — see below. |
| `upgrade.md` (M) | `avo-update` | **Edited.** New "Upgrade to 4.2.0" section, whose four changes are all `avo-api`'s. |
| `ai.md` (M) | — | No change needed — `avo-ai`-owned page, and the edit (chat pages carry no page origin) has no core-skill surface. |
| `custom-controls.md` (M) | — | No change needed here; belongs to `avo-custom_controls`. See follow-ups. |
| `custom-controls-api.md` (M) | — | Same. |
| `rest-api.md` (M) | — | No change needed — `avo-api`-owned page. |
| `rest-api-api.md` (A) | — | New page, `avo-api`-owned. See follow-ups. |

## Skills edited

**`avo-branding-appearance`** — fonts are a genuine capability gap, not a rewording. Avo reads its typeface through the sans font variable (and monospace through the mono one), so a `:root` override in `avo-overrides.css` retypes the whole admin with no build step. Added to the CSS-variable step, plus a gotcha covering the two ways it silently fails: an `@import` that isn't the first rule in the file is ignored outright, and Avo sets text at weights 400/500/600/700 so a partial load gets synthesized faces. Description and the `index.md` line gained font triggers.

**`avo-i18n`** — the skill stated the English-only rule for add-ons without exception. `avo-api` ships `avo.api.token.*` in all 19 locales core does (verified: `gems/avo-api/config/locales/` holds 19 files), so an agent following the old text would have hand-written translations that already exist. Added the REST API row, softened the claim, and noted the `rails-i18n` caveat for the lifecycle chips' relative times.

**`avo-update`** — two fixes, both from the new 4.2.0 section:

- It is the first **bare version heading whose content is entirely an add-on's**. The skill said per-add-on sections name the gem (`## Upgrade to avo-dashboards 4.0.9`), which would have an agent skip 4.2.0 for an app whose `avo-api` crossed it but whose core didn't. `avo-api`'s version is `4.2.0`, matching core's — the paid gems ride a shared release train.
- Upgrade sections can now prescribe a **generator plus a migration**. Verified against the gem's source rather than the docs alone: `AvoApi::TokensGenerator` rescues the already-installed case and says rather than failing (idempotent), while `AvoApi::InstallGenerator` invokes `avo_api:generate` as well, which offers to overwrite customized resource controllers.

## Skills reviewed and left alone

- **`avo-associations`** — flagged only because it links `i18n.md` from its `name` option row as a generic "also achievable via field i18n" pointer. The i18n change was to the add-on namespace table; the link and the surrounding claim are both still correct.

## Follow-ups / could not resolve

Flagging rather than guessing:

- **`custom-controls.md` + `custom-controls-api.md` → `avo-custom_controls`'s own skill.** This one is worth passing on with some urgency: the docs correction reverses a claim that had been wrong since 2023 — a directly declared `action` control *does* honor the action's `visible` block and `authorize` rule. I confirmed no core skill repeats the stale claim (`avo-actions` correctly defers placement to `avo-custom-controls`), so nothing here needed fixing, but the add-on's skill should be checked for it.
- **`rest-api.md` / `rest-api-api.md` → `avo-api`'s own skill (`avo-rest-api` in `package-map.md`).** `rest-api-api.md` is a brand-new page; the API tokens and per-token scopes feature it documents has no owning core skill, which is correct — but the add-on's skill needs the new page in its Docs list.
- **`ai.md` → `avo-ai`'s own skill.**
- **No new core skills created**, per the task's standing instruction. No gap found that warranted one.

## Checklist:

- [x] I have made corresponding changes to the documentation — n/a, this change *is* the docs reconciliation, in the other direction
- [ ] I have added tests that prove my fix is effective or that my feature works — n/a, prose-only; the existing `spec/lib/avo/skills` suite is the gate
- [ ] I tested the design in Light and Dark mode, and all default themes — n/a, no UI

## Screenshots & recording

n/a — Markdown only.

## Manual review steps

Run from the avo checkout, with a `docs.avohq.io` clone available:

```bash
# 1. No VitePress artifacts pasted into the skills. Must print nothing.
grep -rnE '\[!code|:::|<(Option|Image|CustomCode|FieldTypesList)' lib/avo/skills/

# 2. The integrity gate. Passed locally: 167 examples, 0 failures.
bundle exec rspec spec/lib/avo/skills

# 3. Marker moved. Should now report "Up to date".
AVO_DOCS_PATH=/path/to/docs.avohq.io ruby lib/avo/skills/bin/docs_drift.rb
```

Note for step 3: `docs_drift.rb` reads the skill files with the default external encoding, so it raises `invalid byte sequence in US-ASCII` under a non-UTF-8 locale. Prefix with `LANG=C.UTF-8 LC_ALL=C.UTF-8` if you hit that — it's a pre-existing sensitivity in the script, not something this PR introduces, and I've left it alone as out of scope.

Reviewer note: the three prose edits are the whole substance. Worth checking the `avo-update` claim that the paid gems share core's version number, since that generalization is drawn from 4.2.0 being the first instance.

https://claude.ai/code/session_01VsPDwh52DpGsgxiSgiH4Ei


adrianthedev:
Superseded by #4797, which folds this PR's skill edits into one reconciled docs-sync.

PR #4788, opened 2026-09-10
docs: teach avo-troubleshoot the 4.2.2 custom-JS collision

adrianthedev:
# Description

Adds one symptom entry to the `avo-troubleshoot` skill for the packaging leak #4785 fixed.

An app on `4.2.2` whose `avo.custom.js` is served through Sprockets got the gem's stray copy instead of its own: the page loads, the panel renders, and the app's Stimulus controllers never register — no exception, nothing in the log. That is the worst shape a symptom can take for an agent, because there is no reason to suspect the gem's file list; it reads the app's entrypoint, its `_head.html.erb` and its importmap pin, finds all three correct, and has nowhere else to go. avohq.io spent 17 red system specs on it.

The new section says to run `bundle info avo` before reading any of the app's code, and carries the fallback for apps on any other version (entrypoint loaded from the ejected partial, `defer: true` on `javascript_include_tag`) so the entry doesn't dead-end when the version rules it out.

Body-only change to one `SKILL.md` — no frontmatter touched, no skill added, so `index.md` parity and the packaging globs are unchanged. It ships in the gem, so an agent working in a customer's app on `4.2.2` reads it from the version that app has locked.

# Checklist:

- [x] I have made corresponding changes to the [documentation](https://github.com/avo-hq/docs.avohq.io) — [docs.avohq.io#687](https://github.com/avo-hq/docs.avohq.io/pull/687) adds the matching `## Skip version 4.2.2` upgrade entry and the `asset-handling.md` warning this section links to
- [ ] I have added tests that prove my fix is effective or that my feature works — n/a, documentation
- [ ] I tested the design in Light and Dark mode, and all default themes — n/a, no UI

## Manual review steps

1. Open `lib/avo/skills/avo-troubleshoot/SKILL.md` and confirm the new table row's jump link matches the section's generated anchor (`#custom-javascript-loads-but-never-runs`)
2. `bundle exec rspec spec/lib/avo/skills` covers this tree (index integrity, packaging, resolver) — CI runs it here

Note on `ws audit-skills`: it is scoped to `gems/` by design and never reaches core's skills, so it says nothing about this change either way — core's tree is validated by the rspec files above.

https://claude.ai/code/session_01LCimkWnjJixrJBs5N5tAGz

### DIFF
--- a/lib/avo/skills/avo-troubleshoot/SKILL.md
+++ b/lib/avo/skills/avo-troubleshoot/SKILL.md
@@ -84,6 +84,7 @@ grep -rn "explicit_authorization" "$(bundle show avo)/lib"
 | `No valid predicate for combinator` when filtering | `config.ignore_unknown_conditions` is `false` | [↓](#no-valid-predicate-for-combinator-when-filtering) |
 | `undefined method 'xxx_path'` inside an Avo block | Avo is a Rails engine — needs `main_app.` prefix | [↓](#url-helpers-blow-up-inside-avo-blocks) |
 | Index search box does nothing — the unfiltered rows stay | `self.search[:query]` returns an Array; the index needs a relation | [↓](#the-index-search-box-silently-does-nothing) |
+| Custom JS loads but its Stimulus controllers never register | On `4.2.2` only: the gem packaged its own `avo.custom.js` and won that Sprockets logical path | [↓](#custom-javascript-loads-but-never-runs) |
 | License won't validate / status page errors | Key not set on the server; check the status page | [↓](#license-wont-validate) |
 | Tests fail after adding/upgrading Avo (`WebMock::NetConnectNotAllowedError`) | v4's outbound license check to `clerk-*.avohq.io` is blocked | [↓](#tests-fail-after-adding-or-upgrading-avo) |
 | `bundle install` can't fetch `avo-*` / 401 / 403 | packager.dev token not seen by Bundler, or a blocked host in sandboxes | [↓](#bundle-install-cant-fetch-avo--gems) |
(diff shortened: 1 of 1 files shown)


adrianthedev:
`tailwindcss_integration_test` is red here and is not this PR's — it is red on `main` too, and has been since before this branch existed.

**What fails.** `bundle exec rake assets:precompile`, before the test script ever runs:

```
rake aborted!
ArgumentError: unknown keyword: quirks_mode
  json-3.0.2/lib/json/common.rb:382:in 'JSON.generate'
  activesupport-7.1.2/lib/active_support/json/encoding.rb:92:in 'JSONGemEncoder#stringify'
  propshaft-0.8.0/lib/propshaft/processor.rb:35:in 'block in Propshaft::Processor#write_manifest'
```

`json` 3.0 dropped the `quirks_mode` keyword; ActiveSupport 7.1.2 still passes it when serializing, and Propshaft writes its manifest through `to_json`. Deterministic, not a flake — it dies in the same place every run.

**Why it isn't this PR's.** Two independent reasons:

1. **Red on the base branch.** Every run of this workflow since at least 2026-09-09 has failed, `main` pushes included — runs [5911](https://github.com/avo-hq/avo/actions/runs/34326826643) through [5919](https://github.com/avo-hq/avo/actions/runs/34352488799), among them the bare `Bump avo to v4.2.2` and `Bump avo to v4.2.3` commits, which carry no code at all. Last green was [run 5890](https://github.com/avo-hq/avo/actions/runs/33964721225) (`Bump avo to v4.2.1`, 2026-09-05), which fits a `json` release landing in that window — the job resolves fresh gems on every run.
2. **Nothing in this diff can reach it.** The job checks out a different repository (`avo-hq/avo_tailwindcss_test_repo` @ `avo-1147/tailwindcss-integration`) and adds avo at the PR SHA. This PR changes one markdown file under `lib/avo/skills/`. The failure is in that repo's dependency resolution, upstream of any Avo code.

**No fix exists to port, and it can't land here.** The change belongs in the test repo's `Gemfile` — this repository has no say in what it resolves. Two shapes, either would do:

- `gem "json", "< 3"` — smallest, pins the transitive dependency back to a version that still accepts the keyword.
- Bump that repo off Rails 7.1.2. ActiveSupport stopped passing `quirks_mode` in later versions — 8.1.3.1's `JSONGemEncoder#stringify` has no trace of it — so a current Rails fixes it at the source rather than freezing a dependency.

The first unblocks CI today; the second is the one worth doing.

Not re-running the job: a check that fails identically on the base branch is already established as not-mine, and a re-run would only confirm a deterministic `ArgumentError`.

**What was verified for this PR instead:** `spec/lib/avo/skills` — 167 examples, 0 failures, covering index integrity, packaging, the install panel and the resolver against the edited tree. Every other check on this head is green or in progress; `standardrb`, `erb-lint`, `danger`, `i18n` and both `components` jobs have already passed.

---

